PR #23, opened 2023-01-25
RDSK-1750 Add common sensor definition

acmorrow:
@npmenard - This is basically completing the work that we did together at my desk yesterday. There is definitely more to do, but I wanted to get an early review on what I have so far before attempting to making further progress.


src/common/sensor.rs
```diff
@@ -0,0 +1,48 @@
+#![allow(dead_code)]
+
+use crate::common::status::Status;
+use std::collections::BTreeMap;
+
+type SensorReadings = ::std::collections::HashMap<String, ::prost_types::Value>;
```
acmorrow, line 6: @npmenard

- I'm not sure if the type alias is worthwhile, but I was getting tired of typing `std::collections::HashMap`.
- I don't particularly like having `prost_types::Value` leaking into the user interface here. I suspect that this should be `f64`, but I was having trouble elsewhere when I did that.

src/common/grpc.rs
```diff
@@ -301,6 +304,18 @@ impl GrpcServer {
...
 
+    fn sensor_get_readings(&mut self, message: &[u8]) -> anyhow::Result<()> {
+        let req = component::sensor::v1::GetReadingsRequest::decode(message)?;
+        let sensor = match self.robot.lock().unwrap().get_sensor_by_name(req.name) {
+            Some(b) => b,
+            None => return Err(anyhow::anyhow!("resource not found")),
+        };
+
+        let readings = sensor.lock().unwrap().get_readings()?;
+        let resp = component::sensor::v1::GetReadingsResponse { readings };
```
acmorrow, line 315: This works only because the mapped type in the `HashMap` that we are returning from `Sensor::get_readings` is using `prost_types::Value`. If we change it to `f64` this doesn't work, because Rust doesn't seem to automatically convert `HashMap<String, f64>` to `HashMap<String, prost_types::Value>` (not unreasonable). I'd prefer to change to that, but I was not sure what the right way to do the conversion was. I could obviously make a new `HashMap` and iterate and convert, but I'm guessing there is a better way.

src/common/sensor.rs
npmenard: Type aliases a nice to make the code more readable but I agree we need to hide Value from the user since anyone should be able to use this interface.
`f64` make sense here. Also you should probably let each sensor implement their own version of Readings ( in go a Reading is a `map[string]interface{}`

src/common/grpc.rs
npmenard: A rusty way to do it can look like this : https://gist.github.com/rust-play/a548b9e2b360a0c512d906b24a6f54af

acmorrow:
@npmenard - Updated with the changes we looked at last night, so I think this is ready for another round of review. I will leave a few comments.

src/common/grpc.rs
acmorrow: I'm resolving this comment as I think the `GenericReadingsResult` mechanism has addressed it.

src/common/moisture_sensor.rs
```diff
@@ -0,0 +1,48 @@
...
+
+impl MoistureSensor {
+    pub fn new(analog: Rc<RefCell<dyn AnalogReader<u16, Error = anyhow::Error>>>) -> Self {
+        MoistureSensor { analog: analog }
+    }
+}
+
+impl Sensor for MoistureSensor {
+    fn get_generic_readings(&self) -> anyhow::Result<GenericReadingsResult> {
+        Ok(self
```
acmorrow, line 25: I don't particularly like that each sensor that implements `Sensor` needs to re-implement `get_generic_readings` for types that can be serialized. I tried to do some trait magic to make it happen automatically, but I couldn't quite get there:

```
impl<T, U: SensorT<T>> Sensor for U
where
    SensorResult<T>: Into<prost_types::Value>,
{
    fn get_readings(&self) -> anyhow::Result<GenericReadingsResult> {
        let readings = self.get_readings_t()?;
        let bt: ::std::collections::BTreeMap<::prost::alloc::string::String, ::prost_types::Value> =
            readings
                .into_iter()
                .map(|v| (v.0, SensorResult::<T> { value: v.1 }.into()))
                .collect();
        Ok(GenericReadingsResult { fields: bt })
    }
}
```

Fails with:

```
error[E0207]: the type parameter `T` is not constrained by the impl trait, self type, or predicates
  --> src/common/sensor.rs:77:6
   |
77 | impl<T, U: SensorT<T>> Sensor for U
   |      ^ unconstrained type parameter
```

I believe this is because there is nothing here to ensure that a type doesn't implement both, say `SensorT<u32>` *and* `SensorT<u64>`, so which `get_readings_t` did we mean to call?

src/common/sensor.rs
```diff
@@ -0,0 +1,75 @@
+#![allow(dead_code)]
```
acmorrow, line 1: I don't think this is needed and I will remove it before merging anything.

src/common/sensor.rs
```diff
@@ -0,0 +1,75 @@
...
+
+pub trait Sensor: Status {
+    fn get_generic_readings(&self) -> anyhow::Result<GenericReadingsResult>;
+}
+
+pub trait SensorT<T>: Sensor {
+    fn get_readings(&self) -> anyhow::Result<TypedReadingsResult<T>>;
+}
+
+// A local wrapper type we can use to specialize `From` for `prost_types::Value``
```
acmorrow, line 20: The type isn't really "local" so I will remove that from the comment.

src/common/sensor.rs
```diff
@@ -0,0 +1,75 @@
...
+pub struct SensorResult<T> {
+    pub value: T,
+}
+
+impl From<SensorResult<f64>> for ::prost_types::Value {
```
acmorrow, line 25: I could add some more definitions here for other builtins if we think that is useful.

src/common/sensor.rs
```diff
@@ -0,0 +1,75 @@
+#![allow(dead_code)]
+
+use crate::common::status::Status;
+use std::collections::BTreeMap;
+use std::collections::HashMap;
+
+pub type GenericReadingsResult =
+    ::std::collections::HashMap<::prost::alloc::string::String, ::prost_types::Value>;
+
+pub type TypedReadingsResult<T> = ::std::collections::HashMap<String, T>;
```
acmorrow, line 10: I don't particularly love this, but my attempts to make either a generic`Sensor` type or to give `Sensor` a generic `get_readings` did not succeed, primarily because it meant that the type parameter became required everywhere.

src/common/grpc.rs
```diff
@@ -162,6 +162,9 @@ impl GrpcServer {
             "/viam.robot.v1.RobotService/ResourceNames" => self.resource_names(payload),
             "/viam.robot.v1.RobotService/GetStatus" => self.robot_status(payload),
             "/proto.rpc.v1.AuthService/Authenticate" => self.auth_service_authentificate(payload),
+            "/viam.component.sensor.v1.SensorService/GetReadings" => {
```
npmenard, line 165: nit: prefer removing the brackets

src/common/moisture_sensor.rs
```diff
@@ -0,0 +1,48 @@
...
+use std::collections::HashMap;
+use std::rc::Rc;
+
+pub struct MoistureSensor {
+    analog: Rc<RefCell<dyn AnalogReader<u16, Error = anyhow::Error>>>,
+}
+
+impl MoistureSensor {
+    pub fn new(analog: Rc<RefCell<dyn AnalogReader<u16, Error = anyhow::Error>>>) -> Self {
+        MoistureSensor { analog: analog }
```
npmenard, line 19: you can write this as MoistureSensor {analog}

npmenard (approved):
LGTM! before merging run `cargo clippy --no-deps --features native`

I am fine with the implementation for now let see as we go if we can find a better approach!

src/common/moisture_sensor.rs
acmorrow: Done

src/common/grpc.rs
acmorrow: It seems to want them.

PR #28, opened 2023-02-28
[RSDK-2106] Add CI for micro-rdk

npmenard:
Just laying out the bare minimum for CI to work. I have added the Dockerfile for the micro-rdk canon image which is used for the CI but can also be used as a dev environment. The image itself is around 10G we can probably reduce the size by being more selective about what we are pulling etc... 

For now the build step fails because we are relying on a existing robot to pull data from, we can either add an organisation secret that point to a robot on app.viam.com or find another approach, I am more tempted to find said other approach :)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   build_and_test:
     name: Build and Test
-    runs-on: [x64, qemu-host]
+    runs-on: [internal]
     container:
       image: ghcr.io/viamrobotics/micro-rdk-canon:latest
     timeout-minutes: 10
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,7 @@ jobs:
       with:
         ref: ${{ github.event.pull_request.head.sha }}
         fetch-depth: 2
+        
     - name: Write Viam JSON
       uses: "DamianReeves/write-file-action@master"
       with:
@@ -34,16 +35,20 @@ jobs:
         write-mode: overwrite
         contents: |
             ${{ secrets.VIAM_BUILD_JSON }}
+            
     - name: Test
       run: |
         chown -R testbot .
         sudo -u testbot bash -lc 'make test'
+        
     - name: Clippy
       run: |
         sudo -u testbot bash -lc 'make clippy'
+        
     - name: Build ESP32
       run: |
         sudo -u testbot bash -lc 'export MICRO_RDK_WIFI_PASSWORD=0 && . "$IDF_PATH"/export.sh && . "$ESP_ROOT"/export-esp.sh && make build'
+        
     - name: Build Native
       run: |
         sudo -u testbot bash -lc 'make build-native'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,15 +27,16 @@ jobs:
       with:
         ref: ${{ github.event.pull_request.head.sha }}
         fetch-depth: 2
-            
-    - name: Test
-      run: |
-        chown -R testbot .
-        sudo -u testbot bash -lc 'make test'
+          
         
     - name: Clippy
       run: |
+        chown -R testbot .
         sudo -u testbot bash -lc 'make clippy'
+        
+    - name: Format
+      run: |
+        sudo -u testbot bash -lc 'make format'
         
     - name: Build ESP32
       run: |
@@ -44,3 +45,7 @@ jobs:
     - name: Build Native
       run: |
         sudo -u testbot bash -lc 'make build-native'
+        
+    - name: Test
+      run: |
+        sudo -u testbot bash -lc 'make test'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,49 @@
+name: Test
+
+on:
+  workflow_dispatch:
+  pull_request:
+    branches: ['main']
+  push:
+
+jobs:
+  build_and_test:
+    name: Build and Test
+    runs-on: [x64, qemu-host]
+    container:
+      image: ghcr.io/viamrobotics/micro-rdk-canon:latest
+    timeout-minutes: 10
+
+    steps:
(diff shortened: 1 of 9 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,54 @@
...
+  pull_request:
+    branches: ['main']
+  push:
+
+jobs:
+  build_and_test:
+    name: Build and Test
+    runs-on: [x64, qemu-host]
+    container:
+      image: ghcr.io/viamrobotics/micro-rdk-canon:latest
```
Otterverse, line 14: Probably okay for now, but... it looks like you're planning to build multi-arch images in the dockerfile, and it's easier sometimes to tag per architecture, instead of just "latest" here.

Otterverse (commented):
Mostly looks good for now. So just leaving a few comments. I don't think anything is critical, because if this works for your needs, then that's all that really matters for the moment.

etc/docker/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+
+   
+WORKDIR /src/
+
+RUN git clone https://github.com/espressif/qemu
+
+RUN cd qemu &&./configure --target-list=xtensa-softmmu     --enable-gcrypt \
+   --enable-debug --enable-sanitizers  --disable-strip --disable-user \
+   --disable-capstone --disable-vnc --disable-sdl --disable-gtk && cd build && ninja && ninja install
+RUN rm -rf /src
```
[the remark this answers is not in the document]
npmenard: Yep its on purpose once qemu-xtensa is installed we don't need anything else here

etc/docker/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+RUN cargo install ldproxy
+RUN rustup component add clippy-preview
+
+RUN rm -rf $CARGO_HOME/registry
+
+RUN echo ". $ESP_ROOT/export-esp.sh" >> ~/.bashrc && \
+    echo ". $IDF_PATH/export.sh"  >> ~/.bashrc && \
+    echo "PATH=$PATH" >> ~/.bash_profile
+
+USER root
```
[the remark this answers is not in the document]
npmenard: that's a good point I am going to change it to testbot then :)

etc/docker/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+    echo "${rustupSha256} *rustup-init" | sha256sum -c -; \
+    chmod +x rustup-init; \
+    ./rustup-init -y --no-modify-path --profile minimal --default-toolchain $RUST_VERSION --default-host ${rustArch}; \
+    rm rustup-init; \
+    chmod -R a+w $RUSTUP_HOME $CARGO_HOME
+
+   
+WORKDIR /src/
+
+RUN git clone https://github.com/espressif/qemu
```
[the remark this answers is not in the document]
npmenard: good point I am going to lock it to a specific version

.github/workflows/test.yml
npmenard: yep I have to work my way through buildx so I can generate non native arch

etc/docker/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+
+ENV RUSTUP_HOME=/opt/rust/rustup \
+    CARGO_HOME=/opt/rust/cargo \
+    PATH=/opt/rust/cargo/bin:$PATH \
+    RUST_VERSION=1.65.0 \
+    ESPRS_VERSION=v1.65.0.0 \
+    ESP_ROOT=/opt/esp \
+    IDF_TOOLS_PATH=/opt/esp/toolchain
+
+RUN echo "deb http://deb.debian.org/debian bullseye-backports main" >> /etc/apt/sources.list.d/backports.list
```
acmorrow, line 14: You can also do this with `apt-add-repository` from `software-properties-common`:
- https://github.com/viamrobotics/viam-cpp-sdk/blob/961544fa1490c3514f26eae7c255d46a7703fe9f/etc/docker/Dockerfile.debian.bullseye#L27
- https://github.com/viamrobotics/viam-cpp-sdk/blob/961544fa1490c3514f26eae7c255d46a7703fe9f/etc/docker/Dockerfile.debian.bullseye#L31

etc/docker/entrypoint.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
acmorrow, line 1: The purpose of this entrypoint file is somewhat opaque to me. Can I get some additional context? The majority of the file seems to be about dealing with the cases where `testbot-uid` or `testbot-gid` are provided as arguments, but I don't see anything in this repo that does that.

acmorrow (approved):

etc/docker/entrypoint.sh
npmenard: @Otterverse would explain better than me but my understanding is that it is a workaround permission issues. Docker can make virtual permissions, which mean that if a user has a UID different from testbot then we may not be able to created files or created files may not be readable by said user. This is mainly relevant when using the image for compiling a local repository (for ex `docker run -it --rm --name esp32 --mount type=bind,source="$(pwd)"/docker,target=/src -v "$SSH_AUTH_SOCK:$SSH_AUTH_SOCK" -e SSH_AUTH_SOCK="$SSH_AUTH_SOCK" test-rust-image:latest --testbot-uid 1000 --testbot-gid 1000  '. "$IDF_PATH"/export.sh && . "$IDF_PATH"/../export-esp.sh && make build'`)

Otterverse: This was copied from rdk basically, and the canon image that uses (which is in the SysOps repo.) It allows the "make canon-shell" (and other canon targets) to remap UID/GIDs internally. But it's not used directly by the CI.

As mentioned to @npmenard in conversation, I'm in the process of creating a canon utility that will handle this at runtime, so this could be removed entirely if it's not needed for the next short while until that utility is ready for release. (Without the makefile targets to make use of this, is't a complex command to run, so unlikely to be used.)

PR #29, opened 2023-03-08
Static config esp32

npmenard:
First draft for supporting configuration ingestion at runtime by the esp32, this is not the 'end state' but a good step towards a more correct approach. This should let us add easily more components to the config until we finalize the proper approach.

Supports Board and Motor, would still need to test on some hardware.


src/common/analog.rs
```diff
@@ -32,3 +34,86 @@ pub trait AnalogReader<Word> {
...
+            _ => Err(AttributeError::ConversionImpossibleError),
+        }
+    }
+}
+
+impl TryFrom<&Kind> for AnalogReaderConfig {
+    type Error = AttributeError;
+    fn try_from(value: &Kind) -> Result<Self, Self::Error> {
+        match value {
+            Kind::StructValueStatic(v) => {
```
acmorrow, line 68: The T / &T impl duplication is a nuisance. I assume that had you found a way to avoid needing to write it all you would have done so, but it does at least look like the body of the two functions could be lifted into a common function. Something like:

```
fn arc_try_from(value: &Kind) -> Result<AnalogReaderConfig, AttributeError> {
    match value { ... }
}

impl TryFrom<Kind> for AnalogReaderConfig {
    type Error = AttributeError;
    fn try_from(value: Kind) -> Result<Self, Self::Error> {
        return arc_try_from(&value);
    }
}

impl TryFrom<Kind> for AnalogReaderConfig {
    type Error = AttributeError;
    fn try_from(value: &Kind) -> Result<Self, Self::Error> {
        return arc_try_from(value);
    }
}

```

src/esp32/board.rs
```diff
@@ -1,53 +1,228 @@
 #![allow(dead_code)]
-use super::pin::PinExt;
```
acmorrow, line 2: Does `PinExt` still serve a purpose?

src/esp32/board.rs
```diff
@@ -1,53 +1,228 @@
...
+                                            RefCell<dyn AnalogReader<u16, Error = anyhow::Error>>,
+                                        > = Rc::new(RefCell::new(Esp32AnalogReader::new(
+                                            v.name.to_string(),
+                                            AdcChannelDriver::<_, Atten11dB<ADC1>>::new(unsafe {
+                                                esp_idf_hal::gpio::Gpio33::new()
+                                            })
+                                            .ok()?,
+                                            adc1,
+                                        )));
+                                        Some(p)
```
acmorrow, line 92: It looks like AdcChannelDriver::new only needs an argument of type `impl Peripheral<P = T> + 'd`. If you made a helper function that constructed `esp_idf_hal::gpio::GpioN` by number, could you just match on the range to validate that you had an ADC capable pin and then eliminate the duplication?

acmorrow (approved):
This LGTM with questions and some small suggestions.

Makefile
```diff
@@ -24,7 +24,7 @@ build-native:
 	cd examples && cargo build  --example native
 
 build-qemu:
-	cd examples && cargo build  --example esp32  --features qemu --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort && cargo espflash save-image --features qemu --merge --chip esp32 target/xtensa-esp32-espidf/debug/debug.bin -T esp32/partitions.csv -s 4M  --example esp32 --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort
+	cd examples && cargo build  --example esp32_static  --features qemu --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort && cargo espflash save-image --features qemu --merge --chip esp32 target/xtensa-esp32-espidf/debug/debug.bin -T esp32/partitions.csv -s 4M  --example esp32_static --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort
```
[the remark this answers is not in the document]
npmenard: I would leave it for now until we have more features supported with staticConfig, ultimately this will be the default server build

examples/esp32/esp32.rs
```diff
@@ -93,7 +94,7 @@ fn main() -> anyhow::Result<()> {
             chan2,
         );
 
-        let pins = vec![PinDriver::output(periph.pins.gpio15)?];
+        let pins = vec![PinDriver::output(periph.pins.gpio15.downgrade_output())?];
```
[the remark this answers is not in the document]
npmenard: We have to downgrade it to make it a `AnyOutputPin` which is what the esp32 board expect

src/esp32/motor.rs
```diff
@@ -274,6 +296,35 @@ where
     pub fn new(a: A, b: B, pwm: PWM) -> Self {
         MotorEsp32 { a, b, pwm }
     }
+    pub(crate) fn from_config(cfg: ConfigType, _: Option<BoardType>) -> anyhow::Result<MotorType> {
+        match cfg {
+            ConfigType::Static(cfg) => {
+                if let Ok(pins) = cfg.get_attribute::<MotorConfig>("pins") {
+                    if pins.a.is_some() && pins.b.is_some() {
+                        use esp_idf_hal::units::FromValueType;
+                        let pwm_tconf = TimerConfig::default().frequency(10.kHz().into());
```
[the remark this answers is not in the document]
npmenard: default PWM freq for motor control, same here we would probably support a frequency setting in the future

src/esp32/motor.rs
```diff
@@ -1,4 +1,7 @@
 #![allow(dead_code)]
+use esp_idf_hal::gpio::{AnyOutputPin, Output, PinDriver};
+use esp_idf_hal::ledc::config::TimerConfig;
+use esp_idf_hal::ledc::{LedcDriver, LedcTimerDriver, CHANNEL0, CHANNEL1, CHANNEL2};
```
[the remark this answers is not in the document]
npmenard: no reason in particular, I am not entirely set on how I want us to construct ADC/PWM etc... I want it to be handled by the board so we have some logic that let's us consume config and assign resources appropriately

src/esp32/motor.rs
```diff
@@ -328,3 +379,67 @@ where
...
     }
 }
+
+/// Below is a first attempt as an approach to runtime configuration, the problem raise with runtime configuration is
+/// enforcing single instance of any peripheral at any point in the program. For example say you have a Motor that uses pins 33,34 and 35 and
+/// a AnalogReader that uses pin 35. This situation is wrong since two objects own pin 35. In embedded rust this is avoided by having any hardware
+/// peripherals be singleton and leveraging the borrow checker so that single ownership rules are enforced. When dealing with runtime configuration,
+/// the borrow checker cannot help us. We can however follow the singleton approach and wrap peripherals into options that will be 'taken out' when
+/// something needs an hardware component. Following is an implementation of the proposed approach, a significant limitation is that the hardware can
+/// only be taken once and can never be returned.
```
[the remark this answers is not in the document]
npmenard: Yes we could use Rc/Arc pointers for that, the big question remains how do we value flexibility with added overhead. This would make a lot of sense if we want to support full live reconfiguration like in RDK, we could also accept that reconfiguration requires a reboot!

src/common/analog.rs
npmenard: I think we could use the [std](https://doc.rust-lang.org/std/index.html)::[borrow](https://doc.rust-lang.org/std/borrow/index.html)::[Borrow](https://doc.rust-lang.org/std/borrow/trait.Borrow.html#) trait I haven't played with this yet will had a ticket to track this

src/esp32/board.rs
npmenard: Yes, we need this to construct motor toe for example. In practice we could do away with it but that would mean making the ESP32Motor only work on esp32 so we would have to re-implement it for each new platform we support. I would ideally keep it abstracted if we can it just unclear for now if we can actually achieve this

src/esp32/board.rs
npmenard: The ADC trait is not implemented for `AnyOutputPin` onlye for GpioXX by the `esp-idf-hal` crate. We can't construct a GpioN just using the pin number. I am leaning toward moving all this logic on the esp32 board and use a macro to generate the relevant code. But that would take a separate PR

PR #34, opened 2023-03-28
RSDK-767 - add I2C trait infrastructure to board

gvaradarajan:


npmenard:
What is the rational behind adding the I2C trait to the board trait? Ideally we want to keep the Board trait as similar as possible to the GRPC api. I don't see the benefit vs having board implementation keep I2C handles or similar especially since a board may have multiple i2c interfaces. That might interfere with the configuration process so we would have to figure out something here we that can wait a bit.

src/common/i2c.rs
```diff
@@ -0,0 +1,170 @@
+#![allow(dead_code)]
+
+use super::config::{AttributeError, Kind};
+use std::sync::{Arc, Mutex};
+
+// A trait representing blocking I2C communication for a board. TODO: replace with the
+// embedded_hal I2C trait when supporting boards beyond ESP32. AddressType is
+// either u8 (indicating support for 7-bit addresses) or u16 (for supporting 10-bit addresses)
+pub trait I2CHandle<AddressType> {
```
npmenard, line 9: Don't think we need to have the address type be a template parameter, it's going to be whatever is implementing the trait that is going to make the distinction between 10/7 at the config/build level

npmenard (commented):
looks great just a comment about the I2C trait :)

src/common/i2c.rs
```diff
@@ -0,0 +1,170 @@
...
+    fn name(&self) -> String;
+
+    fn read_i2c(&mut self, _address: AddressType, _buffer: &mut [u8]) -> anyhow::Result<()> {
+        anyhow::bail!("read_i2c unimplemented")
```
npmenard, line 13: might be worth considering adding a function like this one (https://docs.rs/embedded-hal/latest/embedded_hal/blocking/i2c/trait.WriteIterRead.html) since some I2c hardware requires you to write to a register before reading it. We can wait until we are implementing an actual hardware piece !

gvaradarajan: funny story; I had write_read in the trait previously but I removed it by accident, I'll add it back

PR #37, opened 2023-04-11
RSDK-767 - add mpu6050 driver

gvaradarajan:


src/common/mpu6050.rs
```diff
@@ -0,0 +1,348 @@
...
+        self.linear_acceleration
+    }
+
+    fn set_error(&mut self, err: Option<Arc<anyhow::Error>>) {
+        self.error = err;
+    }
+
+    fn set_linear_acceleration_from_reading(&mut self, reading: &[u8; 14]) {
+        let mut slice_copy = vec![0; 6];
+        slice_copy.clone_from_slice(&(reading[0..6]));
```
npmenard, line 69: avoid making a copy here you should be able to convert from the buffer directly (use from_be_bytes/from_le_bytes)

src/common/mpu6050.rs
```diff
@@ -0,0 +1,348 @@
...
+        let raw_state = MPU6050State::new();
+        let (canceller, rx) = mpsc::channel();
+        let state = Arc::new(Mutex::new(raw_state));
+        // reference copies for sending memory into thread
+        let mut i2c_handle_copy = Arc::clone(&i2c_handle);
+        let state_copy = Arc::clone(&state);
+        // start a polling thread that reads from the MPU every millisecond and mutates state.
+        // This allows multi-read access to the state for the functions satisfying the
+        // Movement Sensor API
+        thread::spawn(move || loop {
```
npmenard, line 118: we can't spawn a thread here on the esp32 platform we don't have much control on priority/stack allocation/core allocation and it could preempt the thread running the executor slowing down IO or worse deadlocking.

Can we just poll on demand for now? (eg read i2c when a grpc request comes in?) It will be slower but given the high latency I don't think it would be significant.

Ultimately we could spwan a future on the local executor

src/esp32/board.rs
```diff
@@ -17,6 +17,7 @@ use esp_idf_hal::adc::AdcDriver;
 use esp_idf_hal::adc::Atten11dB;
 use esp_idf_hal::adc::ADC1;
 use esp_idf_hal::gpio::{AnyInputPin, AnyOutputPin, Output, PinDriver};
+// use esp_idf_hal::peripherals::Peripherals;
```
npmenard, line 20: nit: remove this line

npmenard (changes requested):
Done with first pass, left a couple of comments

npmenard (approved):
LGTM

PR #38, opened 2023-04-13
RSDK-2618 - implement adxl-345 accelerometer driver

gvaradarajan:


src/common/adxl345.rs
```diff
@@ -0,0 +1,174 @@
...
+                }
+                Ok(Arc::new(Mutex::new(ADXL345::new(i2c_handle, 83)?)))
+            }
+        }
+    }
+
+    pub fn close(&mut self) -> anyhow::Result<()> {
+        // put the MPU in the sleep state
+        let off_data: [u8; 2] = [STANDBY_MODE_REGISTER, 0];
+        if let Err(err) = self.i2c_handle.write_i2c(self.i2c_address, &off_data) {
```
npmenard, line 86: is the sleep command used for other purpose than when closing the ADXL?

gvaradarajan: using the driver in rdk as a reference, there seems to be no other use case

PR #42, opened 2023-05-12
RSDK-2937 - implement single pin encoder for Esp32

gvaradarajan:
Hardware Tests Conducted:
- two motor-encoder pairs on viam-rover working (single encoders)
- motor testing rig with quadrature encoder

### DIFF
--- a/src/common/encoder.rs
+++ b/src/common/encoder.rs
@@ -100,6 +100,25 @@ pub trait Encoder: Status {
     }
 }
 
+#[derive(Clone, Copy)]
+pub enum Direction {
+    Forwards,
+    Backwards,
+    StoppedForwards,
+    StoppedBackwards,
+}
+
+impl Direction {
+    pub fn is_forwards(&self) -> bool {
+        matches!(self, Self::Forwards) || matches!(self, Self::StoppedForwards)
+    }
+}
+
+pub trait SingleEncoder: Encoder {
+    fn set_direction(&mut self, dir: Direction) -> anyhow::Result<()>;
+    fn get_direction(&self) -> anyhow::Result<Direction>;
+}
+
 pub(crate) type EncoderType = Arc<Mutex<dyn Encoder>>;
 
 pub struct FakeIncrementalEncoder {
@@ -257,5 +276,31 @@ where
     }
     fn get_position(&self, position_type: EncoderPositionType) -> anyhow::Result<EncoderPosition> {
         self.lock().unwrap().get_position(position_type)
+    }
+}
+
+impl<A> SingleEncoder for Mutex<A>
+where
+    A: ?Sized + SingleEncoder,
+{
+    fn set_direction(&mut self, dir: Direction) -> anyhow::Result<()> {
+        self.get_mut().unwrap().set_direction(dir)
+    }
+
+    fn get_direction(&self) -> anyhow::Result<Direction> {
+        self.lock().unwrap().get_direction()
+    }
+}
+
+impl<A> SingleEncoder for Arc<Mutex<A>>
+where
+    A: ?Sized + SingleEncoder,
+{
+    fn set_direction(&mut self, dir: Direction) -> anyhow::Result<()> {
+        self.lock().unwrap().set_direction(dir)
+    }
+
+    fn get_direction(&self) -> anyhow::Result<Direction> {
+        self.lock().unwrap().get_direction()
     }
 }
--- a/src/esp32/pulse_counter.rs
+++ b/src/esp32/pulse_counter.rs
@@ -1,0 +1,55 @@
+use esp_idf_sys::{pcnt_isr_service_install, pcnt_isr_service_uninstall, EspError, ESP_OK};
+use std::sync::atomic::{AtomicBool, AtomicU32, Ordering};
+use std::sync::Arc;
+
+/*
+This module exists because we want to ensure uniqueness of unit number
+across instances of an Esp32 Pulse Counter unit and make sure the isr service
+is only installed once.
+
+THIS MODULE IS A TEMPORARY MEASURE. When abstracting the atomicity of Esp32
+peripherals to board, this logic should be moved there.
+
+TODO: v5 of ESP-IDF has refactored pulse counter to manage what this module
+accomplishes for us. Potentially only use this module when on chips on v4.
+
+*/
+
+lazy_static::lazy_static! {
+    static ref NEXT_UNIT: Arc<AtomicU32> = Arc::new(AtomicU32::new(0));
+
+    static ref ISR_INSTALLED: Arc<AtomicBool> = Arc::new(AtomicBool::new(false));
+
+    static ref NUMBER_OF_UNITS: Arc<AtomicU32> = Arc::new(AtomicU32::new(0));
+}
+
+pub(crate) fn get_unit() -> anyhow::Result<u32> {
+    NUMBER_OF_UNITS.fetch_add(0, Ordering::Relaxed);
+    Ok(NEXT_UNIT.fetch_add(1, Ordering::SeqCst))
+}
+
+pub(crate) fn isr_install() -> anyhow::Result<()> {
+    if !ISR_INSTALLED.fetch_or(true, Ordering::SeqCst) {
+        unsafe {
+            match pcnt_isr_service_install(0) {
+                ESP_OK => {}
+                err => return Err(EspError::from(err).unwrap().into()),
+            }
+        }
+    }
+    Ok(())
+}
+
+pub(crate) fn isr_installed() -> bool {
+    ISR_INSTALLED.load(Ordering::SeqCst)
+}
+
+pub(crate) fn isr_remove_unit() {
+    if NUMBER_OF_UNITS.fetch_sub(1, Ordering::Relaxed) <= 1
+        && ISR_INSTALLED.fetch_xor(false, Ordering::SeqCst)
+    {
+        unsafe {
+            pcnt_isr_service_uninstall();
+        }
+    };
+}
--- a/src/esp32/single_encoded_motor.rs
+++ b/src/esp32/single_encoded_motor.rs
@@ -1,0 +1,79 @@
+use super::single_encoder::SingleEncoderType;
+use crate::common::encoder::{
+    Direction, Encoder, EncoderPositionType, EncoderSupportedRepresentations, SingleEncoder,
+};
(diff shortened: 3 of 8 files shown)


src/esp32/pulse_counter.rs
```diff
@@ -0,0 +1,50 @@
...
+THIS MODULE IS A TEMPORARY MEASURE. When abstracting the atomicity of Esp32
+peripherals to board, this logic should be moved there.
+
+TODO: v5 of ESP-IDF has refactored pulse counter to manage what this module
+accomplishes for us. Potentially only use this module when on chips on v4.
+
+*/
+
+lazy_static::lazy_static! {
+    static ref NEXT_UNIT: Arc<AtomicU32> = Arc::new(AtomicU32::new(0));
```
npmenard, line 19: Feels weird to wrap an atomic inside an Arc, would a zero size type make more sense?

src/esp32/pulse_counter.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    static ref ISR_INSTALLED: Arc<AtomicBool> = Arc::new(AtomicBool::new(false));
```
npmenard, line 21: doesn't this count properly multiple instance of encoder (isr doesn't get uninstall when calling isr_uninstall?)

src/common/encoder.rs
```diff
@@ -100,6 +100,25 @@ pub trait Encoder: Status {
     }
 }
 
+#[derive(Clone, Copy)]
+pub enum Direction {
+    Forwards,
+    Backwards,
+    StoppedForwards,
```
npmenard, line 107: do we need these two?

src/esp32/single_encoded_motor.rs
```diff
@@ -0,0 +1,78 @@
...
+            x if x == 0.0 => {
+                let prev_dir = self.encoder.get_direction()?;
+                match prev_dir {
+                    Direction::Backwards | Direction::StoppedBackwards => {
+                        Direction::StoppedBackwards
+                    }
+                    Direction::Forwards | Direction::StoppedForwards => Direction::StoppedForwards,
+                }
+            }
+            _ => anyhow::bail!("how did we get here?"),
```
npmenard, line 35: use unreachable!() or Ord::cmp :)

src/esp32/single_encoder.rs
```diff
@@ -0,0 +1,341 @@
...
+use espsys::{esp, EspError, ESP_OK};
+
+use std::collections::BTreeMap;
+use std::sync::atomic::{AtomicI32, Ordering};
+use std::sync::{Arc, Mutex};
+
+use crate::common::status::Status;
+
+// TODO: Make configurable?
+const MAX_GLITCH_MICROSEC: u16 = 1;
```
npmenard, line 30: why did you end up needing this?

src/esp32/single_encoder.rs
```diff
@@ -0,0 +1,341 @@
...
+        println!("unit received: {:?}", unit);
+        let pcnt = Box::new(PulseStorage {
+            acc: Arc::new(AtomicI32::new(0)),
+            unit,
+        });
+        let mut enc = Esp32SingleEncoder {
+            pulse_counter: pcnt,
+            config: pcnt_config_t {
+                pulse_gpio_num: encoder_pin.pin(),
+                ctrl_gpio_num: -1,
```
npmenard, line 54: did you try to set it to the pin dir or one of the A/B pin?

src/esp32/single_encoder.rs
```diff
@@ -0,0 +1,341 @@
...
+impl Encoder for Esp32SingleEncoder {
+    fn get_properties(&mut self) -> EncoderSupportedRepresentations {
+        EncoderSupportedRepresentations {
+            ticks_count_supported: true,
+            angle_degrees_supported: false,
+        }
+    }
+    fn get_position(&self, position_type: EncoderPositionType) -> anyhow::Result<EncoderPosition> {
+        match position_type {
+            EncoderPositionType::TICKS | EncoderPositionType::UNSPECIFIED => {
```
npmenard, line 224: this still feels weird to me. Do we expect an encoder type to support both absolute or Ticks?

src/esp32/single_encoder.rs
```diff
@@ -0,0 +1,341 @@
...
+    #[inline(always)]
+    #[link_section = ".iram1.pcnt_srv"]
+    unsafe extern "C" fn irq_handler_increment(arg: *mut core::ffi::c_void) {
+        let arg: &mut PulseStorage = &mut *(arg as *mut _);
+        let mut status = 0;
+        esp_idf_sys::pcnt_get_event_status(arg.unit, &mut status as *mut c_ulong);
+        if status & pcnt_evt_h_lim != 0 {
+            arg.acc.fetch_add(1, Ordering::SeqCst);
+        }
+        if status & pcnt_evt_l_lim != 0 {
```
npmenard, line 195: do we need both?

src/esp32/single_encoder.rs
```diff
@@ -0,0 +1,341 @@
...
+        self.dir = dir;
+        let isr_is_installed = isr_installed();
+        if reconfigure && isr_is_installed {
+            unsafe {
+                match esp_idf_sys::pcnt_counter_pause(self.config.unit) {
+                    ESP_OK => {}
+                    err => return Err(EspError::from(err).unwrap().into()),
+                }
+
+                match esp_idf_sys::pcnt_unit_config(&self.config as *const pcnt_config_t) {
```
npmenard, line 269: do we need to go through all of the config steps or can we just change the interrupt function then resume

npmenard (commented):
done with first pass !

npmenard (changes requested):
done with first pass !

src/common/encoder.rs
gvaradarajan: maybe not? I'm using it to keep track of whether to reverse the counter once the motor is run after having been stopped

src/esp32/pulse_counter.rs
gvaradarajan: I'm don't quite see how to use a zero size type to produce the same functionality. However, I can explain the Arc. Atomics are thread-safe but don't allow shared ownership and that led to some problems whose nature I don't quite remember.

src/esp32/pulse_counter.rs
gvaradarajan: I'm not sure what you mean here? If you mean that it appears dropping one encoder from memory will cause another to cease to function, this was an oversight. I'll change the behavior

src/esp32/single_encoded_motor.rs
gvaradarajan: thanks, I was wondering how I was supposed to deal with that

src/esp32/single_encoder.rs
gvaradarajan: Some of these steps are necessary, but perhaps not all. Reconfiguring the unit disables the high and low events, so those need to be enabled. I found that I had to re-add the glitch filter, but this might actually be an indicator of a bug in the original `setup_pcnt` function. I'll play around with it on hardware tomorrow

src/esp32/single_encoder.rs
gvaradarajan: when I had two encoders wired at the same time, I found that powering one encoder's associated motor was creating enough noise on the other encoder's signal to increment the count for both units instead of just the one. The glitch filter solves this problem.

src/esp32/single_encoder.rs
gvaradarajan: The single encoder implementation in RDK simply counts events received by interrupt, so I thought it was safer to emulate that behavior as closely as possible.

src/esp32/single_encoder.rs
gvaradarajan: Probably not? I recall trying using only one or the other and finding it not to work, but there may have been other problems with my code at the time. I'll try removing

src/esp32/single_encoder.rs
```diff
@@ -0,0 +1,341 @@
...
+            arg.acc.fetch_add(1, Ordering::SeqCst);
+        }
+        if status & pcnt_evt_l_lim != 0 {
+            arg.acc.fetch_add(1, Ordering::SeqCst);
+        }
+    }
+
+    #[inline(always)]
+    #[link_section = ".iram1.pcnt_srv"]
+    unsafe extern "C" fn irq_handler_decrement(arg: *mut core::ffi::c_void) {
```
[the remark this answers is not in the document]
gvaradarajan: I agree, I'll play around with maybe using an Arc that's aware of the direction and can increment/decrement accordingly

src/esp32/single_encoder.rs
gvaradarajan: It feels weird to me too, but currently we assume that as a possibility in our proto API design for encoder, so I felt I had to support it

src/esp32/single_encoder.rs
npmenard: I agree but if we can leverage the pcnt as much as possible and avoid code it would be best. 

With the control pin you can set the counter to increase or decrease based on the pin level so if it's attached to the dir pin it can make the encoder work like the Quadrature encoder. It should also work with A/B. 

Otherwise you can configure the PCNT to count backward or forward programmatically.

src/esp32/pulse_counter.rs
npmenard: My understanding is that the Arc count will be the same as the Atomic variable to you might be able to get away with just using the arc count and not the inner atomic

src/esp32/single_encoder.rs
npmenard: but does the config changes? or just the interrupts?

gvaradarajan: update: I did have to re-add the glitch filter

src/esp32/single_encoder.rs
gvaradarajan: Wait, so why do we have a single encoder implementation even in RDK? Couldn't we have just used the incremental encoder implementation and supply it with the dir or A/B pin?

src/esp32/pulse_counter.rs
gvaradarajan: None of the methods for returning the reference count on an Arc appear to be thread-safe, is this ok?

npmenard (approved):
LGTM

PR #47, opened 2023-06-02
[RSDK-3397] stop pin config change when getting pin level

mattjperez:
Behavior before:
- changed pin config to Input after reading the pin level, breaking ability to change pin state anytime afterwards

Behavior after:
- will not change pin configuration
- when pin level is set and floating, will read the level last set (given minimal environmental interference)
- when pin level is set and connected, will read the level of what is connected (low on ground, high on 3.3v)


npmenard (approved):
LGTM pending passing the CI

PR #49, opened 2023-06-07
[RSDK-2620] base movement 1 of 4: go_for()

mattjperez:


src/common/error.rs
```diff
@@ -0,0 +1,27 @@
+#![allow(dead_code)]
```
npmenard, line 1: I would rather leave the errors in config.rs for now, we might want to revisit this when we work on transitioning away from using anyhow

src/common/error.rs
```diff
@@ -0,0 +1,27 @@
...
+#[derive(Error, Debug, Eq, PartialEq)]
+pub enum AttributeError {
+    #[error("failed to parse number")]
+    ParseNumError,
+    #[error("value not possible")]
+    ConversionImpossibleError,
+    #[error("attribute `{0}` was not found")]
+    KeyNotFound(String),
+    #[error("config has no attribute map")]
+    NoAttributeMap,
```
npmenard, line 14: I don't think we need this error returning absence of a key is enough

src/common/motor.rs
```diff
@@ -35,47 +48,56 @@ pub(crate) struct MotorConfig {
...
-                if v.contains_key("b") {
-                    motor.b = Some(v.get("b").unwrap().try_into()?);
-                }
+                motor.pwm = v
+                    .get("pwm")
+                    .ok_or(AttributeError::KeyNotFound("pwm".to_string()))?
+                    .try_into()?;
+                motor.a = Some(
+                    v.get("a")
+                        .ok_or(AttributeError::KeyNotFound("a".to_string()))?
```
npmenard, line 66: I don't think we should be returning an error in the absence of A or B some motor controller will have logics different from  A/B/PWM

src/common/motor.rs
```diff
@@ -141,6 +171,18 @@ impl Motor for FakeMotor {
...
         Ok(())
     }
+    fn go_for(&mut self, rpm: f64, revolutions: f64) -> anyhow::Result<()> {
+        // get_max_rpm
+        let (pwr, dur) = go_for_math(self.max_rpm, rpm, revolutions).unwrap(); 
+        if revolutions == 0.0 {
+            self.set_power(pwr)?;
+        } else {
+            self.set_power(pwr)?;
+            std::thread::sleep(dur);
```
npmenard, line 181: we can't use thread sleep here because it will prevent the thread running the executor from making progress. We could return a Future to be polled to completion at an higher level

src/esp32/motor.rs
```diff
@@ -296,6 +328,20 @@ where
...
+    fn go_for(&mut self, rpm: f64, revolutions: f64) -> anyhow::Result<()> {
+        let (pwr, dur) = go_for_math(self.max_rpm, rpm, revolutions).unwrap();
+
+        assert!(revolutions >= 0.0);
+
+        if revolutions == 0.0 {
+            self.set_power(pwr)?;
+        } else {
+            self.set_power(pwr)?;
+            std::thread::sleep(dur);
```
npmenard, line 340: same as above

src/common/motor.rs
```diff
@@ -161,6 +203,26 @@ impl Status for FakeMotor {
         Ok(Some(prost_types::Struct { fields: bt }))
     }
 }
+// If revolutions is 0, the returned wait duration will be 0 representing that
+// the motor should run indefinitely.
+pub(crate) fn go_for_math(max_rpm: f64, rpm: f64, revolutions: f64) -> Result<(f64, Duration), ()> {
```
npmenard, line 208: should be moved to math_utils

src/common/motor.rs
```diff
@@ -161,6 +203,26 @@ impl Status for FakeMotor {
...
+    }
+
+    if revolutions == 0.0 {
+        return Ok((rpm / max_rpm, Duration::from_millis(0)));
+    }
+
+    let dir = rpm * revolutions / (revolutions * rpm).abs();
+    let pct = rpm.abs() / max_rpm * dir;
+    let dur = Duration::from_secs_f64((revolutions / rpm).abs() * 60.0);
+    Ok((pct, dur))
```
npmenard, line 224: Should also add a test for this function and it's potential output

src/common/motor.rs
```diff
@@ -161,6 +203,26 @@ impl Status for FakeMotor {
...
+    // need to do this so time is reasonable
+    let mut rpm = rpm;
+    if rpm > max_rpm {
+        rpm = max_rpm
+    } else if rpm < -1.0 * max_rpm {
+        rpm = -1.0 * max_rpm;
+    }
+
+    if revolutions == 0.0 {
+        return Ok((rpm / max_rpm, Duration::from_millis(0)));
```
npmenard, line 218: Why not using an Option here?

src/common/grpc.rs
```diff
@@ -300,8 +300,21 @@ where
...
+        let req = component::motor::v1::GoForRequest::decode(message)
+            .map_err(|_| GrpcError::RpcInvalidArgument)?;
+        let motor = match self.robot.lock().unwrap().get_motor_by_name(req.name) {
+            Some(m) => m,
+            None => return Err(GrpcError::RpcUnavailable),
+        };
+        motor
+            .lock()
+            .unwrap()
+            .go_for(req.rpm, req.revolutions)
```
npmenard, line 314: In RDK this call is non-blocking and we should follow the same pattern

npmenard (changes requested):
Done with first pass. Did you test the implementation on hardware?

mattjperez:
> Done with first pass. Did you test the implementation on hardware?

Thanks! 

Not yet. Made a lot of changes so far and wanted to get a review pass in to make sure I'm heading in the right direction.
Will test on hardware after I apply the updates.

src/common/math_utils.rs
```diff
@@ -27,3 +28,24 @@ impl From<Vector3> for common::v1::Vector3 {
         }
     }
 }
+// If revolutions is 0, the returned wait duration will be 0 representing that
+// the motor should run indefinitely.
+pub(crate) fn go_for_math(max_rpm: f64, rpm: f64, revolutions: f64) -> Result<(f64, Duration), ()> {
```
gvaradarajan, line 33: I don't see any case where this returns an Error, can we just return the tuple instead of a Result?

src/esp32/motor.rs
```diff
@@ -77,6 +78,7 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
 pub struct EncodedMotor<M, Enc> {
     motor: M,
     enc: Enc,
+    max_rpm: f64,
```
gvaradarajan, line 81: I don't think we should bother storing `max_rpm` here or in `SingleEncodedMotor`. These motors should just implement GoFor by calling it on the underlying motor

gvaradarajan (changes requested):
I agree with Nico's comments, and I have a couple additional requested changes. Thank you for making that change to `KeyNotFound`

src/common/math_utils.rs
mattjperez: now errors of either rpm or revolutions are NaN

src/esp32/motor.rs
mattjperez: removed the fields from encoded motor and defer to internal moto's go_for until we add the more advanced encoder logic

mattjperez:
Made the changes. Just need to verify on hardware.

mattjperez:
Conclusion: go_for requires awaiting without blocking the executor and this would require a large refactor.

src/common/math_utils.rs
```diff
@@ -27,3 +33,106 @@ impl From<Vector3> for common::v1::Vector3 {
...
+
+        let max_nan = go_for_math(f64::NAN, rpm, revolutions);
+        assert!(max_nan.is_err());
+        let rpm_nan = go_for_math(max_rpm, f64::NAN, revolutions);
+        assert!(rpm_nan.is_err());
+        let rev_nan = go_for_math(max_rpm, rpm, f64::NAN);
+        assert!(rev_nan.is_err());
+    }
+
+    // TODO: put real inputs and expected outcomes
```
npmenard, line 85: is this todo still relevant?

src/esp32/motor.rs
```diff
@@ -152,13 +158,18 @@ where
     PWM: PwmPin<Duty = u32>,
 {
     pub fn new(a: A, b: B, pwm: PWM) -> Self {
-        ABMotorEsp32 { a, b, pwm }
+        ABMotorEsp32 {
+            a,
+            b,
+            pwm,
+            max_rpm: 200.0,
```
npmenard, line 165: what is the rational for this default value?

src/esp32/motor.rs
```diff
@@ -214,12 +225,20 @@ where
         }
         self.pwm
             .set_duty(((max_duty as f64) * pct.abs()).floor() as u32);
-        debug!("set to {:?} pct", pct);
```
npmenard, line 217: leave it as debug

src/common/math_utils.rs
```diff
@@ -1,5 +1,11 @@
 #![allow(dead_code)]
 use crate::proto::common;
+use anyhow::bail;
+use std::time::Duration;
+
+pub enum MathUtilError {
+    No,
```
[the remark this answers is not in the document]
mattjperez: Ah, forgot to remove it. Was overthinking the returned Result

src/esp32/motor.rs
```diff
@@ -265,7 +285,12 @@ where
     PWM: PwmPin<Duty = u32>,
 {
     pub fn new(dir: DIR, pwm: PWM, dir_flip: bool) -> Self {
-        Self { dir, pwm, dir_flip }
+        Self {
+            dir,
+            pwm,
+            dir_flip,
+            max_rpm: f64::default(),
```
npmenard, line 292: what is the value of a f64 default? can you add a comment so it's easy to figure out?

src/esp32/motor.rs
```diff
@@ -291,11 +316,20 @@ where
         }
         self.pwm
             .set_duty(((max_duty as f64) * pct.abs()).floor() as u32);
+        info!("set to {:?} pct", pct);
```
npmenard, line 319: remove or debug

src/common/motor.rs
```diff
@@ -141,6 +171,15 @@ impl Motor for FakeMotor {
         self.power = pct;
         Ok(())
     }
+    fn go_for(&mut self, rpm: f64, revolutions: f64) -> anyhow::Result<Option<Duration>> {
+        // get_max_rpm
+        let (pwr, dur) = go_for_math(self.max_rpm, rpm, revolutions).unwrap();
+        self.set_power(pwr)?;
+        if dur.is_some() {
```
npmenard, line 178: if dur is alreay returned as an option can we just return Ok(dur)?

src/common/motor.rs
```diff
@@ -141,6 +171,15 @@ impl Motor for FakeMotor {
         self.power = pct;
         Ok(())
     }
+    fn go_for(&mut self, rpm: f64, revolutions: f64) -> anyhow::Result<Option<Duration>> {
+        // get_max_rpm
+        let (pwr, dur) = go_for_math(self.max_rpm, rpm, revolutions).unwrap();
```
npmenard, line 176: we shouldn't panic on error here.

npmenard (changes requested):
Couple of nits, looks pretty good

src/common/math_utils.rs
mattjperez: nope! ported the tests from the go rdk

npmenard (approved):
LGTM

PR #50, opened 2023-06-09
RSDK-3169 - pull component configs dynamically on server start

gvaradarajan:
Not actually ready for review, please hold


examples/sdkconfig.defaults
```diff
@@ -1,5 +1,5 @@
 # Rust often needs a bit of an extra main task stack size compared to C (the default is 3K)
-CONFIG_ESP_MAIN_TASK_STACK_SIZE=32768
+CONFIG_ESP_MAIN_TASK_STACK_SIZE=45768
```
npmenard, line 2: di you try to reduce it a little bit? I would like to limit stack usage as much as possible

src/common/app_client.rs
```diff
@@ -172,6 +172,59 @@ impl<'a> AppClientBuilder<'a> {
     }
 }
 
+pub fn fetch_config(
```
npmenard, line 175: there is a lot of code duplication here what was the rational behind it?

src/common/robot.rs
```diff
@@ -121,6 +133,105 @@ impl LocalRobot {
...
+        let b = match r {
+            Some(r) => {
+                let mut full_model = r.model.to_string();
+                if !full_model.starts_with("micro-rdk:builtin") {
+                    anyhow::bail!(
+                        "model name must be prefixed with 'micro-rdk:builtin', model name is {:?}",
+                        full_model
+                    );
+                }
+                let model = full_model.split_off(18);
```
npmenard, line 158: this would panic if there is an absence of model. Would use split or some other iterator base functions to extract the model. Or add a bound check

src/common/robot.rs
```diff
@@ -121,6 +133,105 @@ impl LocalRobot {
...
+            resources: ResourceMap::new(),
+        };
+
+        let conf = fetch_config(client, &config)?;
+        let components = &(conf.config.as_ref().unwrap().components);
+
+        let r = components.iter().find(|x| x.r#type == "board");
+        let b = match r {
+            Some(r) => {
+                let mut full_model = r.model.to_string();
```
npmenard, line 151: maybe put this logic in a helper function to avoid code duplication

npmenard:
Done with first pass. Looks pretty great!!! What kind of config test did you conduct? Did you see any issue with larger configs?

src/common/registry.rs
```diff
@@ -22,15 +22,15 @@ lazy_static::lazy_static! {
...
 impl fmt::Display for RegistryError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
-            RegistryError::ModelNotFound => {
-                write!(f, "RegistryError : Model not found")
+            RegistryError::ModelNotFound(model) => {
+                write!(f, "RegistryError : Model '{model}' not found")
             }
             RegistryError::ModelAlreadyRegistered(model) => {
                 write!(f, "RegistryError : model {model} already exists")
```
mattjperez, line 36: You might be interested in using the [`thiserror`](https://docs.rs/thiserror/latest/thiserror/) crate as an alternative to hand-rolling errors.

src/common/registry.rs
```diff
@@ -185,7 +185,61 @@ impl ComponentRegistry {
         if let Some(ctor) = self.encoders.get(model) {
             return Ok(*ctor);
         }
-        Err(RegistryError::ModelNotFound)
+        Err(RegistryError::ModelNotFound(model.to_string()))
+    }
+
+    pub(crate) fn get_motor_constructor_dyn(
```
mattjperez, line 191: are all these `_dyn` variations only different in that the `model` parameter is a String and not a slice?
If so, you should be able to pass a reference to the strings when calling the non-`dyn` versions. 

edit: I see now that they are taking static strings... might complicate things. Do we even need the `'static`?

mattjperez (commented):
Nice! I like how you pass `get_attribute()` through and remove the `match cfg {}` entirely.

Done with my first pass.

src/common/app_client.rs
gvaradarajan: I wanted to be able to do this outside of having to create an AppClientBuilder because it felt strange to create one of those but never construct an AppClient

npmenard: I see, I think it's fin to build an app client just to take ownership of the config and drop it afterward. then we can have the robot constructor be name new_from_config.

src/common/registry.rs
npmenard: would add a lifetime <'a> to the function solve the issue?

src/common/robot.rs
gvaradarajan: if there was no model, wouldn't it fail the condition above at L.152?

examples/esp32/esp32_webrtc_with_config.rs
```diff
@@ -0,0 +1,213 @@
...
+            ROBOT_DTLS_CERT_FP,
+        ));
+
+        let dtls = Esp32Dtls::new(webrtc_certificate.clone()).unwrap();
+
+        let webrtc = app_client
+            .connect_webrtc(webrtc_certificate, executor.clone(), dtls)
+            .unwrap();
+
+        drop(app_client);
```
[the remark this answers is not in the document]
gvaradarajan: yes, but in between there is a call to `run_server`, which starts up its own grpc client, so I was hoping to free some memory up in advance.

src/common/registry.rs
gvaradarajan: I have a different refactor in mind, which I've pushed up in [this commit](https://github.com/viamrobotics/micro-rdk/pull/50/commits/16749aa2afa4aa6114028ac08a1b1b82a6b1b620). I'd like to know what you both think.

src/common/registry.rs
gvaradarajan: yeah, it looks like a good option. Thanks for the suggestion!

src/common/app_client.rs
gvaradarajan: is [this](https://github.com/viamrobotics/micro-rdk/pull/50/commits/fac3f7161209cf69b8a422e0977efa4da68a8ec5) what you had in mind?

src/common/robot.rs
npmenard: If the model is empty would the string be micro_rdk:builtin:? Or maybe it can never be the case ^^

examples/esp32/esp32_webrtc_with_config.rs
```diff
@@ -0,0 +1,213 @@
+#[allow(dead_code)]
+#[cfg(not(feature = "qemu"))]
+const SSID: &str = env!("MICRO_RDK_WIFI_SSID");
+#[allow(dead_code)]
+#[cfg(not(feature = "qemu"))]
+const PASS: &str = env!("MICRO_RDK_WIFI_PASSWORD");
```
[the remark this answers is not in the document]
gvaradarajan: I think we still need it because we want it specific to those lines when not building the qemu feature? I'm not sure though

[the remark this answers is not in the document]
npmenard: I don't think we need to care too much about these different examples will be removed in a following PR

examples/sdkconfig.defaults
npmenard: ok

mattjperez (approved):
LGTM!

PR #54, opened 2023-06-21
RSDK-2322 - implement support for component dependencies in config

gvaradarajan:
This supports _implicit_ dependencies only; it does not make use of the 'depends_on' field (although this could be implemented in a later PR). I made this choice because it seems that we were headed in the direction of implicit rather than explicit dependencies in RDK anyway. The board is still always passed as a dependency to every component constructor.

Follow-up PRs (still under RSDK-2322) will implement dependency functions for the rest of the components for which such a thing is relevant.

### DIFF
--- a/src/common/registry.rs
+++ b/src/common/registry.rs
@@ -27,25 +27,86 @@ pub enum RegistryError {
     ModelNotFound(String),
     #[error("RegistryError : model '{0}' already exists")]
     ModelAlreadyRegistered(&'static str),
+    #[error("RegistryError: model '{0}' dependency getter already registered")]
+    ModelDependencyFuncRegistered(&'static str),
+    #[error("RegistryError: dependencies unsupported for component type '{0}'")]
+    ComponentTypeNotInDependencies(&'static str),
+    #[error("RegistryError: model '{0}' not found in dependencies under component type '{1}'")]
+    ModelNotFoundInDependencies(String, &'static str),
 }
 
 use std::collections::BTreeMap as Map;
 
+use crate::proto::common::v1::ResourceName;
+
 use super::{
     board::BoardType, config::ConfigType, encoder::EncoderType, motor::MotorType,
-    movement_sensor::MovementSensorType, sensor::SensorType,
+    movement_sensor::MovementSensorType, robot::Resource, sensor::SensorType,
 };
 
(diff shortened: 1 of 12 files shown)


src/common/registry.rs
```diff
@@ -27,44 +27,112 @@ pub enum RegistryError {
...
+}
 
-type MovementSensorConstructor =
-    dyn Fn(ConfigType, Option<BoardType>) -> anyhow::Result<MovementSensorType>;
+impl TryFrom<ResourceName> for ResourceKey {
+    type Error = anyhow::Error;
+    fn try_from(value: ResourceName) -> Result<Self, Self::Error> {
+        let comp_type: &str = &value.subtype;
+        let comp_name = match comp_type {
+            "motor" => crate::common::motor::COMPONENT_NAME,
```
npmenard, line 84: Is Board missing on purpose?

src/common/robot.rs
```diff
@@ -130,14 +143,18 @@ impl LocalRobot {
             resources: ResourceMap::new(),
         };
 
-        let components = &(config_resp.config.as_ref().unwrap().components);
+        let mut components = config_resp.config.as_ref().unwrap().components.clone();
```
npmenard, line 146: is there a way we can avoid cloning here? it doesn't look like you are keeping ownership of the cloned components

npmenard: I see now why you need it. Do you think we could write the algorithm so that the vector you allocate is a Vec<Option<&ComponentConfig>> this way as you can just take the inner value of the Option and put it back if it needs further processing. Otherwise the Vector will contently move things around in memory which can be problemactic for larger configs. Could also use a Map here. If this is too complicated then I think we should do a Vec<&ComponentConfig> so it takes less space in memory

npmenard (changes requested):
Done with first pass. I don't have major comments but I would like to see some more corner case tests (dependency cylcle, absence of dependencies etc...) If you have time today I'd like a run trough the algorithm to make sure I understood it correctly

src/common/registry.rs
gvaradarajan: since we always add the board first, I didn't end up needing it. But I can add it

npmenard (approved):
LGTM

PR #56, opened 2023-06-22
[RSDK-3737] bump dep versions

mattjperez:


gvaradarajan (approved):
LGTM, didn't notice anything worrying

Cargo.toml
```diff
@@ -12,7 +12,7 @@ version = "0.0.1"
 crate-type = ["lib"]
 
 [patch.crates-io]
-esp-idf-sys = {git = "https://github.com/npmenard/esp-idf-sys.git"}
+#esp-idf-sys = {git = "https://github.com/npmenard/esp-idf-sys.git"}
```
npmenard, line 15: I updated my fork of esp-idf-sys, can you try it?

examples/Cargo.toml
```diff
@@ -17,11 +17,11 @@ opt-level = "z"
 qemu = []
 
 [patch.crates-io]
-esp-idf-sys = {git = "https://github.com/npmenard/esp-idf-sys.git"}
+#esp-idf-sys = {git = "https://github.com/npmenard/esp-idf-sys.git"}
```
npmenard, line 20: see above

examples/Cargo.toml
```diff
@@ -17,11 +17,11 @@ opt-level = "z"
...
 polling = { git = "https://github.com/esp-rs-compat/polling" }
 smol = { git = "https://github.com/esp-rs-compat/smol" }
 socket2 = { git = "https://github.com/npmenard/socket2.git" }
-webrtc = {git = " https://github.com/webrtc-rs/webrtc"}
```
npmenard, line 24: you can just remove this line the bug was fixed upstream

examples/build.rs
```diff
@@ -182,6 +182,8 @@ fn main() -> anyhow::Result<()> {
     let robot_name = cloud_cfg.local_fqdn.split('.').next().unwrap_or("");
     let local_fqdn = cloud_cfg.local_fqdn.replace('.', "-");
     let fqdn = cloud_cfg.fqdn.replace('.', "-");
+    // let location_secret = cloud_cfg.location_secrets.first().unwrap_or("");
```
npmenard, line 185: why the commented lines?

examples/esp32/esp32-server.rs
```diff
@@ -7,30 +7,64 @@ const PASS: &str = env!("MICRO_RDK_WIFI_PASSWORD");
...
+
+#[allow(unused_imports)]
+use micro_rdk::esp32::{
+    certificate::WebRtcCertificate, dtls::Esp32Dtls, entry::serve_web, exec::Esp32Executor,
+    tcp::Esp32Stream, tls::Esp32Tls, tls::Esp32TlsServerConfig,
+};
+
+/* webhook
+use embedded_svc::{
+    http::{client::Client as HttpClient, Method, Status},
```
npmenard, line 37: should remove this

examples/esp32/esp32-server.rs
```diff
@@ -62,18 +91,35 @@ fn main() -> anyhow::Result<()> {
...
         info!("creating eth object");
-        let eth = eth_configure(
-            &sys_loop_stack,
-            Box::new(esp_idf_svc::eth::EspEth::wrap(EthDriver::new_openeth(
-                periph.mac,
-                sys_loop_stack.clone(),
-            )?)?),
+
+        let ip =  Ipv4Addr::new(10, 1, 12, 187);
+        let ip_configuration = embedded_svc::ipv4::Configuration::Client(ClientConfiguration::Fixed(ClientSettings{
```
npmenard, line 100: what does this code do?

examples/esp32/esp32-server.rs
```diff
@@ -86,14 +132,14 @@ fn main() -> anyhow::Result<()> {
...
-        (wifi.sta_netif().get_ip_info()?.ip, wifi)
+        (wifi.wifi().sta_netif().get_ip_info()?.ip, wifi)
     };
 
     let cfg = AppClientConfig::new(
         ROBOT_SECRET.to_owned(),
         ROBOT_ID.to_owned(),
         ip,
-        "".to_owned(),
+        "".to_string(),
```
npmenard, line 142: prefer to_owned

examples/esp32/esp32-server.rs
```diff
@@ -116,81 +162,61 @@ fn main() -> anyhow::Result<()> {
...
 
-    wifi.start()?;
+    wifi.set_configuration(&wifi_configuration)?;
 
-    if !WifiWait::new(&sl_stack)?
-        .wait_with_timeout(Duration::from_secs(20), || wifi.is_started().unwrap())
-    {
-        bail!("couldn't start wifi")
-    }
+    block_on( async {
```
npmenard, line 207: if they have a blocking version of the wifi driver why no use that? there are no upside for use to jump into an async context here

examples/esp32/partitions.csv
```diff
@@ -2,4 +2,4 @@
 # Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
 nvs,      data, nvs,     ,        0x6000,
 phy_init, data, phy,     ,        0x1000,
-factory2,  app,  factory, ,       0x360000,
```
npmenard, line 5: I would like to be conscious about the flash usage here since we want to reserve some for future use (see last commit in micro-rdk) Do you have an idea of the current size it takes with your updates?

examples/sdkconfig.defaults
```diff
@@ -1,5 +1,5 @@
 # Rust often needs a bit of an extra main task stack size compared to C (the default is 3K)
-CONFIG_ESP_MAIN_TASK_STACK_SIZE=18576
+CONFIG_ESP_MAIN_TASK_STACK_SIZE=32000
```
npmenard, line 2: I would like to minimize the stack usage any chance we can reduce it closer the the original value>

examples/esp32/esp32-server.rs
```diff
@@ -7,30 +7,64 @@ const PASS: &str = env!("MICRO_RDK_WIFI_PASSWORD");
...
+    utils::io,
+};
+*/
+
+// esp_idf_svc
+
+#[allow(unused_imports)]
+use esp_idf_svc::{
+    eventloop::EspSystemEventLoop,
+    eth::{EspEth, EthDriver},
```
npmenard, line 48: for the eth driver I would like to gate it behind the emu feature

npmenard (changes requested):
Done with first pass, let me know when you are ready for final review

examples/esp32/esp32-server.rs
mattjperez: this sets the ethernet's ip [configuration](https://docs.rs/embedded-svc/0.25.1/embedded_svc/ipv4/enum.Configuration.html) to be a client, with a [fixed](https://docs.rs/embedded-svc/0.25.1/embedded_svc/ipv4/enum.ClientConfiguration.html) ip instead of using dhcp. The original implementation had this IP address so I used it to keep close to the original code.

npmenard: I see, we don't need this the IP was just the qemu-host ip. This way we can connect through emu with http2 & wwebRTC

Cargo.toml
mattjperez: seems to be working without any issues.

examples/build.rs
mattjperez: webhook related. was trying to get the location secret that we decided to get via board attributes for now instead.

examples/esp32/partitions.csv
mattjperez: release: `App/part. size:    3,065,872/4,128,768 bytes, 74.26%`
debug: `App/part. size:    3,964,080/4,128,768 bytes, 96.01%`

Cargo.toml
mattjperez: maybe not...
getting this error now when running it, need to look into it further. 
```
thread 'main' panicked at 'unsafe precondition(s) violated: ptr::read requires that the pointer argument is aligned and non-null', ~/.rustup/toolchains/esp/lib/rustlib/src/rust/library/core/src/panicking.rs:126:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
thread caused non-unwinding panic. aborting.

abort() was called at PC 0x4026cc7a on core 1
```

examples/esp32/partitions.csv
mattjperez: When using the original partitions

```
Partition table:   esp32/partitions.csv
Error: espflash::image_too_big (link)

  × Supplied ELF image of 3964080B is too big, and doesn't fit configured app partition of 3538944B
  help: Reduce the size of the binary or increase the size of the app partition.
```

examples/sdkconfig.defaults
mattjperez: currently at 2400, at original value was experiencing stack overflows

examples/esp32/partitions.csv
mattjperez: increased partition size slightly over original to fit code with webhook implementation.

etc/docker/Dockerfile
```diff
@@ -6,8 +6,8 @@ ENV DEBCONF_NOWARNINGS="yes"
 ENV RUSTUP_HOME=/opt/rust/rustup \
     CARGO_HOME=/opt/rust/cargo \
     PATH=/opt/rust/cargo/bin:$PATH \
-    RUST_VERSION=1.65.0 \
-    ESPRS_VERSION=v1.65.0.0 \
+    RUST_VERSION=1.70.0 \
+    ESPRS_VERSION=v1.66.0.0 \
```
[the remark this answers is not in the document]
mattjperez: changed to version 1.67 to successfully compile with the newer `esp_idf_*` and `embedded_*` dependencies

npmenard (approved):
cool 😎 LGTM

PR #59, opened 2023-06-28
[RSDK-3217] initial webhook implementation

mattjperez:


src/esp32/webhook.rs
```diff
@@ -0,0 +1,94 @@
...
+    let board_cfg: DynamicComponentConfig = components
+        .iter()
+        .find(|x| x.r#type == "board")
+        .ok_or_else(|| {
+            WebhookError::ConfigError("board component not found in config".to_string())
+        })?
+        .try_into()
+        .map_err(|_| {
+            WebhookError::ConfigError(
+                "could not convert board to DynamicComponentConfig".to_string(),
```
gvaradarajan, line 47: this should really say "could not convert board config to DynamicComponentConfig". Since we're not dealing with an actual board at this stage we want to avoid confusion in the wording

src/esp32/webhook.rs
```diff
@@ -0,0 +1,94 @@
...
+/// Checks the board config for a `webhook` and `webhook-secret` attribute, sending a GET
+/// request with credentials for an SDK to make a connection.
+pub fn handle_webhook(config: ConfigResponse) -> Result<(), WebhookError> {
+    let config = config
+        .config
+        .as_ref()
+        .ok_or_else(|| WebhookError::ConfigError("board does not have config".to_string()))?;
+    let components = &config.components; // component config
+    let cloud = config.cloud.as_ref().ok_or_else(|| {
+        WebhookError::ConfigError("board config does not have cloud config".to_string())
```
gvaradarajan, line 35: this isn't the "board config" at this point, I think you meant "robot config" (also applies above on lines 26 and 32)

src/esp32/webhook.rs
```diff
@@ -0,0 +1,94 @@
...
+    let webhook = board_cfg
+        .get_attribute::<String>("webhook")
+        .map_err(|_| WebhookError::NoWebhook)?;
+    let secret = board_cfg
+        .get_attribute::<String>("webhook-secret")
+        .map_err(|_| WebhookError::NoWebhookSecret)?;
+    let payload = json!({
+        "location": fqdn,
+        "secret": secret,
+        "board": board_cfg.name,
```
gvaradarajan, line 60: why do we send the board name here?

src/esp32/webhook.rs
```diff
@@ -0,0 +1,94 @@
...
+    #[error("Error with webhook request: {0}")]
+    RequestError(String),
+    #[error("Error in config: {0}")]
+    ConfigError(String),
+}
+
+/// Checks the board config for a `webhook` and `webhook-secret` attribute, sending a GET
+/// request with credentials for an SDK to make a connection.
+pub fn handle_webhook(config: ConfigResponse) -> Result<(), WebhookError> {
+    let config = config
```
gvaradarajan, line 29: I think this should be handled on the outside and this function should take an instance of `RobotConfig` instead of `ConfigResponse` (I know I've done essentially the same thing in `LocalRobot::new_from_config_response` but I think that was a mistake)

src/esp32/entry.rs
```diff
@@ -76,8 +77,11 @@ pub fn serve_web(
                     .unwrap(),
             ),
             robot,
+            robot_cfg,
         )
     };
 
+    let _ = handle_webhook(*robot_cfg); // TODO: add retry logic
```
gvaradarajan, line 84: log::error! in the case of an error?

src/esp32/webhook.rs
```diff
@@ -0,0 +1,94 @@
...
+    NoWebhookSecret,
+    #[error("Error with webhook request: {0}")]
+    RequestError(String),
+    #[error("Error in config: {0}")]
+    ConfigError(String),
+}
+
+/// Checks the board config for a `webhook` and `webhook-secret` attribute, sending a GET
+/// request with credentials for an SDK to make a connection.
+pub fn handle_webhook(config: ConfigResponse) -> Result<(), WebhookError> {
```
gvaradarajan, line 28: I'd like to see this function split into two functions doing the following:
1) One that gets the webhook and secret from a `ConfigResponse`
2) One that sends the GET request to the webhook
I think doing this also means that the config could be dropped earlier

(Optional: I think having a struct that represents a Webhook and contains its secret and cloud fqdn would be nice. 1) above could be a constructor that takes in a `ConfigResponse` or `RobotConfig` instance and 2) would just be an instance method on the webhook)

gvaradarajan (changes requested):

src/esp32/webhook.rs
mattjperez: removed :)

src/esp32/webhook.rs
mattjperez: done

examples/sdkconfig.defaults
```diff
@@ -1,5 +1,5 @@
 # Rust often needs a bit of an extra main task stack size compared to C (the default is 3K)
-CONFIG_ESP_MAIN_TASK_STACK_SIZE=24000
+CONFIG_ESP_MAIN_TASK_STACK_SIZE=30000
```
npmenard, line 2: Do we still need the increase?

src/esp32/entry.rs
```diff
@@ -76,8 +77,14 @@ pub fn serve_web(
                     .unwrap(),
             ),
             robot,
+            robot_cfg.as_ref().config.as_ref().unwrap().clone(),
         )
     };
 
+    let r = handle_webhook(robot_cfg); // TODO: add retry logic
```
npmenard, line 84: let's move that in the upper block so it's clear we drop both the webhook & the config

src/esp32/entry.rs
```diff
@@ -76,8 +77,14 @@ pub fn serve_web(
...
+    if r.is_err() {
+        log::error!("{:?}", r);
```
npmenard, line 86: nit: webhook error {:?}

src/esp32/webhook.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        let mut request = client
+            .request(Method::Get, &self.endpoint, &headers)
+            .map_err(|_| WebhookError::RequestError)?;
+
+        request
+            .write_all(payload)
+            .map_err(|_| WebhookError::RequestError)?;
+
+        request.flush().map_err(|_| WebhookError::RequestError)?;
```
npmenard, line 123: would alos wrap the returned error so it's a bit more expressive

src/esp32/webhook.rs
```diff
@@ -0,0 +1,128 @@
...
+        let components = config.components;
+        let cloud = config.cloud.as_ref().ok_or_else(|| {
+            WebhookError::ConfigError("robot config does not have cloud config".to_string())
+        })?;
+        let fqdn = cloud.fqdn.clone();
+        let board_cfg: DynamicComponentConfig = components
+            .iter()
+            .find(|x| x.r#type == "board")
+            .ok_or_else(|| {
+                WebhookError::ConfigError("board component not found in robot config".to_string())
```
npmenard, line 73: this should not be an error, we assume that "in theory" a robot can start with an empty config

src/esp32/webhook.rs
```diff
@@ -0,0 +1,128 @@
...
+            })?
+            .try_into()
+            .map_err(|_| {
+                WebhookError::ConfigError(
+                    "could not convert board config to DynamicComponentConfig".to_string(),
+                )
+            })?;
+
+        let endpoint = board_cfg
+            .get_attribute::<String>("webhook")
```
npmenard, line 83: I think the logic should be to have a silent failure if the key is absent and return an error if we couldn't convert into a String

src/esp32/webhook.rs
```diff
@@ -0,0 +1,128 @@
...
+            .map_err(|_| WebhookError::NoWebhook)?;
+        let secret = board_cfg
+            .get_attribute::<String>("webhook-secret")
+            .map_err(|_| WebhookError::NoWebhookSecret)?;
```
npmenard, line 87: would wrap the AtrributError so it's more verbose

src/esp32/webhook.rs
npmenard: also check if the string is a valid URL

npmenard (changes requested):
Done with first pass, this is pretty good. Would add more logic so not adding a webhook in the config doesn't return an error but just silently do nothing

src/esp32/webhook.rs
mattjperez: Was having a lot of issues trying to wrap `embedded_svc::io::Error` with `thiserror` so went with just converting the underlying error to a String and wrapping that.

src/esp32/webhook.rs
```diff
@@ -0,0 +1,192 @@
...
+};
+use esp_idf_svc::http::client::{Configuration as HttpConfiguration, EspHttpConnection};
+use log::*;
+use serde_json::json;
+use thiserror::Error;
+use url::{ParseError, Url};
+
+/// Checks the board config for a `webhook` and `webhook-secret` attribute, sending a GET
+/// request with credentials for an SDK to make a connection.
+pub fn handle_webhook(config: &RobotConfig) -> Result<(), WebhookError> {
```
gvaradarajan, line 19: could we maybe move this to the impl for webhook and have the code in `entry.rs` responsible for calling the webhook constructor and this function?

gvaradarajan (approved):
LGTM. Had one little suggestion for an additional improvement, but I'm not 100% sure it makes things better so I figured it would just be optional

src/esp32/webhook.rs
mattjperez: yeah I was thinking about that as well.

examples/sdkconfig.defaults
mattjperez: verified working with 24000 and on latest commit

npmenard (approved):
LGTM

PR #60, opened 2023-06-29
RSDK-2322 - support configuration for PWM-Direction pin motors

gvaradarajan:


src/esp32/motor.rs
```diff
@@ -328,9 +358,40 @@ where
...
+                    PwmChannel::C1(c) => LedcDriver::new(c, timer, pwm_pin)?,
+                    PwmChannel::C2(c) => LedcDriver::new(c, timer, pwm_pin)?,
+                };
+                let max_rpm: f64 = cfg.get_attribute::<f64>("max_rpm")?;
+                let dir_flip: bool = if let Ok(dir_flip) = cfg.get_attribute::<bool>("dir_flip") {
+                    dir_flip
+                } else {
+                    false
+                };
+                return Ok(Arc::new(Mutex::new(PwmDirectionMotorEsp32::new(
```
mattjperez, line 382: nit: `Self` suggestion

mattjperez (changes requested):
Done with first pass. Nice! I like how you implemented the `gpio_motor_from_config`. As much I like turbo fish, I wonder if this might become a bit unreadable if/when we add more motor types.

src/common/motor.rs
```diff
@@ -75,19 +92,48 @@ pub struct FakeMotor {
...
+            },
+        };
+
+        let dir = match value.get("dir") {
+            Ok(opt) => match opt {
+                Some(val) => Some(val.try_into()?),
+                None => None,
+            },
+            Err(err) => match err {
+                AttributeError::KeyNotFound(_) => None,
```
npmenard, line 127: reading this makes me realize maybe KeyNotFound should disappear and we should return a None instead :P

npmenard (approved):
LGTM

src/esp32/motor.rs
```diff
@@ -193,8 +209,14 @@ where
     B: OutputPin + PinExt,
     PWM: PwmPin<Duty = u32>,
 {
-    pub fn new(a: A, b: B, pwm: PWM, max_rpm: f64) -> Self {
-        ABMotorEsp32 { a, b, pwm, max_rpm }
+    pub fn new(a: A, b: B, pwm: PWM, max_rpm: f64, dir_flip: bool) -> Self {
+        ABMotorEsp32 {
```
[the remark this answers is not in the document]
gvaradarajan: so making this change (and the similar change to the other motor) caused the compiler to fail. I think it has to do with the fact that it's an implementation of a generic? I'm looking into it

mattjperez (approved):

PR #61, opened 2023-06-30
RSDK-3171 - remove special prefix requirement and fix string parsing bug

gvaradarajan:
This commit results in the following UX changes:

- prefixing the model name with "micro-rdk:builtin:" is no longer required when configuring a component
- the config cards for `gpio` motor, as well as for `accel-adxl345` and `gyro-mpu6050`, in the app UI can also now be used


mattjperez (approved):
LGTM!

npmenard (approved):

PR #62, opened 2023-07-10
RSDK-3857

npmenard:
ESP-TLS doesn't let us easily chain DER certificates so for now only the key & the CA are stored in the DER format while the location certificate and it's intermediate certificate remain in PEM format


src/esp32/tls.rs
```diff
@@ -141,20 +134,19 @@ impl Esp32Tls {
...
-                cacert_bytes: ca_cert.len() as u32,
+                cacert_bytes: cfg.srv_cert[1].len() as u32,
             },
             __bindgen_anon_3: esp_idf_sys::esp_tls_cfg_server__bindgen_ty_3 {
-                servercert_buf: cfg.srv_cert,
+                servercert_buf: cfg.srv_cert[0].as_ptr(),
             },
             __bindgen_anon_4: esp_idf_sys::esp_tls_cfg_server__bindgen_ty_4 {
-                servercert_bytes: cfg.srv_cert_len,
+                servercert_bytes: cfg.srv_cert[0].len() as u32,
```
mattjperez, line 149: can we have comments here about what exactly is in `srv_cert[0]` and `srv_cert[1]`? 
I can't tell how many items are (or should be) in `cfg.srv_cert` from `Esp32TlsServerConfig`

mattjperez (approved):
Aside from adding some clarifying comments, LGTM

gvaradarajan (approved):
LGTM

PR #64, opened 2023-07-12
RSDK-3956 - update buf plugin

gvaradarajan:
Our previously used plugin for buf has been deprecated, making this update necessary. All changes have been auto-generated as a consequence (apart from the changes in `Cargo.toml` and `buf.gen.yaml`)


npmenard (approved):
LGTM

gvaradarajan:
It turns out we cannot make this update at all until [this](https://github.com/neoeinstein/protoc-gen-prost/issues/61) is fixed. There are workaround options, but since we don't need proto-generated code to updated for a while we can wait and see if the issue is resolved by then

gvaradarajan:
We found a workaround, re-opening

PR #67, opened 2023-07-25
update examples dependencies

mattjperez:


npmenard (changes requested):
can you do >= 0.0.25 ? that way we can alleviate future issues with a quick cargo update

npmenard (approved):

PR #69, opened 2023-08-09
RSDK-3822 - installer CLI part 1: generate NVS partition

gvaradarajan:


micro-rdk-installer/README.md
```diff
@@ -0,0 +1,23 @@
...
+Storage (NVS) Partition and containing wifi and robot credentials. One could then potentially flash just
+the NVS partition to their esp32 using tools provided by ESP-IDF. 
+
+## Testing NVS
+
+To test this functionality, download a json
+file representing the robot's app config from the Setup tab of the robot part's page on app.viam.com and then
+run the following command from within this subdirectory:
+```
+cargo run -- create-nvs-partition --app-config=<path to config json> --file-dest=<destination path for resulting binary>
```
mattjperez, line 17: nit. I think `--output`, shorthand  `-o`, is a more conventional

micro-rdk-installer/src/main.rs
```diff
@@ -0,0 +1,96 @@
...
+use micro_rdk_installer::nvs::data::{ViamFlashStorageData, WifiCredentials};
+use micro_rdk_installer::nvs::partition::{NVSPartition, NVSPartitionData};
+use micro_rdk_installer::nvs::request::populate_nvs_storage_from_app;
+use serde::{Deserialize, Serialize};
+
+#[derive(Serialize, Deserialize, Debug)]
+struct AppCloudConfig {
+    r#id: String,
+    app_address: String,
+    secret: String,
```
mattjperez, line 14: might consider using something like (secrecy)[https://crates.io/crates/secrecy] to keep debug and other outputs from printing out secrets accidentally.

gvaradarajan: good catch, this is a very necessary change

npmenard (changes requested):
Done with first pass

micro-rdk-installer/src/nvs/request.rs
```diff
@@ -0,0 +1,146 @@
...
+    param.alg = &rcgen::PKCS_ECDSA_P256_SHA256;
+
+    let kp = rcgen::KeyPair::generate(&rcgen::PKCS_ECDSA_P256_SHA256)?;
+    let kp_der = kp.serialize_der();
+    storage_data.robot_credentials.robot_dtls_key_pair = Some(kp_der);
+
+    param.key_pair = Some(kp);
+
+    let cert = rcgen::Certificate::from_params(param).unwrap();
+    let cert_der = cert.serialize_der().unwrap();
```
mattjperez, line 43: adding some context with `.expect()` or mapping to another error would help us troubleshoot any user issues stemming from this part of the process.

micro-rdk-installer/src/nvs/request.rs
```diff
@@ -0,0 +1,146 @@
...
+        .tls_certificate
+        .split_inclusive("----END CERTIFICATE-----");
+    let srv_cert = &mut tls_cert.clone().take(2).collect::<String>();
+    srv_cert.push('\0');
+    let ca_cert = &mut tls_cert
+        .clone()
+        .take(1)
+        .map(der::Document::from_pem)
+        .filter(|s| s.is_ok())
+        .map(|s| s.unwrap().1.to_vec())
```
mattjperez, line 137: might be interested in [`filter_map()`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map)

micro-rdk-installer/src/nvs/request.rs
```diff
@@ -0,0 +1,146 @@
...
+the Viam App and preparing for flash storage
+*/
+
+pub fn populate_nvs_storage_from_app(
+    storage_data: &mut ViamFlashStorageData,
+) -> anyhow::Result<()> {
+    populate_dtls_certificate(storage_data)?;
+    let rt = Runtime::new()?;
+    rt.block_on(read_cloud_config(storage_data))?;
+    rt.block_on(read_certificates(storage_data))?;
```
mattjperez, line 23: nit: these functions are called `read_*`, but don't actually return anything. maybe rename to `set_*` or `process_*`?

mattjperez (commented):
done with first pass.

Nice work!

micro-rdk-installer/src/nvs/partition.rs
```diff
@@ -0,0 +1,408 @@
...
+        header[24..28].copy_from_slice(&data_len_bytes);
+        header[28] = num_chunks as u8;
+        header[29] = 0;
+        set_header_crc(&mut header);
+        header
+    }
+}
+
+pub enum NVSValue {
+    String(String),
```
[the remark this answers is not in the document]
gvaradarajan: you would need somewhat different write logic for any primitive data, but it shouldn't be very difficult

micro-rdk-installer/src/nvs/partition.rs
```diff
@@ -0,0 +1,408 @@
+use std::collections::VecDeque;
+
+use crc32fast::Hasher;
+
+use super::data::ViamFlashStorageData;
+
+const MAX_BLOB_SIZE: usize = 4000;
+const BLOB_DATA_FORMAT: u8 = 0x42;
+const STRING_VALUE_FORMAT: u8 = 0x21;
+const BLOB_IDX_FORMAT: u8 = 0x48;
```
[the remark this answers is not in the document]
gvaradarajan: If I do that, won't these values default to `isize` rather than `u8`? Is this actually better?

micro-rdk-installer/src/nvs/partition.rs
```diff
@@ -0,0 +1,408 @@
...
+        }
+        self.sections[self.current_section].close();
+        let num_sections = self.sections.len();
+        println!("num sections: {:?}", num_sections);
+        for _ in 0..num_sections {
+            let section = self.sections.pop_front().unwrap();
+            res.append(&mut section.data.to_vec());
+        }
+        let mut reserved_section = std::iter::repeat(0xFF).take(4096).collect::<Vec<u8>>();
+        res.append(&mut reserved_section);
```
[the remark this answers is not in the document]
gvaradarajan: don't mind using chain, but I think this code is actually clearer as two lines

micro-rdk-installer/src/nvs/partition.rs
```diff
@@ -0,0 +1,408 @@
...
+An NVS Partition is sectioned into Pages and each Page has 32-byte entries. Every key-value pair
+stored in NVS consists of a 32-byte header entry (which contains the key and other metadata) followed
+by data split into however many entries are required based on length. Since the above credentials data
+consists only of string or binary values, only those two entry formats have been implemented
+
+More information on the structure of NVS and its API can be found in Espressif's online documentation
+(https://docs.espressif.com/projects/esp-idf/en/release-v4.4/esp32/api-reference/storage/nvs_flash.html)
+*/
+
+// computes the checksum of the contents of the header and stores it at index 4
```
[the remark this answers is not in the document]
gvaradarajan: I'm not sure I agree, this is a helper function that I don't think would ever become public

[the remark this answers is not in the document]
mattjperez: True. Was moreso docs for us than end-users.

micro-rdk-installer/src/nvs/partition.rs
```diff
@@ -359,26 +383,29 @@ impl NVSPartitionData {
...
     pub fn to_bytes(&mut self) -> Vec<u8> {
         let mut res = vec![];
-        let total_sections = 32768 / 4096 - 1;
+        let total_sections = self.size / 4096 - 1;
         let empty_sections = total_sections - self.sections.len();
         for _ in 0..empty_sections {
-            self.start_new_section();
+            match self.start_new_section() {
+                Ok(_) => {}
+                Err(_) => unreachable!(),
```
npmenard, line 403: if err can never happen unwrap should be fine.

npmenard (approved):
LGTM, if you are not going to use thiserror (totally fine) add a ticket so we can track this.

mattjperez (approved):
LGTM, definitely better with the new error handling

PR #70, opened 2023-08-15
RSDK-4499 - add digital interrupt support to board

gvaradarajan:
Also contains a small refactor on board that goes part of the way to improving management of pin resources on the ESP32 across components (see `GPIOPin` struct)

Hardware Test Cases Conducted
- getting/setting level for a pin not configured as an interrupt
- getting level for a pin configured as an interrupt
- calling `GetDigitalInterruptValue` on the board from a Python SDK for a pin configured as an interrupt (touched the pin to a 3v3 connection repeatedly to increase the event count)

I'm willing to remove the pins subscription support for other drivers if we don't like that I used channels. I felt it might be ok because our cargo dependencies implied that we were already implicitly using them

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ crate-type = ["lib"]
 
 [patch.crates-io]
 esp-idf-sys = {git = "https://github.com/npmenard/esp-idf-sys.git"}
-polling = { git = "https://github.com/esp-rs-compat/polling" }
 smol = { git = "https://github.com/esp-rs-compat/smol" }
 socket2 = { git = "https://github.com/npmenard/socket2.git" }
 async-io = {git = "https://github.com/npmenard/async-io.git"}
@@ -73,6 +72,7 @@ async-std-openssl = {version = "0.6.3", optional = true }
 thiserror = "1.0.40"
 pin-project-lite = "0.2.9"
 url = "2.4.0"
+once_cell = "1.18.0"
 
 [build-dependencies]
 anyhow = "1"
--- a/src/common/digital_interrupt.rs
+++ b/src/common/digital_interrupt.rs
@@ -1,0 +1,28 @@
+use super::config::{AttributeError, Kind};
+
+#[derive(Copy, Clone, Debug)]
+pub struct DigitalInterruptConfig {
+    pub pin: i32,
+}
+
+impl TryFrom<Kind> for DigitalInterruptConfig {
+    type Error = AttributeError;
+    fn try_from(value: Kind) -> Result<Self, Self::Error> {
+        if !value.contains_key("pin")? {
+            return Err(AttributeError::KeyNotFound("pin".to_string()));
+        }
+        let pin = value.get("pin")?.unwrap().try_into()?;
+        Ok(DigitalInterruptConfig { pin })
+    }
+}
+
+impl TryFrom<&Kind> for DigitalInterruptConfig {
+    type Error = AttributeError;
+    fn try_from(value: &Kind) -> Result<Self, Self::Error> {
+        if !value.contains_key("pin")? {
+            return Err(AttributeError::KeyNotFound("pin".to_string()));
+        }
+        let pin = value.get("pin")?.unwrap().try_into()?;
+        Ok(DigitalInterruptConfig { pin })
+    }
+}
--- a/src/esp32/board.rs
+++ b/src/esp32/board.rs
@@ -4,6 +4,7 @@ use crate::common::analog::AnalogReaderConfig;
 use crate::common::board::Board;
 use crate::common::board::BoardType;
 use crate::common::config::ConfigType;
+use crate::common::digital_interrupt::DigitalInterruptConfig;
 use crate::common::i2c::I2cHandleType;
 use crate::common::registry::ComponentRegistry;
 use crate::common::status::Status;
@@ -16,10 +17,13 @@ use esp_idf_hal::adc::AdcChannelDriver;
 use esp_idf_hal::adc::AdcDriver;
 use esp_idf_hal::adc::Atten11dB;
 use esp_idf_hal::adc::ADC1;
-use esp_idf_hal::gpio::{AnyIOPin, InputOutput, PinDriver};
+use esp_idf_hal::gpio::{AnyIOPin, InputOutput, InterruptType, PinDriver, Pull};
+use esp_idf_sys::{esp, gpio_install_isr_service, gpio_isr_handler_add, ESP_INTR_FLAG_IRAM};
 use log::*;
+use once_cell::sync::{Lazy, OnceCell};
 use std::collections::{BTreeMap, HashMap};
 use std::rc::Rc;
+use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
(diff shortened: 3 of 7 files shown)


src/common/digital_interrupt.rs
```diff
@@ -0,0 +1,126 @@
...
+            broadcasters: vec![],
+            intr_type,
+            previously_high: initially_high,
+            in_use: Arc::new(AtomicBool::new(false)),
+            event_count: Arc::new(AtomicI64::new(0)),
+        }
+    }
+
+    pub fn emit_event(&mut self) -> anyhow::Result<()> {
+        self.in_use.store(true, Ordering::Release);
```
npmenard, line 68: - Does it protect against multiple access if you don't check it? Also is there scenario where concurrent emit_event can happen?

src/common/digital_interrupt.rs
```diff
@@ -0,0 +1,126 @@
...
+        self.event_count.fetch_add(1, Ordering::Relaxed);
+        let event = self.create_event();
+        for broadcaster in self.broadcasters.iter_mut() {
+            broadcaster.send(event)?;
```
npmenard, line 72: the type is copy so you'll duplicate it for each send is that intended?

src/esp32/board.rs
```diff
@@ -35,15 +47,193 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
     }
 }
 
+lazy_static::lazy_static! {
+    static ref GPIO_ISR_SERVICE_INSTALLED: Arc<AtomicBool> = Arc::new(AtomicBool::new(false));
+}
+
+fn install_gpio_isr_service() -> anyhow::Result<()> {
+    if !GPIO_ISR_SERVICE_INSTALLED.fetch_or(true, Ordering::SeqCst) {
```
npmenard, line 55: using once_cell would be better here

Makefile
```diff
@@ -47,7 +47,7 @@ endif
 	cd examples && $(QEMU_ESP32_XTENSA)/qemu-system-xtensa -nographic -machine esp32 -gdb tcp::3334 -nic user,model=open_eth,hostfwd=udp::-:61205 -drive file=target/xtensa-esp32-espidf/debug/debug.bin,if=mtd,format=raw -S
 
 upload: cargo-ver
-	cd examples && cargo espflash flash --monitor --partition-table esp32/partitions.csv --baud 460800 -f 80M --use-stub --bin esp32-server --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort
+	cd examples && cargo espflash flash --monitor --partition-table esp32/partitions.csv --baud 460800 -f 80M --use-stub --bin esp32-server --target=xtensa-esp32-espidf --release -Zbuild-std=std,panic_abort
```
npmenard, line 50: why add the release flag here?

npmenard (changes requested):
Done with first pass, not a fan for using channels to send notifications. They introduce a lot of overhead and using them within an interrupt context will likely incur  a performance hit. I think we should keep it simple and just use a counter. If we really like the timestamp then we can use a stack or queue of the last 2 interrupts since they are the most important

Cargo.toml
```diff
@@ -14,7 +14,6 @@ crate-type = ["lib"]
 
 [patch.crates-io]
 esp-idf-sys = {git = "https://github.com/npmenard/esp-idf-sys.git"}
-polling = { git = "https://github.com/esp-rs-compat/polling" }
```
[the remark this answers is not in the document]
gvaradarajan: Uh, I didn't mean to. Let me double-check

src/common/digital_interrupt.rs
gvaradarajan: Yes this is intended

Makefile
gvaradarajan: I needed to use the release version to get it to fit

src/common/digital_interrupt.rs
```diff
@@ -0,0 +1,126 @@
...
+        self.timestamp
+    }
+}
+
+pub(crate) struct PinEventTransmitter {
+    broadcasters: Vec<Sender<InterruptEvent>>,
+    intr_type: InterruptEventType,
+    previously_high: bool,
+    in_use: Arc<AtomicBool>,
+    event_count: Arc<AtomicI64>,
```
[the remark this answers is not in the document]
gvaradarajan: When would the event count be negative?

src/esp32/board.rs
```diff
@@ -35,15 +47,193 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
+    config: gpio_config_t,
+    transmitter: Option<Box<PinEventTransmitter>>,
+    interrupt_type: Option<InterruptEventType>,
+}
+
+impl GPIOPin {
+    pub fn new(
+        pin: i32,
+        config: Option<gpio_config_t>,
+        pull_up: Option<bool>,
```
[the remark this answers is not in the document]
gvaradarajan: it does those actions through the PinDriver, or did I misunderstand the question?

src/esp32/board.rs
```diff
@@ -35,15 +47,193 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
+    ) -> anyhow::Result<Self> {
+        let pull_up_en = pull_up.unwrap_or_default();
+        let config = config;
+        let config = match config {
```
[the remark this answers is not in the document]
gvaradarajan: would we get that check if we moved L.129 up?

mattjperez (commented):
done with first pass.

I like the idea of wrapping/converting hal-specific GpioPins to our own representation.

src/common/digital_interrupt.rs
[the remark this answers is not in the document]
gvaradarajan: oh I swapped the two in my mind

src/esp32/board.rs
```diff
@@ -35,15 +39,119 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
+    #[link_section = ".iram1.intr_srv"]
+    unsafe extern "C" fn interrupt(arg: *mut core::ffi::c_void) {
+        let arg: &mut Arc<AtomicU32> = &mut *(arg as *mut _);
+        arg.fetch_add(1, Ordering::Relaxed);
+    }
+}
+
 pub struct EspBoard {
-    pins: Vec<PinDriver<'static, AnyIOPin, InputOutput>>,
+    // pins: Vec<PinDriver<'static, AnyIOPin, InputOutput>>,
```
npmenard, line 145: stray comment

src/esp32/board.rs
```diff
@@ -35,15 +39,119 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
+    pins: Vec<GPIOPin>,
     analogs: Vec<Rc<RefCell<dyn AnalogReader<u16, Error = anyhow::Error>>>>,
     i2cs: HashMap<String, I2cHandleType>,
 }
 
 impl EspBoard {
     pub fn new(
-        pins: Vec<PinDriver<'static, AnyIOPin, InputOutput>>,
+        // pins: Vec<PinDriver<'static, AnyIOPin, InputOutput>>,
```
npmenard, line 153: stray comment

src/esp32/board.rs
```diff
@@ -221,6 +329,21 @@ impl EspBoard {
             let i2c_wrapped: I2cHandleType = Arc::new(Mutex::new(i2c));
             i2cs.insert(name.to_string(), i2c_wrapped);
         }
+        if let Ok(interrupt_confs) =
+            cfg.get_attribute::<Vec<DigitalInterruptConfig>>("digital_interrupts")
+        {
+            for conf in interrupt_confs {
+                let p = pins.iter_mut().find(|p| p.pin() == conf.pin);
+                if let Some(p) = p {
+                    // TODO: make event type configurable
```
npmenard, line 338: nit: add a ticket and link the number

src/esp32/board.rs
```diff
@@ -35,15 +39,119 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
+            None => {
+                self.interrupt_type = Some(intr_type);
+            }
+        };
+        install_gpio_isr_service()?;
+        self.driver.set_interrupt_type(intr_type)?;
+        self.event_count.store(0, Ordering::Relaxed);
+        unsafe {
+            // we can't use the subscribe method on PinDriver to add the handler
+            // because it requires an FnMut with a static lifetime. A possible follow-up
```
npmenard, line 120: We should keep that in mind for when we are going to redesign the Board interface

src/esp32/board.rs
```diff
@@ -35,15 +39,119 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
+        Ok::<(), anyhow::Error>(())
+    })?;
+    Ok(())
+}
+
+/// GPIOPin is a wrapper for a pin on ESP32 as represented in esp-idf-hal
+/// and esp-idf-sys. This exists so that all micro-RDK drivers can interact
+/// with pins through the board instance and avoid conflicting uses of pins
+/// by multiple processes
+pub struct GPIOPin {
```
npmenard, line 58: nit: Esp32GPIOPin

src/esp32/board.rs
```diff
@@ -35,15 +39,119 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
+    driver: PinDriver<'static, AnyIOPin, InputOutput>,
+    pull: Pull,
+    interrupt_type: Option<InterruptType>,
+    event_count: Arc<AtomicU32>,
+}
+
+impl GPIOPin {
+    pub fn new(pin: i32, pull: Option<Pull>) -> anyhow::Result<Self> {
+        let mut driver = PinDriver::input_output(unsafe { AnyIOPin::new(pin) })?;
+        let pull = pull.unwrap_or(Pull::Up);
```
npmenard, line 69: this wouldn't get called when setting up I2C on the esp32?

src/esp32/board.rs
```diff
@@ -35,15 +39,119 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
+        driver.set_pull(pull)?;
```
npmenard, line 70: I am a bit concerned by setting the default drive mode to Pull:Up. Is it the default setup in the esp32 docs? Did you try to read a pin state after setting up as an interrupt? Does that affect setting the pin high or low?

npmenard (approved):
LGTM couple of nits and some minor concerns

src/esp32/board.rs
gvaradarajan: I was following the lead of what RDK does for pi, but Alan has confirmed that genericlinux boards do not do this. I will skip setting the pull up when pull is None

src/esp32/board.rs
gvaradarajan: as mentioned below, I will change this so it does not pull the pin at all in the default behavior

PR #71, opened 2023-08-24
RSDK-4251 - installer CLI part 2: implement option to edit binary with NVS data

gvaradarajan:
Also includes some fixes to NVS data generation


micro-rdk-installer/src/main.rs
```diff
@@ -25,13 +25,26 @@ struct AppConfig {
...
 
 #[derive(Args)]
-struct WriteBinary {}
+struct WriteCredentials {
+    #[arg(long = "app-config")]
+    config: String,
+    #[arg(long = "bin")]
+    binary_path: String,
+    #[arg(long = "nvs-size", default_value = "32768")]
+    size: usize,
```
npmenard, line 40: can we just deduce the binary size from the filed passed?

micro-rdk-installer/src/main.rs
```diff
@@ -25,13 +25,26 @@ struct AppConfig {
...
+    // the default here represents the offset as declared in
+    // examples/esp32/partitions.csv (0x9000, here written as 36864),
+    // as that is the partition table that will be used to compile
+    // the default application binary
+    #[arg(long = "nvs-offset-address", default_value = "36864")]
+    nvs_offset: u64,
```
npmenard, line 46: maybe as a later ticket. But would be nice to search for an nvs slot in the partition table and pull the numbers from there

micro-rdk-installer/src/main.rs
```diff
@@ -92,13 +105,39 @@ fn create_nvs_partition_binary(config_path: String, size: usize) -> Result<Vec<u
...
+    nvs_size: u64,
+    nvs_start_address: u64,
+) -> Result<(), Error> {
+    let mut app_file = OpenOptions::new()
+        .read(true)
+        .write(true)
+        .open(binary_path)
+        .map_err(Error::FileError)?;
+    let file_len = app_file.metadata().map_err(Error::FileError)?.len();
+    if (file_len - nvs_start_address) <= nvs_size {
```
npmenard, line 120: this check is a bit dubious? you would really need the info from the partition table to make sure the NVS partition doesn't override others partition

.github/workflows/test.yml
```diff
@@ -39,7 +39,7 @@ jobs:
 
     - name: Clippy Micro-RDK Installer
       run: |
-        bash -c 'make clippy-cli'
+        bash -c 'cargo update -p clap --precise v4.3.0 && make clippy-cli'
```
npmenard: ?

npmenard (approved):
LGTM couple of concerns but shouldn't prevent us to move forward. Would like to have a ticket that tracks sourcing the nvs par info from the partition table

.github/workflows/test.yml
gvaradarajan: it just keeps failing on building clap even though I've made no changes to the way I use clap or to Cargo.toml. So this is me grasping at straws

gvaradarajan: yeah, the thing is, it didn't happen in the previous flasher PR (which actually introduced using clap) and I can't replicate it locally. Essentially the complaint is that some clap dependency is requiring rustc 1.70, but again I have 1.67 locally and can't replicate the problem

npmenard: try to commit the Cargo.lock for the flasher

gvaradarajan: I tried having the actions cargo clean in the directory and now it's trying to use clap v4.4 even though I've specifically marked "4.3" in the Cargo.toml

micro-rdk-installer/src/main.rs
gvaradarajan: https://viam.atlassian.net/browse/RSDK-4711

PR #74, opened 2023-08-29
RSDK-4695 - installer CLI part 3: implement write and flash

gvaradarajan:
This PR looks larger than it is. Unfortunately, for builds to work I have to commit the Cargo.lock file which creates a very large diff because of espflash's dependencies

Testing instructions

1. Download the viam.json config file from app
2. From the top level of the directory (not from within examples), run `MICRO_RDK_USE_NVS=true make build-esp32-with-cred-bin`
3. `cd micro-rdk-installer && cargo run -- write-flash --app-config=<path to viam.json file> --binary-path=<path to local copy of micro-rdk repo>/examples/target/esp32-server-with-cred.bin --monitor` 
4. Follow the instructions in the terminal (it should ask you to input your wi-fi SSID and then your wi-fi password)

### DIFF
--- a/micro-rdk-installer/src/error.rs
+++ b/micro-rdk-installer/src/error.rs
@@ -1,3 +1,4 @@
+use espflash::error::Error as EspFlashError;
 use rcgen::RcgenError;
 use thiserror::Error;
 
@@ -9,6 +10,8 @@ pub enum Error {
     FileError(std::io::Error),
     #[error("Binary Too Small, File Size: {0}")]
     BinaryEditError(u64),
+    #[error("Binary Too Large, File Size: {0}")]
+    BinaryBufferError(u64),
     #[error("{0}")]
     WifiPasswordTooLongError(String),
     #[error("Wifi Credentials Error: {0}")]
@@ -27,6 +30,12 @@ pub enum Error {
     CertificateRequestError(String),
     #[error("App Connection Error: {0}")]
     AppConnectionError(anyhow::Error),
+    #[error("Flash connection error")]
+    FlashConnect,
+    #[error("EspFlash Flash error: {0}")]
+    EspFlashError(EspFlashError),
+    #[error("Monitor serial error: {0}")]
+    MonitorError(String),
     #[error("Unimplemented command: {0}")]
     UnimplementedError(String),
     #[error("No command received")]
@@ -39,7 +48,11 @@ impl From<RcgenError> for Error {
     }
 }
 
-// impl std::error::Error for Error {}
+impl From<EspFlashError> for Error {
+    fn from(value: EspFlashError) -> Self {
+        Self::EspFlashError(value)
+    }
+}
 
 impl From<serde_json::Error> for Error {
     fn from(value: serde_json::Error) -> Self {
--- a/micro-rdk-installer/src/main.rs
+++ b/micro-rdk-installer/src/main.rs
@@ -35,7 +35,7 @@ enum Commands {
 struct WriteCredentials {
     #[arg(long = "app-config")]
     config: String,
-    #[arg(long = "binary-path", short = "bin")]
+    #[arg(long = "app-binary")]
     binary_path: String,
     #[arg(long = "nvs-size", default_value = "32768")]
     nvs_size: usize,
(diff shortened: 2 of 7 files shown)


micro-rdk-installer/src/error.rs
```diff
@@ -27,6 +30,12 @@ pub enum Error {
     CertificateRequestError(String),
     #[error("App Connection Error: {0}")]
     AppConnectionError(anyhow::Error),
+    #[error("Flash connection error")]
+    FlashConnect,
+    #[error("EspFlash Flash error: {0}")]
+    EspFlashError(EspFlashError),
```
penguinland, line 36: Is this line going to lead to shadowing and synonyms that are hard to deal with? "No" is a plausible answer, but part of me wants the enum value to have a different name than the type it contains. You get around this later by using either `Self::EspFlashError` or `Error::EspFlashError`, but if they were different you wouldn't need to specify the namespace like that.

Admittedly, I don't have good suggestions of alternatives. Making one `EspFlasherError` could just make it more confusing because you won't remember when to add in the `er` and when to leave it out. Maybe just leave as-is and use `Self::` when relevant.

micro-rdk-installer/src/main.rs
```diff
@@ -1,9 +1,11 @@
 use std::fs::{self, File, OpenOptions};
-use std::io::{Seek, SeekFrom, Write};
+use std::io::{Read, Seek, SeekFrom, Write};
 
 use clap::{arg, command, Args, Parser, Subcommand};
 use dialoguer::theme::ColorfulTheme;
 use dialoguer::{Input, Password};
+use espflash::cli::config::Config;
+use espflash::cli::{connect, monitor::monitor, ConnectArgs, EspflashProgress};
```
penguinland, line 8: I'm about to learn something about Rust! Why is line 7 separate from line 8? Why not either roll them both together, or else split out `monitor::monitor` separate from the rest?

micro-rdk-installer/src/main.rs
```diff
@@ -47,7 +49,19 @@ struct WriteCredentials {
 }
 
 #[derive(Args)]
-struct WriteFlash {}
+struct WriteFlash {
+    #[arg(long = "app-config")]
+    config: String,
+    #[arg(long = "bin")]
+    binary_path: String,
+    #[arg(long = "nvs-size", default_value = "32768")]
```
penguinland, line 57: I'm about to learn something else about Rust! What does `#[arg(...)]` do? If `long` specifies a different name for the thing, why are the names different? Why is the "long" version of `binary_path` just "bin"?

micro-rdk-installer/src/main.rs
```diff
@@ -47,7 +49,19 @@ struct WriteCredentials {
...
+    size: usize,
+    // see comment for corresponding argument in WriteCredentials
+    #[arg(long = "nvs-offset-address", default_value = "36864")]
+    nvs_offset: u64,
```
penguinland, line 61: Why does this line start with `nvs_` but up on line 58, `size` does not have an `nvs_` prefix even though the long name for it does? It feels inconsistent, but perhaps it's consistent and I just don't have the right background to recognize that.

micro-rdk-installer/src/main.rs
```diff
@@ -127,6 +141,33 @@ fn write_credentials_to_app_binary(
...
+    let connect_args = ConnectArgs {
+        baud: Some(460800),
+        // let espflash auto-detect the port
+        port: None,
+        no_stub: false,
+    };
+    let conf = Config::default();
+    let mut flasher = connect(&connect_args, &conf).map_err(|_| Error::FlashConnect)?;
+    let mut f = File::open(binary_path).map_err(Error::FileError)?;
+    let size = f.metadata().map_err(Error::FileError)?.len();
```
penguinland, line 154: :+1: Nice use of `map_err` here!

penguinland (approved):
LGTM! Assuming the code works, my only comments are pointing out slight in consistencies, and if none of them get changed, that's okay with me.

micro-rdk-installer/src/main.rs
gvaradarajan: Alas, you're not about to learn something about Rust, it was just a simple mistake

micro-rdk-installer/src/main.rs
gvaradarajan: This is a feature of clap, the crate that I'm using for building the installer CLI. `arg` is essentially sugar for auto-parsing a CLI argument into a corresponding field on the struct. As for the names, I can make them more consistent (e.g. "bin" should probably be the short name and the long name should be "binary-path")

micro-rdk-installer/src/error.rs
gvaradarajan: I don't think this is an issue. It's not like you would import this enum representation in isolation right?

PR #75, opened 2023-09-02
RSDK-4249 - have installer download compiled binary of latest release when binary path not supplied

gvaradarajan:
Test instructions:
1. Download a copy of the server config (viam.json) of a robot (from step 1 of the Setup tab)
2. `cd micro-rdk-installer && cargo run -- write-flash --app-config=<path to viam.json>  --monitor`

### DIFF
--- a/micro-rdk-installer/src/main.rs
+++ b/micro-rdk-installer/src/main.rs
@@ -1,5 +1,6 @@
 use std::fs::{self, File, OpenOptions};
 use std::io::{Read, Seek, SeekFrom, Write};
+use std::path::PathBuf;
 
 use clap::{arg, command, Args, Parser, Subcommand};
 use dialoguer::theme::ColorfulTheme;
@@ -8,9 +9,12 @@ use espflash::cli::{config::Config, connect, monitor::monitor, ConnectArgs, Espf
 use micro_rdk_installer::error::Error;
 use micro_rdk_installer::nvs::data::{ViamFlashStorageData, WifiCredentials};
 use micro_rdk_installer::nvs::partition::{NVSPartition, NVSPartitionData};
-use micro_rdk_installer::nvs::request::populate_nvs_storage_from_app;
+use micro_rdk_installer::nvs::request::{
+    download_micro_rdk_release, populate_nvs_storage_from_app,
+};
 use secrecy::Secret;
 use serde::Deserialize;
+use tokio::runtime::Runtime;
 
 #[derive(Deserialize, Debug)]
 struct AppCloudConfig {
@@ -52,14 +56,20 @@ struct WriteFlash {
     #[arg(long = "app-config")]
     config: String,
     #[arg(long = "bin")]
-    binary_path: String,
+    binary_path: Option<String>,
+    #[arg(long = "version")]
+    version: Option<String>,
     #[arg(long = "nvs-size", default_value = "32768")]
     nvs_size: usize,
     // see comment for corresponding argument in WriteCredentials
     #[arg(long = "nvs-offset-address", default_value = "36864")]
     nvs_offset: u64,
+    #[arg(long = "baud-rate")]
+    baud_rate: Option<u32>,
     #[arg(long = "monitor")]
     monitor: bool,
+    #[arg(long = "debug")]
+    debug: bool,
 }
 
 #[derive(Args)]
@@ -119,7 +129,7 @@ fn create_nvs_partition_binary(config_path: String, size: usize) -> Result<Vec<u
 }
 
 fn write_credentials_to_app_binary(
-    binary_path: &str,
+    binary_path: PathBuf,
     nvs_data: &[u8],
     nvs_size: u64,
     nvs_start_address: u64,
@@ -130,7 +140,7 @@ fn write_credentials_to_app_binary(
         .open(binary_path)
         .map_err(Error::FileError)?;
     let file_len = app_file.metadata().map_err(Error::FileError)?.len();
-    if (file_len - nvs_start_address) <= nvs_size {
+    if (nvs_start_address + nvs_size) >= file_len {
         return Err(Error::BinaryEditError(file_len));
     }
     app_file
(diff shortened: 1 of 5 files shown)


micro-rdk-installer/src/main.rs
```diff
@@ -167,27 +172,52 @@ fn flash(binary_path: &str, should_monitor: bool) -> Result<(), Error> {
...
                 &nvs_data,
                 args.nvs_size as u64,
                 args.nvs_offset,
-            )?;
-            flash(&args.binary_path, args.monitor)?;
+            ) {
+                Ok(_) => {}
+                Err(e) => {
+                    if args.debug {
+                        std::fs::rename(path.clone(), "bad_binaries/micro-rdk-esp32-server.bin")
```
npmenard, line 214: what the purpose of bad_binaries?

micro-rdk-installer/src/nvs/request.rs
```diff
@@ -15,6 +19,22 @@ This module contains the logic for acquiring the security credentials for a robo
 the Viam App and preparing for flash storage
 */
 
+pub async fn download_latest_release(tmp_dir: &TempDir) -> Result<PathBuf, Error> {
+    let release_url = "https://github.com/viamrobotics/micro-rdk/releases/latest/download/micro-rdk-esp32-server.bin";
+    let fname = tmp_dir.path().join("micro-rdk-esp32-server.bin");
+    println!("will be located under: '{:?}'", fname);
```
npmenard, line 25: use log

micro-rdk-installer/src/main.rs
```diff
@@ -167,27 +172,52 @@ fn flash(binary_path: &str, should_monitor: bool) -> Result<(), Error> {
     Ok(())
 }
 
+fn get_path_or_download(path: Option<String>, tmp_dir: &TempDir) -> Result<PathBuf, Error> {
+    Ok(match path {
+        Some(path) => PathBuf::from(path),
+        None => {
+            let rt = Runtime::new().map_err(Error::AsyncError)?;
```
npmenard, line 179: i think it's fine here to decorate the main function with the tokio macro

micro-rdk-installer/src/main.rs
```diff
@@ -140,7 +145,7 @@ fn write_credentials_to_app_binary(
     Ok(())
 }
 
-fn flash(binary_path: &str, should_monitor: bool) -> Result<(), Error> {
+fn flash(binary_path: PathBuf, should_monitor: bool) -> Result<(), Error> {
     let connect_args = ConnectArgs {
         baud: Some(460800),
```
npmenard, line 150: that should be an option

micro-rdk-installer/src/main.rs
```diff
@@ -130,7 +135,7 @@ fn write_credentials_to_app_binary(
         .open(binary_path)
         .map_err(Error::FileError)?;
     let file_len = app_file.metadata().map_err(Error::FileError)?.len();
-    if (file_len - nvs_start_address) <= nvs_size {
+    if (nvs_start_address + nvs_size) >= file_len {
```
npmenard, line 138: right the total hardcoded length is 4MB right?

micro-rdk-installer/src/main.rs
```diff
@@ -167,27 +172,52 @@ fn flash(binary_path: &str, should_monitor: bool) -> Result<(), Error> {
...
+            let rt = Runtime::new().map_err(Error::AsyncError)?;
+            // let tmp_dir = Builder::new().prefix("micro-rdk-bin").tempdir().map_err(Error::FileError)?;
+            rt.block_on(download_latest_release(tmp_dir))?
+        }
+    })
+}
+
 fn main() -> Result<(), Error> {
     let cli = Cli::parse();
+    let tmp_dir = Builder::new()
```
npmenard, line 188: for clarity use tempdir::Builder

micro-rdk-installer/src/nvs/request.rs
```diff
@@ -15,6 +19,22 @@ This module contains the logic for acquiring the security credentials for a robo
 the Viam App and preparing for flash storage
 */
 
+pub async fn download_latest_release(tmp_dir: &TempDir) -> Result<PathBuf, Error> {
+    let release_url = "https://github.com/viamrobotics/micro-rdk/releases/latest/download/micro-rdk-esp32-server.bin";
```
npmenard, line 23: this should be an option too, by default it should point to release but users should be able to point to a specific version too. I don't know how github deals with releases but passing a version flag with version=latest or x.x.x would be perfect

npmenard (changes requested):
Done with first pass. I don't know why we have a bad_binaries folder now but ideally I would remove it or rename it to a better name :P

micro-rdk-installer/src/main.rs
gvaradarajan: I wanted to give the user a way to examine the downloaded binary if there was something wrong with it and still clean up the file afterwards. But now that I think about it, they could just download from the assets page themselves

micro-rdk-installer/src/main.rs
gvaradarajan: the hardcoded length of the binary?

npmenard: sorry I missed read the code. I actually don't quite understand the test here, if I interpret correctly it would trigger is nvs_start_address + nvs_size > 4MB (the resulting binary max size)

micro-rdk-installer/src/main.rs
gvaradarajan: I've refactored this a bit and not the run time is being used in main, is it still a good idea to use the macro?

npmenard: yes it would be better it's more inline with the rest of rust programs that uses tokio. In general unless you have a good reason for it you should avoid instantiating a runtime yourself

micro-rdk-installer/src/main.rs
gvaradarajan: this is for the (admittedly horrific scenario) that we have a bad binary and the end of the file cuts off in the middle of what would be the NVS partition.

npmenard (approved):
LGTM

PR #77, opened 2023-09-08
RSDK-4784 - add pwm support to board

gvaradarajan:
Tested on hardware: setting and getting PWM duty cycle and frequency works

This commit does not include having the motors in esp32/motor use pins via the board (if it does, potentially move the motor implementations to common?)

### DIFF
--- a/src/common/board.rs
+++ b/src/common/board.rs
@@ -34,6 +34,8 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
 pub struct FakeBoard {
     analogs: Vec<Rc<RefCell<dyn AnalogReader<u16, Error = anyhow::Error>>>>,
     i2cs: HashMap<String, Arc<Mutex<FakeI2CHandle>>>,
+    pin_pwms: HashMap<i32, f64>,
+    pin_pwm_freq: HashMap<i32, u64>,
 }
 pub trait Board: Status {
     fn set_gpio_pin_level(&mut self, pin: i32, is_high: bool) -> anyhow::Result<()>;
@@ -54,6 +56,10 @@ pub trait Board: Status {
     fn get_digital_interrupt_value(&self, _pin: i32) -> anyhow::Result<u32> {
         anyhow::bail!("this board does not support digital interrupts")
     }
+    fn get_pwm_duty(&self, pin: i32) -> f64;
+    fn set_pwm_duty(&mut self, pin: i32, duty_cycle_pct: f64) -> anyhow::Result<()>;
+    fn get_pwm_frequency(&self, pin: i32) -> anyhow::Result<u64>;
+    fn set_pwm_frequency(&mut self, pin: i32, frequency_hz: u64) -> anyhow::Result<()>;
 }
 
 pub type BoardType = Arc<Mutex<dyn Board>>;
@@ -65,7 +71,12 @@ impl FakeBoard {
         i2cs.insert(i2c0.name(), i2c0);
         let i2c1 = Arc::new(Mutex::new(FakeI2CHandle::new("i2c1".to_string())));
         i2cs.insert(i2c1.name(), i2c1);
-        FakeBoard { analogs, i2cs }
+        FakeBoard {
+            analogs,
+            i2cs,
+            pin_pwms: HashMap::new(),
+            pin_pwm_freq: HashMap::new(),
+        }
     }
     pub(crate) fn from_config(cfg: ConfigType) -> anyhow::Result<BoardType> {
         let analogs = if let Ok(analog_confs) = cfg.get_attribute::<HashMap<&str, f64>>("analogs") {
@@ -96,7 +107,12 @@ impl FakeBoard {
             HashMap::new()
         };
 
-        Ok(Arc::new(Mutex::new(FakeBoard { analogs, i2cs })))
+        Ok(Arc::new(Mutex::new(FakeBoard {
+            analogs,
+            i2cs,
+            pin_pwms: HashMap::new(),
+            pin_pwm_freq: HashMap::new(),
+        })))
     }
 }
 
@@ -155,6 +171,20 @@ impl Board for FakeBoard {
         } else {
             anyhow::bail!("could not find I2C with name {}", name)
         }
+    }
+    fn get_pwm_duty(&self, pin: i32) -> f64 {
+        *self.pin_pwms.get(&pin).unwrap_or(&0.0)
+    }
+    fn set_pwm_duty(&mut self, pin: i32, duty_cycle_pct: f64) -> anyhow::Result<()> {
+        self.pin_pwms.insert(pin, duty_cycle_pct);
+        Ok(())
+    }
+    fn get_pwm_frequency(&self, pin: i32) -> anyhow::Result<u64> {
+        Ok(*self.pin_pwm_freq.get(&pin).unwrap_or(&0))
+    }
+    fn set_pwm_frequency(&mut self, pin: i32, frequency_hz: u64) -> anyhow::Result<()> {
+        self.pin_pwm_freq.insert(pin, frequency_hz);
+        Ok(())
     }
 }
 
@@ -230,5 +260,21 @@ where
 
     fn get_digital_interrupt_value(&self, pin: i32) -> anyhow::Result<u32> {
         self.lock().unwrap().get_digital_interrupt_value(pin)
+    }
+
+    fn get_pwm_duty(&self, pin: i32) -> f64 {
+        self.lock().unwrap().get_pwm_duty(pin)
+    }
+
+    fn set_pwm_duty(&mut self, pin: i32, duty_cycle_pct: f64) -> anyhow::Result<()> {
+        self.lock().unwrap().set_pwm_duty(pin, duty_cycle_pct)
+    }
+
+    fn get_pwm_frequency(&self, pin: i32) -> anyhow::Result<u64> {
+        self.lock().unwrap().get_pwm_frequency(pin)
+    }
+
+    fn set_pwm_frequency(&mut self, pin: i32, frequency_hz: u64) -> anyhow::Result<()> {
+        self.lock().unwrap().set_pwm_frequency(pin, frequency_hz)
     }
 }
--- a/src/common/grpc.rs
+++ b/src/common/grpc.rs
@@ -412,12 +412,38 @@ where
         self.encode_message(status)
     }
 
-    fn board_pwm(&mut self, _message: &[u8]) -> Result<(), GrpcError> {
-        Err(GrpcError::RpcUnimplemented)
+    fn board_pwm(&mut self, message: &[u8]) -> Result<(), GrpcError> {
+        let req = component::board::v1::PwmRequest::decode(message)
+            .map_err(|_| GrpcError::RpcInvalidArgument)?;
+        let board = match self.robot.lock().unwrap().get_board_by_name(req.name) {
+            Some(b) => b,
+            None => return Err(GrpcError::RpcUnavailable),
+        };
+        let pin: i32 = req
+            .pin
+            .parse::<i32>()
+            .map_err(|_| GrpcError::RpcInvalidArgument)?;
+        let duty_cycle_pct = board.get_pwm_duty(pin);
+        let resp = component::board::v1::PwmResponse { duty_cycle_pct };
+        self.encode_message(resp)
     }
 
-    fn board_pwm_frequency(&mut self, _message: &[u8]) -> Result<(), GrpcError> {
-        Err(GrpcError::RpcUnimplemented)
+    fn board_pwm_frequency(&mut self, message: &[u8]) -> Result<(), GrpcError> {
+        let req = component::board::v1::PwmFrequencyRequest::decode(message)
+            .map_err(|_| GrpcError::RpcInvalidArgument)?;
+        let board = match self.robot.lock().unwrap().get_board_by_name(req.name) {
+            Some(b) => b,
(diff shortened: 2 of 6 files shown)


npmenard (commented):
I think we will need to devise another approach, we are reusing the timer and the channels used by the motor driver so setting the pwm frequency on any pin will affect the motor as well. We can probably use Rc and RefCell to circumvent most issues. Let's chat about an approach when you get back

gvaradarajan:
> I think we will need to devise another approach, we are reusing the timer and the channels used by the motor driver so setting the pwm frequency on any pin will affect the motor as well. We can probably use Rc and RefCell to circumvent most issues. Let's chat about an approach when you get back

Is this still an issue if we make the change I mentioned above? (where the motor no longer interacts with the pins independently and instead calls board functions)

src/common/board.rs
```diff
@@ -156,6 +172,20 @@ impl Board for FakeBoard {
             anyhow::bail!("could not find I2C with name {}", name)
         }
     }
+    fn get_pwm_duty(&self, pin: i32) -> anyhow::Result<f64> {
+        Ok(*self.pin_pwms.get(&pin).unwrap_or(&0.0))
```
npmenard, line 176: is it the behavior in RDK when a pin is not configured as a PWM? if so we don't need the Result then

src/common/grpc.rs
```diff
@@ -412,12 +412,34 @@ where
...
-    fn board_pwm(&mut self, _message: &[u8]) -> Result<(), GrpcError> {
-        Err(GrpcError::RpcUnimplemented)
+    fn board_pwm(&mut self, message: &[u8]) -> Result<(), GrpcError> {
+        let req = component::board::v1::PwmRequest::decode(message)
+            .map_err(|_| GrpcError::RpcInvalidArgument)?;
+        let board = match self.robot.lock().unwrap().get_board_by_name(req.name) {
+            Some(b) => b,
+            None => return Err(GrpcError::RpcUnavailable),
+        };
+        let pin: i32 = req.pin.parse::<i32>().unwrap();
```
npmenard, line 422: could you return an error instead here?

src/esp32/pin.rs
```diff
@@ -9,3 +14,143 @@ impl<'d, T: Pin, MODE> PinExt for PinDriver<'d, T, MODE> {
...
+
+    pub fn pin(&self) -> i32 {
+        self.pin
+    }
+
+    pub fn is_high(&self) -> bool {
+        self.driver.is_high()
+    }
+
+    pub fn set_high(&mut self) -> anyhow::Result<()> {
```
npmenard, line 65: do we need to adapt the behavior here when a PWM is configured on the pin?

src/esp32/pwm.rs
```diff
@@ -0,0 +1,209 @@
...
+                    Some(x) => x == frequency_hz,
+                    None => false
+                });
+            match timer_number {
+                Some(timer_number) => {
+                    self.timer_channels[timer_number].push(channel);
+                    timer_number
+                },
+                None => {
+                    let timer_number = self.timer_channels.iter().position(|x| x.is_empty()).unwrap();
```
npmenard, line 160: can we return an error instead of unwrap?

src/esp32/pwm.rs
```diff
@@ -0,0 +1,209 @@
...
+        let mut ledc_manager = LEDC_MANAGER.lock().unwrap();
+        ledc_manager.drop_timer(self.timer_number, self.channel);
+        let timer_number = ledc_manager.take_timer(frequency_hz, self.channel)?;
+        let timer: ledc_timer_t = (timer_number as u8).into();
+        match esp_idf_sys::esp!(unsafe {
+            ledc_set_freq(SpeedMode::LowSpeed.into(), timer, frequency_hz)
+        }) {
+            Ok(val) => { Ok(val) },
+            Err(err) => {
+                ledc_manager.drop_timer(timer_number, self.channel);
```
npmenard, line 107: why is it potentially called twice?

src/esp32/pwm.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    fn take_next_channel(&mut self, pin: i32) -> anyhow::Result<PwmChannel> {
+        for i in 0..2 {
+            match self.take_channel(i, pin) {
+                Ok(ret) => {
+                    return Ok(ret);
+                }
+                Err(err) => {
+                    log::debug!("Skipping: {:?}", err);
```
npmenard, line 203: skipping?

src/esp32/pin.rs
```diff
@@ -9,3 +14,143 @@ impl<'d, T: Pin, MODE> PinExt for PinDriver<'d, T, MODE> {
...
+
+    pub fn get_pwm_frequency(&self) -> u64 {
+        match &self.pwm_driver {
+            Some(pwm_driver) => pwm_driver.get_frequency() as u64,
+            None => 0,
+        }
+    }
+
+    pub fn set_pwm_frequency(&mut self, freq: u64) -> anyhow::Result<()> {
+        match self.pwm_driver.as_mut() {
```
npmenard, line 104: how do we protect against configuring a pin that's already a digital interrupt?

src/esp32/pwm.rs
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    pub fn get_frequency(&self) -> u32 {
+        unsafe { ledc_get_freq(SpeedMode::LowSpeed.into(), self.ledc_driver.timer()) }
+    }
+
+    pub fn set_frequency(&mut self, frequency_hz: u32) -> anyhow::Result<()> {
+        let mut ledc_manager = LEDC_MANAGER.lock().unwrap();
+        ledc_manager.drop_timer(self.timer_number, self.channel);
+        let timer_number = ledc_manager.take_timer(frequency_hz, self.channel)?;
```
npmenard, line 100: does this function supposed to release the timer if the frequency is changed?

src/esp32/pwm.rs
```diff
@@ -0,0 +1,209 @@
...
+            6 => PwmChannel::C6,
+            7 => PwmChannel::C7,
+            _ => anyhow::bail!("Could not find CHANNEL{:}", n),
+        };
+        self.channel_pin_subscriptions.insert(n, pin);
+        Ok(res)
+    }
+
+    fn take_next_channel(&mut self, pin: i32) -> anyhow::Result<PwmChannel> {
+        for i in 0..2 {
```
npmenard, line 197: 0..2 ?

npmenard (changes requested):
Done with first pass

src/common/board.rs
```diff
@@ -64,7 +70,12 @@ impl FakeBoard {
         i2cs.insert(i2c0.name(), i2c0);
         let i2c1 = Arc::new(Mutex::new(FakeI2CHandle::new("i2c1".to_string())));
         i2cs.insert(i2c1.name(), i2c1);
-        FakeBoard { analogs, i2cs }
+        FakeBoard {
+            analogs,
+            i2cs,
+            pin_pwms: HashMap::new(),
```
mattjperez, line 76: nit: can we use `..Default::default()`?

src/common/board.rs
```diff
@@ -96,7 +107,12 @@ impl FakeBoard {
             HashMap::new()
         };
 
-        Ok(Arc::new(Mutex::new(FakeBoard { analogs, i2cs })))
+        Ok(Arc::new(Mutex::new(FakeBoard {
+            analogs,
+            i2cs,
+            pin_pwms: HashMap::new(),
```
mattjperez, line 113: Same as above

src/common/board.rs
mattjperez: Is there any reason, at least for the Micro-RDK, we would want to know if the pwm was actually 0.0 or not there? Or do we just interpret 0.0 as not configured in any case?

src/esp32/pwm.rs
```diff
@@ -0,0 +1,209 @@
...
+            timer_channels: [Vec::new(), Vec::new(), Vec::new(), Vec::new()]
+        }
+    }
+
+    fn find_timer_by_frequency(&mut self, frequency_hz: u32) -> Option<usize> {
+        self.timer_frequencies.iter().position(|&x| x == Some(frequency_hz))
+    }
+
+    fn take_timer(&mut self, frequency_hz: u32, channel: PwmChannel) -> anyhow::Result<usize> {
+        let timer_number = if self.timer_frequencies.iter().all(|&x| x.is_some()) {
```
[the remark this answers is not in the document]
gvaradarajan: I can't with the way it currently works. Changing the PWM frequency for a pin that already has a PWM signal makes the pin take a new timer. If there was no other pin using that timer, then that timer is made available for a new choice of PWM frequency.

Would it be preferable to change the timer frequency if not other pins are using that timer?

src/esp32/pwm.rs
```diff
@@ -0,0 +1,209 @@
...
+        Ok(())
+    }
+
+    pub fn get_duty_pct(&self) -> f64 {
+        let max_duty = self.ledc_driver.get_max_duty();
+        (self.ledc_driver.get_duty() as f64) / (max_duty as f64)
+    }
+
+    pub fn get_frequency(&self) -> u32 {
+        unsafe { ledc_get_freq(SpeedMode::LowSpeed.into(), self.ledc_driver.timer()) }
```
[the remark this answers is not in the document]
gvaradarajan: I didn't realize this, but this makes ESP32 implementation of `ledc_timer_t` very confusing. Are you saying, for a particular choice of timer number, if I change the speed mode I've technically changed the underlying timer?

src/esp32/pwm.rs
gvaradarajan: because according to the LedcManager, the timer has been taken in L. 100, but I can just move L. 99 down

src/esp32/pwm.rs
gvaradarajan: oops I reverted some code that wasn't working before I pushed the last commit and this accidentally stayed. It should be 8

src/esp32/pin.rs
gvaradarajan: ah I simply forgot, I'll just error if `self.interrupt_type` is not `None`

src/esp32/pwm.rs
gvaradarajan: it does not, that's why drop_timer is called. Should it?

src/esp32/pwm.rs
gvaradarajan: I put it in while debugging and left it in by accident

src/esp32/pwm.rs
gvaradarajan: due to the condition in L. 36, it's impossible that this is an error. Would it be acceptable to make it a match statement and use unreachable?

src/esp32/pin.rs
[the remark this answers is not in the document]
gvaradarajan: I don't think RDK does any such thing, is that an oversight? I can return an error if there's a PWM signal

src/common/board.rs
gvaradarajan: @npmenard yes this is the behavior in RDK I believe, I'll remove Result

@mattjperez you can probably distinguish by querying the frequency (if frequency is 0, then the PWM was not there), so I don't think we have to distinguish between the scenarios in this call

src/esp32/pwm.rs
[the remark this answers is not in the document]
npmenard: yeah they do imply that timer are different but I agree it looks weird, some people have reported that they were able to run 10 PWMs so I am tempted to assume this is possible. For this round of implementation let's leave it like this

src/esp32/pwm.rs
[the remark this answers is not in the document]
npmenard: I see would it work if we first look if there is a matching timer and if not if there is a free timer available? The logic feels a bit convoluted and hard to read

npmenard (changes requested):
done with second pass

src/esp32/pwm.rs
```diff
@@ -0,0 +1,239 @@
...
+                    self.timer_channels[timer_number].push(channel);
+                    timer_number
+                }
+            }
+        };
+        Ok(timer_number)
+    }
+
+    fn drop_timer(&mut self, timer_number: usize, channel: PwmChannel) {
+        let timer_channels = &mut self.timer_channels[timer_number];
```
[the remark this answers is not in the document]
gvaradarajan: I'm not sure I understand what you mean by this?

src/esp32/pwm.rs
```diff
@@ -0,0 +1,239 @@
...
+            6 => PwmChannel::C6,
+            7 => PwmChannel::C7,
+            _ => { return Err(Esp32PwmError::ChannelNotFound(n)); },
+        };
+        self.channel_pin_subscriptions.insert(n, pin);
+        Ok(res)
+    }
+
+    fn take_next_channel(&mut self, pin: i32) -> Result<PwmChannel, Esp32PwmError> {
+        for i in 0..8 {
```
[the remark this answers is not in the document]
gvaradarajan: Or I could use `ledc_channel_t_LEDC_CHANNEL_MAX` instead of 8, does that work?

src/esp32/pwm.rs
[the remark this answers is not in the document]
npmenard: My understanding is that ledc work the following way for example TIMER0 -> Channel0 -> GPIOPinX so if this function change the timer then LedcDriver needs to reconfigure the channel with the new Timer but I don't see where this is happening

npmenard (approved):
LGTM !

PR #80, opened 2023-09-18
[RSDK-4960] Publish workflow

npmenard:
Automatically publish a esp32-serrver binary when creating a tag. Treat rc release as pre-release

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -18,12 +18,12 @@ jobs:
       uses: actions/checkout@v3
       with:
         fetch-depth: 2
-    - name: Build release image
+    - name: Build esp32 binary
       run: |
         bash -c 'export MICRO_RDK_USE_NVS=true && . "$IDF_PATH"/export.sh && . "$ESP_ROOT"/export-esp.sh && make build-esp32-with-cred-bin'
         cp examples/target/esp32-server-with-cred.bin  micro-rdk-esp32-server.bin
         sha256sum  micro-rdk-esp32-server.bin  >> sha256sums.txt
-    - name: Check if RC release
+    - name: Check release type
       id: check-tag
       run: |
         if echo ${{ github.event.ref }} | grep -Eq '^refs/tags/v.*rc[0-9]{1}$'; then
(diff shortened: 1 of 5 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,51 @@
...
+        fi
+    - name: Publish release candidate
+      if: steps.check-tag.outputs.match == 'true'
+      uses: marvinpinto/action-automatic-releases@latest
+      with:
+        repo_token: ${{ secrets.GITHUB_TOKEN }}
+        files: |
+          micro-rdk-esp32-server.bin
+          sha256sums.txt
+        prerelease: true
```
acmorrow, line 42: Could you dedup this with something like `prerelease: ${{needs.check-tag.outputs.output1}}` per https://docs.github.com/en/actions/using-jobs/defining-outputs-for-jobs?

I think you would need to make "Publish" its own job though. So something like:

```
jobs:
  tests: ...
  build-release:
      outputs:
        prerelease: ${{ steps.check-tag.outputs.prerelease }}
    steps:
     - name: Checkout...
     - name: Build ...
     - name: Check release type
        id: check-tag
        run: |
          if echo ${{ github.event.ref }} | grep -Eq '^refs/tags/v.*rc[0-9]{1}$'; then
                echo "prerelease=true" >> $GITHUB_OUTPUT
          else
                echo "prerelease=false" >> $GITHUB_OUTPUT
          fi
  publish-release:
    needs: build-release
    steps:
    - publish:
      uses: marvinpinto/action-automatic-releases@latest
      with:
        repo_token: ${{ secrets.GITHUB_TOKEN }}
        files: |
          micro-rdk-esp32-server.bin
          sha256sums.txt
        prerelease: ${{needs.build-release.outputs.prerelease}}
```

Something like that.

acmorrow (approved):
LGTM with an optional suggestion to remove the duplication around `prerelease`

.github/workflows/test.yml
```diff
@@ -37,16 +42,16 @@ jobs:
       with:
         filters: |
           src-esp32:
-            - 'src/esp32/**'
+            - 'src/esp3233/**'
```
acmorrow, line 45: Was this maybe left over from some debugging?

npmenard: yes I'll remove them before merge

PR #81, opened 2023-09-26
RSDK-4711 - read NVS size and offset from app binary in micro-RDK installer

gvaradarajan:


Makefile
```diff
@@ -79,7 +79,7 @@ build-esp32-bin:
 	cd examples && cargo espflash save-image --merge --chip esp32 target/esp32-server.bin -T esp32/partitions.csv -s 4M  --bin esp32-server --target=xtensa-esp32-espidf  -Zbuild-std=std,panic_abort --release
 
 build-esp32-with-cred-bin:
-	cd examples && cargo espflash save-image --merge --chip esp32 target/esp32-server-with-cred.bin -T esp32/partitions.csv -s 4M  --bin esp32-server-with-cred --target=xtensa-esp32-espidf  -Zbuild-std=std,panic_abort --release
+	export MICRO_RDK_USE_NVS=true && cd examples && cargo espflash save-image --merge --chip esp32 target/esp32-server-with-cred.bin -T esp32/partitions.csv -s 4M  --bin esp32-server-with-cred --target=xtensa-esp32-espidf  -Zbuild-std=std,panic_abort --release
```
npmenard, line 82: why do we need to add the env variable now?

micro-rdk-installer/src/nvs/metadata.rs
```diff
@@ -0,0 +1,56 @@
...
+}
+
+fn get_nvs_metadata_from_entry(line: &[u8; 30]) -> Result<Option<NVSMetadata>, Error> {
+    let type_is_data = line[0] == 0x01;
+    let subtype_is_nvs = line[1] == 0x02;
+    if type_is_data && subtype_is_nvs {
+        if line[2..6] == [0xFF, 0xFF, 0xFF, 0xFF] {
+            return Err(Error::NVSOffsetMissingError);
+        }
+        let offset = u32::from_le_bytes(line[2..6].try_into().unwrap());
```
npmenard, line 48: let's not unwrap if we can return the error

npmenard (approved):
LGTM couple of minor comments

PR #82, opened 2023-09-26
RSDK-4252 - publish builds of installer on release

gvaradarajan:


.github/workflows/publish.yml
```diff
@@ -8,6 +8,62 @@ on:
...
+  
+  publish-installer-release:
+    needs: tests
+    strategy:
+      fail-fast: false
+      matrix:
+        job:
+          - { target: x86_64-unknown-linux, exe: amd64-linux, os: ubuntu-latest }
+          - { target: aarch64-unknown-linux, exe: aarch64-linux, os: ubuntu-latest }
+          - { target: x86_64-apple-darwin, exe: macos, os: macos-latest }
```
randhid, line 20: What about darwin-aarch64?

.github/workflows/publish.yml
```diff
@@ -8,6 +8,62 @@ on:
...
+    steps:
+    - name: Checkout main branch code
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 2
+    - name: Set toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: 1.62.0
```
npmenard, line 35: we can use 1.70 here

.github/workflows/publish.yml
```diff
@@ -8,6 +8,62 @@ on:
...
+        override: true
+        target: ${{ matrix.job.target }}
+    - name: Build installer binary
+      uses: actions-rs/cargo@v1
+      with:
+        use-cross: true
+        args: --release --target=${{ matrix.job.target }}
```
npmenard, line 42: add --locked

.github/workflows/publish.yml
```diff
@@ -8,6 +8,62 @@ on:
...
+        command: build 
+    - name: Rename result
```
npmenard, line 44: i think this is going to clash (remains to try) with the release of the micro-rdk binary so we want to make sure to test that otherwise integrate both workflow together

npmenard (changes requested):
done with first pass

.github/workflows/publish.yml
```diff
@@ -8,7 +8,100 @@ on:
...
+        if [[ $(find . -wholename 'target/${{ matrix.job.target }}/release/micro-rdk-installer*' -printf c | wc -c) == "1" ]]; then
+            cp target/${{ matrix.job.target }}/release/micro-rdk-installer* micro-rdk-installer-${{ matrix.job.exe }}
+        else
+            exit 1
+        fi
+    - name: Check release type
+      id: check-tag
+      run: |
+        if echo ${{ github.event.ref }} | grep -Eq '^refs/tags/v.*rc[0-9]{1}$'; then
+              echo "match=true" >> $GITHUB_OUTPUT
```
abe-winter, line 62: optional: would name this 'is_rc_release' or something for clarity, 'match' could mean anything

.github/workflows/publish.yml
```diff
@@ -8,7 +8,100 @@ on:
...
+        else
+              echo "match=false" >> $GITHUB_OUTPUT
+        fi
+    - name: Upload release
+      uses: actions/upload-artifact@v2
+      with:
+        name: micro-rdk-installer-${{ matrix.job.exe }}
+        path: micro-rdk-installer-${{ matrix.job.exe }}
+  
+  build-installer-arm64-linux:
```
abe-winter, line 72: ideally add comment here saying why this needs its own job (vs the other 3 architectures which share a job)

.github/workflows/publish.yml
```diff
@@ -8,7 +8,100 @@ on:
...
   tests:
     uses: ./.github/workflows/test.yml
-  publish-release:
+  
+  build-installer:
+    strategy:
+      fail-fast: false
+      matrix:
+        job:
+          - { target: x86_64-unknown-linux-gnu, exe: amd64-linux, os: ubuntu-latest}
```
abe-winter, line 17: might rename 'exe' to 'suffix' for these for clarity

FYI another way to do this is to use [include](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#expanding-or-adding-matrix-configurations) -- I have no preference, basically only do this if it gives you better vscode completions and that's useful to you

abe-winter (approved):
I just did a once over for workflow stuff, but looks good to me

left a bunch of optional comments but nothing pressing

.github/workflows/publish.yml
```diff
@@ -31,11 +133,42 @@ jobs:
...
         repo_token: ${{ secrets.GITHUB_TOKEN }}
         files: |
+          micro-rdk-installer-arm64-linux
+          micro-rdk-installer-amd64-linux
+          micro-rdk-installer-macos
+          micro-rdk-installer-windows.exe
           micro-rdk-esp32-server.bin
           sha256sums.txt
-        prerelease: ${{ steps.check-tag.outputs.match }}
+        prerelease: ${{ steps.check-tag.outputs.match }}
```
randhid, line 174:
```suggestion
        prerelease: ${{ steps.check-tag.outputs.match }}
        
```

npmenard (approved):
LGTM

randhid (approved):
😃

.github/workflows/publish.yml
```diff
@@ -8,7 +8,102 @@ on:
...
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 2
+    - name: Set toolchain
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: 1.70.0
+        override: true
+        target:
```
abe-winter, line 93: target is blank fyi

.github/workflows/publish.yml
```diff
@@ -8,7 +8,102 @@ on:
...
+        fi
+    - name: Upload release
+      uses: actions/upload-artifact@v2
+      with:
+        name: micro-rdk-installer-${{ matrix.job.file_suffix }}
+        path: micro-rdk-installer/micro-rdk-installer-${{ matrix.job.file_suffix }}
+
+  # github-actions does not have a runner that supports arm64 linux and cross-compilation does not seem to work for the ring crate
+  build-installer-arm64-linux:
+    runs-on: [x64, qemu-host]
```
abe-winter, line 74: re runners: you're hitting an x86 box with QEMU I think? you could alternatively hit one of our arm runners or use buildjet

guessing native ARM will be faster than x86, but if you're okay with current build speed, ignore this

abe-winter (approved):
still lgtm, left comments

.github/workflows/publish.yml
gvaradarajan: yeah it's a bad copy/paste job from my testing branch, I'm gonna re-copy

PR #87, opened 2023-10-09
adjust esp32 motors to use create_pwm_driver

gvaradarajan:


gvaradarajan:
Closing in favor of the approach in https://github.com/viamrobotics/micro-rdk/pull/88

PR #88, opened 2023-10-09
RSDK-4784 - impl common motor using board + remove esp32 specific motor

gvaradarajan:


src/common/gpio_motor.rs
```diff
@@ -0,0 +1,404 @@
...
+        match dep {
+            Resource::Encoder(found_enc) => {
+                enc = Some(found_enc.clone());
+                break;
+            }
+            _ => {
+                continue;
+            }
+        };
+    }
```
acmorrow, line 94: Do you need to do this now, or could you move it down around line 108?

src/common/gpio_motor.rs
```diff
@@ -0,0 +1,404 @@
...
+        }
+    };
+    if let Some(enc) = enc {
+        let enc_motor = EncodedMotor::new(motor, enc.clone());
+        return Ok(Arc::new(Mutex::new(enc_motor)));
+    }
+    Ok(motor)
+}
+
+pub struct EncodedMotor<M, Enc> {
```
acmorrow, line 115: Getting a little out of my zone of rust confidence here, but should there be some constraints here that `M` and `Enc` are actually `Motor`'s and `Encoder`'s?

acmorrow (commented):
I'm still working my way through but here are some initial thoughts.

src/common/gpio_motor.rs
```diff
@@ -0,0 +1,404 @@
...
+    }
+}
+
+impl<M, Enc> Stoppable for EncodedMotor<M, Enc>
+where
+    M: Motor,
+    Enc: Encoder,
+{
+    fn stop(&mut self) -> anyhow::Result<()> {
+        self.set_power(0.0)
```
acmorrow, line 157: I think maybe this would be better written as `self.motor.stop`.

src/common/gpio_motor.rs
```diff
@@ -0,0 +1,404 @@
...
+    }
+}
+
+impl<B> Status for PwmDirectionMotor<B>
+where
+    B: Board,
+{
+    fn get_status(&self) -> anyhow::Result<Option<google::protobuf::Struct>> {
+        let mut hm = HashMap::new();
+        let pos = 0.0;
```
acmorrow, line 386: Just a question, but why would an unencoded motor report a position in its status?

src/common/gpio_motor.rs
```diff
@@ -0,0 +1,404 @@
...
+            google::protobuf::Value {
+                kind: Some(google::protobuf::value::Kind::NumberValue(pos)),
+            },
+        );
+        Ok(Some(google::protobuf::Struct { fields: hm }))
+    }
+}
+
+// Represents a motor using a A, B, and PWM pins
+pub struct ABPwmMotor<B> {
```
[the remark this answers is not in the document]
gvaradarajan: Hmm, I'm not sure. The other structs in common all seem to be public, so I was just following the pattern

src/common/gpio_motor.rs
```diff
@@ -0,0 +1,404 @@
...
+    ) -> anyhow::Result<Self> {
+        let mut res = Self {
+            board,
+            a_pin,
+            b_pin,
+            pwm_pin,
+            max_rpm,
+            dir_flip,
+        };
+        res.board.set_pwm_frequency(pwm_pin, 10000)?;
```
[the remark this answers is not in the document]
gvaradarajan: This was what the existing esp32 motor implementation was using. From what I can gather looking into this, the actual of the frequency is not important but rather its order of magnitude (which should be on the order of 10s of kHz)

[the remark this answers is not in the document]
gvaradarajan: I'll put a clarifying comment and make it a const

src/common/gpio_motor.rs
gvaradarajan: I need to do it here because `get_board_from_dependencies` takes ownership of `deps` and I didn't want to change its method signature

src/common/gpio_motor.rs
```diff
@@ -0,0 +1,404 @@
...
+        if !(-1.0..=1.0).contains(&pct) {
+            anyhow::bail!("power outside limit")
+        }
+        let set_high = (pct > 0.0) && !self.dir_flip;
+        self.board.set_gpio_pin_level(self.dir_pin, set_high)?;
+        self.board.set_pwm_duty(self.pwm_pin, pct)?;
+        Ok(())
+    }
+
+    fn get_position(&mut self) -> anyhow::Result<i32> {
```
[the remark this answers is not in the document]
gvaradarajan: Good point

src/common/gpio_motor.rs
```diff
@@ -0,0 +1,404 @@
...
+        }
+        let set_forwards = (pct > 0.0) && !self.dir_flip;
+        if set_forwards {
+            self.board.set_gpio_pin_level(self.a_pin, false)?;
+            self.board.set_gpio_pin_level(self.b_pin, true)?;
+        } else {
+            self.board.set_gpio_pin_level(self.a_pin, true)?;
+            self.board.set_gpio_pin_level(self.b_pin, false)?;
+        }
+        self.board.set_pwm_duty(self.pwm_pin, pct)?;
```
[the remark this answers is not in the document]
gvaradarajan: I actually think probably not. I think that the A-B pins determine direction and the PWM signal is what drives the motor?

src/common/gpio_motor.rs
gvaradarajan: In order to remove the position from this map, we have to implement `GetProperties` in the motor API because otherwise the position badge on the motor RC card will read `NaN`. Should I make a ticket for implementing `GetProperties`?

src/common/gpio_motor.rs
```diff
@@ -0,0 +1,411 @@
...
+            anyhow::bail!("power outside limit")
+        }
+        let set_high = (pct > 0.0) && !self.dir_flip;
+        self.board.set_gpio_pin_level(self.dir_pin, set_high)?;
+        self.board.set_pwm_duty(self.pwm_pin, pct)?;
+        Ok(())
+    }
+
+    fn get_position(&mut self) -> anyhow::Result<i32> {
+        anyhow::bail!("position reporting not supported for PwmDirectionMotor, did you mean to attach an encoder?")
```
npmenard, line 374: nit: position reporting not supported without an encoder

npmenard (changes requested):
Done with first pass some minor comments
In general we should be mindful about not exposing public function unless necessary or required by traits

src/common/gpio_motor.rs
```diff
@@ -0,0 +1,411 @@
...
+    }
+    Ok(motor)
+}
+
+// Motors generally don't care about the PWM frequency, so long as
+// it is in the order of magnitude 0f 10s of kHZ. For simplicity, we
+// just select 10kHz.
+const MOTOR_PWM_FREQUENCY: u64 = 10000;
+
+pub struct EncodedMotor<M, Enc> {
```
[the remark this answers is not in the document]
gvaradarajan: discussed offline, going to do pub(crate) for everything except EncodedMotor (so that people can better test wrapping their custom encoder/motor implementations if needed)

src/common/gpio_motor.rs
gvaradarajan: ticket created: https://viam.atlassian.net/browse/RSDK-5331

src/common/gpio_motor.rs
acmorrow: Discussed offline, resolving.

acmorrow (approved):
LGTM when @npmenard is happy, mod one pending request to improve an error message.

npmenard (approved):
LGTM

PR #91, opened 2023-10-11
add functionality to installer to just monitor an ESP32

gvaradarajan:


micro-rdk-installer/src/nvs/metadata.rs
```diff
@@ -53,16 +59,16 @@ fn get_nvs_metadata_from_entry(line: &[u8; 30]) -> Result<Option<NVSMetadata>, E
...
         });
-        let size = u32::from_le_bytes(match line[6..10].try_into(){
+        let size = u32::from_le_bytes(match line[6..10].try_into() {
             Ok(val) => val,
             Err(_) => {
                 unreachable!()
             }
         });
-        return Ok(Some(NVSMetadata{
+        return Ok(Some(NVSMetadata {
```
npmenard, line 68: don't want to be pedantic but can we remove the spaces l68 & 62.
Also do we run cargo fmt for the micro-rd-installer ?

npmenard (approved):
LGTM

micro-rdk-installer/src/nvs/metadata.rs
gvaradarajan: we don't, I'll add it to the workflow. Also, cargo fmt added these spaces, should I still try to leave them out?

PR #92, opened 2023-10-17
RSDK-5263 - push logs to app concerning the status of robot configuration

gvaradarajan:


src/common/app_client.rs
```diff
@@ -1,12 +1,15 @@
 #![allow(unused)]
+use anyhow::Context;
 use bytes::{BufMut, Bytes, BytesMut};
+use chrono::{format::ParseError, DateTime, FixedOffset};
 use futures_lite::Future;
 use prost::{DecodeError, EncodeError, Message};
-use std::{net::Ipv4Addr, pin::Pin, rc::Rc};
+use std::{net::Ipv4Addr, pin::Pin, rc::Rc, time::SystemTime};
+use stun_codec::rfc5766::attributes::Data;
```
npmenard, line 8: why did you need to pull in this?

src/common/app_client.rs
```diff
@@ -196,10 +211,34 @@ impl<'a> AppClient<'a> {
         };
         let body = encode_request(req)?;
 
-        let mut r = self.grpc_client.send_request(r, body)?;
+        let (mut r, headers) = self.grpc_client.send_request(r, body)?;
+        let date_str = headers
```
npmenard, line 215: would not prevent building robot / continuing in the absence of date. just set in to 0 logs would still be pushed in app

src/common/log.rs
```diff
@@ -0,0 +1,40 @@
+use crate::google::protobuf::{value::Kind, Struct, Timestamp, Value};
+use crate::proto::app::v1::LogEntry;
+use chrono::{DateTime, FixedOffset};
+use std::collections::HashMap;
+
+pub fn config_log_entry(
+    cfg_received_datetime: DateTime<FixedOffset>,
```
npmenard, line 7: nit: rename as time

src/common/log.rs
```diff
@@ -0,0 +1,40 @@
...
+    err: Option<&anyhow::Error>,
```
npmenard, line 8: can we pass it ass std::error::Error?

src/common/log.rs
```diff
@@ -0,0 +1,40 @@
...
+) -> LogEntry {
+    let secs = cfg_received_datetime.timestamp();
+    let nanos = cfg_received_datetime.timestamp_subsec_nanos();
+    let level = match err {
+        Some(_) => "error".to_string(),
+        None => "info".to_string(),
+    };
+    let message = match err {
+        Some(err) => format!("could not create robot from config: {err}"),
```
npmenard, line 17: looks like we could pass it as a generic implementing the display trait

src/common/log.rs
```diff
@@ -0,0 +1,40 @@
...
+        level,
+        time: Some(Timestamp {
+            seconds: secs,
+            nanos: nanos as i32,
+        }),
+        logger_name: "robot_server".to_string(),
+        message,
+        caller: Some(Struct {
+            fields: HashMap::from([(
+                "Defined".to_string(),
```
npmenard, line 31: what does that mean?

src/esp32/entry.rs
```diff
@@ -50,16 +51,29 @@ pub fn serve_web(
...
+                    let r = match LocalRobot::new_from_config_response(&cfg_response, registry) {
+                        Ok(robot) => {
+                            let logs = vec![config_log_entry(cfg_received_datetime, None)];
+                            client.push_logs(logs).expect("could not push logs to app");
+                            robot
+                        }
+                        Err(err) => {
+                            let logs = vec![config_log_entry(cfg_received_datetime, Some(&err))];
+                            client.push_logs(logs).expect("could not push logs to app");
+                            panic!("{}", err)
```
npmenard, line 69: nit: I don't like the panic here, we could ultimatey re-poll the config until the error is fixed. I think we can live with this for this pr but would add a ticket

npmenard:
Maybe have Dan take a look at it with his redesign of logging. not sure if the API will be affected

src/common/log.rs
gvaradarajan: I honestly don't know, but without this the logs won't show up in app because the frontend tries to access undefined properties for display

src/common/log.rs
gvaradarajan: I'm not sure I understand, do you mean the message should be the argument to the function instead of an error and that argument should be type `dyn std::fmt::Display`?

npmenard (approved):
LGTM

PR #105, opened 2023-10-30
`Board` docs

mattjperez:
Changes to crate documentation for developer use. 
Method descriptions try to mirror the [Python SDK](https://python.viam.dev/autoapi/viam/components/index.html).

This PR highlights different options for how documentation can be added and rendered, as well as the directory structure changes (individual `mod.rs` or using the root `lib.rs`).

If this is approved, I'll add docs for the other components.

use `make doc-open` to generate docs and view in browser

### DIFF
--- a/src/common/board.rs
+++ b/src/common/board.rs
@@ -2,35 +2,19 @@
 
 #![allow(dead_code)]
 use crate::{
-    common::{
-        analog::AnalogReader,
-        status::Status,
-    },
+    common::{analog::AnalogReader, status::Status},
     google,
-    proto::{
-        common, 
-        component::board::v1::PowerMode
-    },
+    proto::{common, component::board::v1::PowerMode},
 };
 
-use log::*;
 use core::cell::RefCell;
-use std::{
-    collections::HashMap,
-    rc::Rc,
-    sync::Arc,
-    sync::Mutex,
-    time::Duration,
-};
+use log::*;
+use std::{collections::HashMap, rc::Rc, sync::Arc, sync::Mutex, time::Duration};
 
 use super::{
     analog::FakeAnalogReader,
     config::ConfigType,
-    i2c::{
-        FakeI2CHandle, 
-        FakeI2cConfig, 
-        I2CHandle, 
-        I2cHandleType},
+    i2c::{FakeI2CHandle, FakeI2cConfig, I2CHandle, I2cHandleType},
     registry::ComponentRegistry,
 };
 
@@ -63,12 +47,9 @@ pub trait Board: Status {
     ) -> anyhow::Result<Rc<RefCell<dyn AnalogReader<u16, Error = anyhow::Error>>>>;
 
     /// Set the board to the indicated [PowerMode]
-    fn set_power_mode(
-        &self,
-        mode: PowerMode,
-        duration: Option<Duration>,
-    ) -> anyhow::Result<()>;
+    fn set_power_mode(&self, mode: PowerMode, duration: Option<Duration>) -> anyhow::Result<()>;
 
+    /// Get a wrapped [I2CHandle] by name.
     fn get_i2c_by_name(&self, name: String) -> anyhow::Result<I2cHandleType>;
 
     /// Return the amount of detected interrupt events on a pin. Should error if the
@@ -86,8 +67,9 @@ pub trait Board: Status {
     /// Get the PWM frequency of the pin
     fn get_pwm_frequency(&self, pin: i32) -> anyhow::Result<u64>;
 
-    /// Set the pin to the given PWM frequency (in Hz). 
-    /// When frequency is 0, the board will unregister the pin and PWM signal
+    /// Set the pin to the given PWM frequency (in Hz).
+    /// When frequency is 0, the board will unregister the pin and PWM channel from
+    /// the timer and removes the PWM signal.
     fn set_pwm_frequency(&mut self, pin: i32, frequency_hz: u64) -> anyhow::Result<()>;
 }
 
@@ -193,11 +175,7 @@ impl Board for FakeBoard {
         }
     }
 
-    fn set_power_mode(
-        &self,
-        mode: PowerMode,
-        duration: Option<Duration>,
-    ) -> anyhow::Result<()> {
+    fn set_power_mode(&self, mode: PowerMode, duration: Option<Duration>) -> anyhow::Result<()> {
         info!(
             "set power mode to {} for {} milliseconds",
             mode.as_str_name(),
@@ -297,11 +275,7 @@ where
         self.lock().unwrap().get_analog_reader_by_name(name)
     }
 
-    fn set_power_mode(
-        &self,
-        mode: PowerMode,
-        duration: Option<Duration>,
-    ) -> anyhow::Result<()> {
+    fn set_power_mode(&self, mode: PowerMode, duration: Option<Duration>) -> anyhow::Result<()> {
         self.lock().unwrap().set_power_mode(mode, duration)
     }
 
--- a/src/common/board.rs
+++ b/src/common/board.rs
@@ -77,17 +77,17 @@ pub trait Board: Status {
         anyhow::bail!("this board does not support digital interrupts")
     }
 
-    /// Get the pin's given duty cycle
+    /// Get the pin's given duty cycle, returns percentage as float between 0.0 and 1.0
     fn get_pwm_duty(&self, pin: i32) -> f64;
 
-    /// Set the pin to the given duty cycle 
+    /// Set the pin to the given duty cycle , `duty_cycle_pct` is a float between 0.0 and 1.0
     fn set_pwm_duty(&mut self, pin: i32, duty_cycle_pct: f64) -> anyhow::Result<()>;
 
     /// Get the PWM frequency of the pin
     fn get_pwm_frequency(&self, pin: i32) -> anyhow::Result<u64>;
 
     /// Set the pin to the given PWM frequency (in Hz). 
-    /// When frequency is 0, it will use the board’s default PWM frequency.
+    /// When frequency is 0, the board will unregister the pin and PWM signal
     fn set_pwm_frequency(&mut self, pin: i32, frequency_hz: u64) -> anyhow::Result<()>;
 }
 
--- a/src/common/board.rs
+++ b/src/common/board.rs
@@ -1,24 +1,38 @@
-#![allow(dead_code)]
-use crate::common::analog::AnalogReader;
-use crate::common::status::Status;
-use crate::google;
-use crate::proto::common;
-use crate::proto::component;
-use core::cell::RefCell;
-use log::*;
-use std::collections::HashMap;
-use std::rc::Rc;
-use std::sync::Arc;
-use std::sync::Mutex;
-use std::time::Duration;
+//! general-purpose compute board
 
-use super::analog::FakeAnalogReader;
-use super::config::ConfigType;
-use super::i2c::FakeI2CHandle;
-use super::i2c::FakeI2cConfig;
-use super::i2c::I2CHandle;
-use super::i2c::I2cHandleType;
-use super::registry::ComponentRegistry;
+#![allow(dead_code)]
+use crate::{
+    common::{
+        analog::AnalogReader,
+        status::Status,
+    },
+    google,
+    proto::{
+        common, 
+        component::board::v1::PowerMode
+    },
+};
+
+use log::*;
+use core::cell::RefCell;
+use std::{
+    collections::HashMap,
+    rc::Rc,
+    sync::Arc,
+    sync::Mutex,
+    time::Duration,
+};
+
+use super::{
+    analog::FakeAnalogReader,
+    config::ConfigType,
+    i2c::{
+        FakeI2CHandle, 
+        FakeI2cConfig, 
+        I2CHandle, 
+        I2cHandleType},
+    registry::ComponentRegistry,
+};
 
 pub static COMPONENT_NAME: &str = "board";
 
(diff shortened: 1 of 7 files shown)


src/common/board.rs
```diff
@@ -56,14 +70,28 @@ pub trait Board: Status {
     fn get_digital_interrupt_value(&self, _pin: i32) -> anyhow::Result<u32> {
         anyhow::bail!("this board does not support digital interrupts")
     }
+    /// Get the pin's given duty cycle
     fn get_pwm_duty(&self, pin: i32) -> f64;
+    /// Set the pin to the given duty cycle
```
gvaradarajan, line 75: Should specify here (and in L. 73) that the domain for duty cycle is a percent value float between 0 and 1

src/common/board.rs
```diff
@@ -56,14 +70,28 @@ pub trait Board: Status {
...
         anyhow::bail!("this board does not support digital interrupts")
     }
+    /// Get the pin's given duty cycle
     fn get_pwm_duty(&self, pin: i32) -> f64;
+    /// Set the pin to the given duty cycle 
     fn set_pwm_duty(&mut self, pin: i32, duty_cycle_pct: f64) -> anyhow::Result<()>;
+    /// Get the PWM frequency of the pin
     fn get_pwm_frequency(&self, pin: i32) -> anyhow::Result<u64>;
+    /// Set the pin to the given PWM frequency (in Hz). 
+    /// When frequency is 0, it will use the board’s default PWM frequency.
```
gvaradarajan, line 80: This is not true, setting the frequency to 0 unregisters the pin and PWM channel from the timer and removes the PWM signal

src/common/mod.rs
```diff
@@ -0,0 +1,61 @@
+//! Structs, traits, and utils to develop [component](https://docs.viam.com/components/)
```
gvaradarajan, line 1: Some of these are actually natively-included general purpose drivers (gpio_motor, adxl, mpu), so maybe we should note those separately?

src/lib.rs
```diff
@@ -1,88 +1,12 @@
...
-        pub mod sctp;
-    }
-    pub mod conn {
-        pub mod errors;
-        pub mod mdns;
-        pub mod server;
-        mod utils;
-    }
-}
+pub mod common;
```
gvaradarajan, line 1: I think this reorganization is definitely more intuitive at the very least, but it might be two big of a change as we're going into beta? We might want to hold off on this

gvaradarajan (changes requested):
Thought I'd put in a review anyway, but it's really up to @npmenard whether this goes in. I like the PR for the most part.

src/common/board.rs
mattjperez: Definitely something that should be included. Thanks for clarifying.

src/lib.rs
mattjperez: Good to know it feels more intuitive. 

While it definitely feels like a big change, it doesn't affect any of the main logic or imports.

There is an alternative though if you'd still like the docs without altering the structure.

https://doc.rust-lang.org/rustdoc/write-documentation/the-doc-attribute.html#the-doc-attribute

This would just be missing the compartmentalizing advantages of having inline documentation.

src/common/board.rs
```diff
@@ -31,23 +45,23 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
-    analogs: Vec<Rc<RefCell<dyn AnalogReader<u16, Error = anyhow::Error>>>>,
-    i2cs: HashMap<String, Arc<Mutex<FakeI2CHandle>>>,
-    pin_pwms: HashMap<i32, f64>,
-    pin_pwm_freq: HashMap<i32, u64>,
-}
+/// Represents the functionality of a general purpose compute board that contains various components such as analog readers and digital interrupts.
 pub trait Board: Status {
+    /// Set a pin to high or low
     fn set_gpio_pin_level(&mut self, pin: i32, is_high: bool) -> anyhow::Result<()>;
+    /// Return the current [BoardStatus] of the board
```
acmorrow, line 52: Very minor, but could these be offset by a blank line so the code and comments don't run together? Or will the rust formatter just shove them backtogether? Personally, I find it much more readable as:

```
    /// Set a pin to high or low
    fn set_gpio_pin_level(&mut self, pin: i32, is_high: bool) -> anyhow::Result<()>;

    /// Return the current [BoardStatus] of the board
    fn get_board_status(&self) -> anyhow::Result<common::v1::BoardStatus>;
```

It seems, at least, to be the style within the rust std lib: https://doc.rust-lang.org/src/core/alloc/mod.rs.html#164

mattjperez: Sure!

We may want to look at adding a `rustfmt.toml` file in the future to enforce these things. 

Though it seems this specifically is still unstable. [Link](https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#blank_lines_lower_bound)

src/common/board.rs
```diff
@@ -56,14 +70,28 @@ pub trait Board: Status {
...
     fn set_pwm_frequency(&mut self, pin: i32, frequency_hz: u64) -> anyhow::Result<()>;
 }
 
+/// An alias for a thread-safe handle to a struct that implements the [Board] trait
 pub type BoardType = Arc<Mutex<dyn Board>>;
 
+/// A test implementation of a generic compute board
+pub struct FakeBoard {
+    analogs: Vec<Rc<RefCell<dyn AnalogReader<u16, Error = anyhow::Error>>>>,
+    i2cs: HashMap<String, Arc<Mutex<FakeI2CHandle>>>,
```
randhid, line 90: we... maybe be getting rid of i2c altogether and having the sensors handle them. so I'd hold off on that  pr to go in before doing too much there - the i2c call is not part of our api.

randhid (changes requested):

src/common/board.rs
mattjperez: Ok. I don't think I touched anything specifically with i2c though, this change was just me moving `FakeBoard`'s struct definition below the `Board` trait definition, to group related things closer together.

mattjperez: Ahh I think I see what you're saying. The `Board` trait itself defines some functions for i2c that aren't part of the Board API. Yeah I was only putting docstrings on methods that I found equivalents for in the Python SDK, which I believe does accurately reflect the api.

src/common/board.rs
mattjperez: I looked at `esp32::board`'s implementation of this, and it doesn't unregister the PWM channel from the timer. It just sets the pin's `pwm_driver` to `None`. 

There is a `Drop` implementation for `PwmDriver` but I don't think it's ever actually dropped since `pin.pwm_driver` is `Option<PwmDriver<'static>>`. If so, we need to change how `0` is handled. Though this would be outside the scope of this PR.

src/common/board.rs
randhid: sort of - here's the ground truth for the [board API](https://github.com/viamrobotics/api/blob/main/proto/viam/component/board/v1/board.proto), python has a lot of client side-only helper functions, which might be going away.

src/common/board.rs
gvaradarajan: I've confirmed via print statement that it is in fact dropped

src/common/board.rs
acmorrow: Since it is just code motion though I don't think this PR should be blocked on any i2c work. Is that ok with you @randhid ?

src/common/board.rs
```diff
@@ -31,39 +45,63 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
     ) -> anyhow::Result<Rc<RefCell<dyn AnalogReader<u16, Error = anyhow::Error>>>>;
+
+    /// Set the board to the indicated [PowerMode]
     fn set_power_mode(
         &self,
-        mode: component::board::v1::PowerMode,
+        mode: PowerMode,
         duration: Option<Duration>,
     ) -> anyhow::Result<()>;
+
```
acmorrow, line 71: Looks like no doc comment for this function. Per the other discussion about i2c I recognize there may be upcoming changes, but it feels OK to me to document what exists here now, even if it ends up moving elsewhere later.

mattjperez: I agree

src/common/board.rs
randhid: totally okay.

randhid (approved):
LGTM

src/common/board.rs
```diff
@@ -1,24 +1,22 @@
...
 #![allow(dead_code)]
-use crate::common::analog::AnalogReader;
-use crate::common::status::Status;
-use crate::google;
-use crate::proto::common;
-use crate::proto::component;
+use crate::{
+    common::{analog::AnalogReader, status::Status},
+    google,
+    proto::{common, component::board::v1::PowerMode},
```
npmenard, line 7: nit: would just import proto::component so we know PowerMode is a generated type

src/common/board.rs
```diff
@@ -31,39 +29,61 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
-    i2cs: HashMap<String, Arc<Mutex<FakeI2CHandle>>>,
-    pin_pwms: HashMap<i32, f64>,
-    pin_pwm_freq: HashMap<i32, u64>,
-}
+/// Represents the functionality of a general purpose compute board that contains various components such as analog readers and digital interrupts.
 pub trait Board: Status {
+    /// Set a pin to high or low
     fn set_gpio_pin_level(&mut self, pin: i32, is_high: bool) -> anyhow::Result<()>;
+
+    /// Return the current [BoardStatus] of the board
```
npmenard, line 37: common::v1::BoardStatus

src/common/board.rs
```diff
@@ -31,39 +29,61 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
     /// Return the amount of detected interrupt events on a pin. Should error if the
     /// pin has not been configured as an interrupt
     fn get_digital_interrupt_value(&self, _pin: i32) -> anyhow::Result<u32> {
         anyhow::bail!("this board does not support digital interrupts")
     }
+
+    /// Get the pin's given duty cycle, returns percentage as float between 0.0 and 1.0
     fn get_pwm_duty(&self, pin: i32) -> f64;
+
+    /// Set the pin to the given duty cycle , `duty_cycle_pct` is a float between 0.0 and 1.0
```
npmenard, line 64: Can you add the condition when this function can fail?

src/common/board.rs
```diff
@@ -31,39 +29,61 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
     fn set_pwm_duty(&mut self, pin: i32, duty_cycle_pct: f64) -> anyhow::Result<()>;
+
+    /// Get the PWM frequency of the pin
```
npmenard, line 67: same

src/common/board.rs
```diff
@@ -31,39 +29,61 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
+    /// Set the pin to the given PWM frequency (in Hz).
+    /// When frequency is 0, the board will unregister the pin and PWM channel from
+    /// the timer and removes the PWM signal.
     fn set_pwm_frequency(&mut self, pin: i32, frequency_hz: u64) -> anyhow::Result<()>;
 }
 
+/// An alias for a thread-safe handle to a struct that implements the [Board] trait
 pub type BoardType = Arc<Mutex<dyn Board>>;
 
+/// A test implementation of a generic compute board
```
npmenard, line 79: nit: Emulate functionality of a Board for test purpose. 
I think we should maybe hide this from docs

npmenard (changes requested):
done with first pass

gvaradarajan (approved):
Needs a minor adjustment due to the two most recently merged PRs, so make sure to rebase in main and make the requisite changes. Otherwise LGTM

src/common/board.rs
mattjperez: To clarify, hide the description or hide the `FakeBoard` itself?
For the first we can just put a regular comment, for the second we can use `#[doc(hidden)]`

src/common/board.rs
mattjperez: In our `FakeBoard`'s impl or `Board`, it never fails and just returns `0.0` if the pin isn't present in the `pwm_pins` map. Should `Board::get_pwm_duty`'s description say it will return `0.0` or say that it's implementation specific?

src/common/board.rs
npmenard: let's just hide it altogether

src/common/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+pub mod i2c;
+pub mod log;
+pub mod math_utils;
+pub mod moisture_sensor;
+pub mod motor;
+pub mod movement_sensor;
+pub mod mpu6050;
+pub mod registry;
+pub mod robot;
+pub mod sensor;
```
gvaradarajan, line 50: needs `pub mod servo;` and `pub mod gpio_servo;`

gvaradarajan (changes requested):

src/common/mod.rs
mattjperez: had it in a local commit. pushed it

gvaradarajan (approved):
Once you rebase again, you will also need `pub mod power_sensor;` in common/mod.rs

npmenard (approved):
LGTM

npmenard (approved):
go for merge

mattjperez:
@viamrobotics/micro-rdk-team could I get help with the merge? Contributors don't have write access.

PR #106, opened 2023-10-30
Support for priority

npmenard:
Covers RSDK-5183 too.

Micro-RDK will now drop an existing webrtc connection if the configured priority of the offered SDP is higher than the current one.
By default it is set to 0 when no connection are opened


examples/sdkconfig.defaults
```diff
@@ -26,8 +26,7 @@ CONFIG_ESP32S3_SPIRAM_SUPPORT=n
 CONFIG_SPIRAM_SPEED_80M=y
 CONFIG_SPIRAM_IGNORE_NOTFOUND=y
 CONFIG_SPIRAM_USE_MALLOC=y
-CONFIG_HEAP_ABORT_WHEN_ALLOCATION_FAILS=y
-CONFIG_SPIRAM_MALLOC_ALWAYSINTERNAL=4096
+CONFIG_SPIRAM_MALLOC_ALWAYSINTERNAL=4090
```
npmenard, line 29: this is to avoid exhausting the internal ram when the SPIRAM is present

src/common/webrtc/api.rs
```diff
@@ -426,6 +431,17 @@ where
             .get(0)
             .ok_or_else(|| WebRtcError::InvalidSDPOffer("no media description".to_owned()))?;
 
+        let remote_prio = attribute
```
gvaradarajan, line 434: [q] what determines the priority number from the other side of the connection? what logic is meant to be used to determine it?

src/common/webrtc/api.rs
```diff
@@ -426,6 +431,17 @@ where
...
+            .attribute("x-priority")
+            .flatten()
+            .map_or(Ok(u32::MAX), |a| a.parse::<u32>())
+            .unwrap_or(u32::MAX);
+        let local_prio = local_prio.unwrap_or(0);
+
+        if local_prio >= remote_prio {
```
acmorrow, line 441: I think you could use `is_some_and` here to avoid needing to `unwrap_or` with a zero default into a shadowed `local_prio`.

```
if local_prio.is_some_and(|x| x >= remote_prio) {
    return Err(...);
```

src/common/conn/server.rs
```diff
@@ -345,22 +345,29 @@ where
...
+                        .and_then(|f| (!f.is_finished()).then_some(&current_prio))
+                        .unwrap_or(&None);
+
                     let sdp = api
-                        .answer()
+                        .answer(prio)
                         .await
                         .map_err(|e| ServerError::Other(Box::new(e)))?;
 
+                    if let Some(task) = self.webtrc_conn.take() {
```
acmorrow, line 360: It took me a moment to see the flow of control here, that if `answer` doesn't fail, it means we didn't reject the new connection based on priority and that we need to cancel (and await cancellation of) the current task. A brief comment before this block would probably go a long way for the next reader.

acmorrow (approved):
This LGTM. I have several comments, questions, and small suggestions, but nothing that would block merging this. Maybe getting an answer to whether new RC should unseat old?

examples/sdkconfig.defaults
npmenard: In the grpc layer we allocate 4096 as a work buffer and I want this buffer in the SPIRAM when present otherwise we are at risk of an allocation failure in the internal ram and if this happens in rust code that will panic immediately. 
Normally this should be allocated only once when the webrtc server is instantiated (after the connection is established) but there is a bug in the grpc layer that causes this buffer to be moved and reallocated. 

The panic on allocation failure is not needed actually and it causes the program to fail if an allocation fails in esp-idf code.

we will have to update the template too

src/common/webrtc/api.rs
npmenard: there is no logic for now, only the RC page will sent a priority value. I don't have an idea as too how we are going to deal with it moving forward

src/common/conn/server.rs
```diff
@@ -311,9 +311,9 @@ where
             .unwrap();
         let _ = self.app_client.insert(app_client);
         let cloned_robot = robot.clone();
-
+        let mut current_prio = None;
         loop {
-            let _ = smol::Timer::after(std::time::Duration::from_millis(100)).await;
+            let _ = smol::Timer::after(std::time::Duration::from_millis(300)).await;
```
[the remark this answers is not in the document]
npmenard: this is to further space out successive SDP offering. so we don't dedicate too much cpu time on them

src/common/webrtc/api.rs
```diff
@@ -410,7 +412,10 @@ where
         Err(WebRtcError::DataChannelOpenError())
     }
 
-    pub async fn answer(&mut self) -> Result<Box<WebRtcSdp>, WebRtcError> {
+    pub async fn answer(
+        &mut self,
+        local_prio: &Option<u32>,
```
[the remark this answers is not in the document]
npmenard: good point

src/common/webrtc/api.rs
npmenard: is_some_and is still unstable on rust 1.67 so we will have to push this until we reach 1.70 for micro-rdk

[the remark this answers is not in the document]
npmenard: it has to be inclusive, we reject new connection of equal priority

acmorrow (approved):

src/common/webrtc/api.rs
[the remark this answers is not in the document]
gvaradarajan: ^ I also am a little confused about this

PR #108, opened 2023-10-31
RSDK-5655 - fix i2c configuration bug for esp32

gvaradarajan:


gvaradarajan:
@randhid this fixes the issue where the MPU6050 doesn't configure properly and show up in the RC page

acmorrow (approved):
LGTM

randhid:
> @randhid this fixes the issue where the MPU6050 doesn't configure properly and show up in the RC page

Was this tested with hardware?

PR #109, opened 2023-11-01
RSDK-5566 - add actuator trait for is_moving API

gvaradarajan:
- stop was moved to the `Actuator` trait and `Stoppable` trait was removed as a consequence
- following component traits implement `Actuator`: `Motor`, `Base`
- everything that needs to implement `stop` also needed to implement `is_moving` and vice versa, but I provided default implementations that error with a message indicating the method is not supported. So, in theory something can implement the `Actuator` trait without implementing both `stop` and `is_moving`. This is possibly a mistake (or on the flip side, the entire `Actuator` trait is a mistake)


src/common/gpio_motor.rs
```diff
@@ -162,11 +162,14 @@ where
...
 }
 
-impl<M, Enc> Stoppable for EncodedMotor<M, Enc>
+impl<M, Enc> Actuator for EncodedMotor<M, Enc>
 where
     M: Motor,
     Enc: Encoder,
 {
+    fn is_moving(&mut self) -> anyhow::Result<bool> {
+        self.motor.is_moving()
```
acmorrow, line 171: It is unfortunate that we can't get a better answer in the case of an encoded motor, but we would need to have been sampling, so I don't think we can.

gvaradarajan:
tested on hardware

src/common/gpio_motor.rs
```diff
@@ -314,10 +317,13 @@ where
     }
 }
 
-impl<B> Stoppable for PwmABMotor<B>
+impl<B> Actuator for PwmABMotor<B>
 where
     B: Board,
 {
+    fn is_moving(&mut self) -> anyhow::Result<bool> {
+        Ok(self.board.get_pwm_duty(self.pwm_pin) != 0.0)
```
acmorrow, line 325: I discussed this a bit with @randhid about whether a stalled motor should return true for `is_moving`, and there isn't really any way to do so for a GPIO motor. Effectively, `is_moving` is being implemented as `is_powered` here. Rand's thought is that there should be some threshold, but as a separate item of work.

src/common/motor.rs
```diff
@@ -302,7 +302,7 @@ impl Status for FakeMotor {
     }
 }
 
-impl Stoppable for FakeMotor {
+impl Actuator for FakeMotor {
```
acmorrow, line 305: I was surprised to not see an `is_moving` for `FakeMotor`

src/common/motor.rs
```diff
@@ -375,7 +375,7 @@ impl Status for FakeMotorWithDependency {
     }
 }
 
-impl Stoppable for FakeMotorWithDependency {
+impl Actuator for FakeMotorWithDependency {
```
acmorrow, line 378: Same thoughts for `is_moving`.

src/esp32/base.rs
```diff
@@ -128,11 +128,14 @@ where
...
 
-impl<ML, MR> Stoppable for Esp32WheelBase<ML, MR>
+impl<ML, MR> Actuator for Esp32WheelBase<ML, MR>
 where
     ML: Motor,
     MR: Motor,
 {
+    fn is_moving(&mut self) -> anyhow::Result<bool> {
+        Ok(self.motor_left.is_moving()? || self.motor_right.is_moving()?)
+    }
```
acmorrow, line 138: Not part of this review, but it is interesting to me that for `stop` here, if we fail to stop the `motor_left` motor, we don't even try to stop the `motor_right` one.

src/common/base.rs
```diff
@@ -47,7 +48,10 @@ impl Base for FakeBase {
     }
 }
 
-impl Stoppable for FakeBase {
+impl Actuator for FakeBase {
```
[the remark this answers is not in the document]
gvaradarajan: this is pure laziness on my part, I'm just going to have it start tracking state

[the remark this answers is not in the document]
gvaradarajan: although, I just realized that you can't configure a FakeBase, so I might actually leave this to a different ticket

src/common/gpio_motor.rs
gvaradarajan: I'm not sure what you mean by this, do you mean determining if it's moving by sampling encoder position?

src/common/gpio_motor.rs
gvaradarajan: wait, I might be able to add the threshold in this PR. @randhid do you have a take on what a good threshold should be, or is it something we'd have to make configurable?

src/common/motor.rs
gvaradarajan: I actually just forgot, maybe this is a sign that I shouldn't include default implementations in the `Actuator` trait?

src/esp32/base.rs
gvaradarajan: yeah, this seems bad. I'll make a ticket for this

src/common/base.rs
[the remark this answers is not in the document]
mattjperez: Does the RDK `FakeBase` track state?

Also, would tracking state make this more of a `VirtualBase` than a `FakeBase`?

src/common/motor.rs
mattjperez: I think it's better not to have default for these because
1) The compiler will catch it if not implemented
2) We won't have to jump around between `common` and the specific implementation to see what the current implementation is

acmorrow (approved):
LGTM

src/common/gpio_motor.rs
randhid: default of 5% for now - configurable if someone complains in the future

randhid: or sets a min_power actually - we can leverage that if it exists for micro-rdk motors.

PR #111, opened 2023-11-06
RSDK-5599 - implement servo component

gvaradarajan:
included in this PR

- new `Servo` trait with registration functions and corresponding gRPC handlers for Viam Servo API
- implementation of a configurable generic servo using a PWM signal on a GPIO pin `GpioServo` (model name is "gpio")

### DIFF
--- a/src/common/gpio_servo.rs
+++ b/src/common/gpio_servo.rs
@@ -1,0 +1,308 @@
+//! A generalized servo implementation that uses a PWM signal on a GPIO pin
+//! via methods on a board component
+//!
+//! # Creating a GPIO pin servo and turning it 90 degrees
+//!
+//! ```ignore
+//!
+//! let board = FakeBoard::new(vec![]);
+//!
+//! let servo_settings = GpioServoSettings {
+//!     min_angle_deg: 0,
+//!     max_angle_deg: 180,
+//!     min_period_us: 500,
+//!     max_period_us: 2500,
+//!     pwm_resolution: 0,
+//!     frequency: 300,
+//! }
+//!
+//! let mut servo = GpioServo::new(board, 12, servo_settings);
+//!
+//! servo.move_to(90).unwrap()
+//!
+//! ```
+
+use std::sync::{Arc, Mutex};
+
+use super::{
+    actuator::Actuator,
+    board::{Board, BoardType},
+    config::{AttributeError, ConfigType},
+    registry::{get_board_from_dependencies, ComponentRegistry, Dependency},
+    servo::{Servo, ServoType},
+    status::Status,
+};
+
+/// Minimum and maximum period widths that should be safe limits for
+/// most servos. It is recommended you configure the servo with
+/// the limits provided by its datasheet if possible
+const SAFE_PERIOD_WIDTH_LIMITS: (u32, u32) = (500, 2500);
+/// Minimum and maximum angular positions that should be safe limits
+/// for most servos. It is recommended you configure the servo with
+/// the limits provided by its datasheet if possible
+const SAFE_ANGULAR_POSITION_LIMITS: (u32, u32) = (0, 180);
+/// Default PWM frequency that should be acceptable for most servos.
+/// It is recommended you configure the servo with the limits
+/// provided by its datasheet if possible
+const SAFE_DEFAULT_FREQUENCY_HZ: u32 = 300;
+
+pub(crate) fn register_models(registry: &mut ComponentRegistry) {
+    if registry.register_servo("gpio", &from_config).is_err() {
+        log::error!("gpio model is already registered")
+    }
+}
+
+pub(crate) fn from_config(
+    cfg: ConfigType,
+    dependencies: Vec<Dependency>,
+) -> anyhow::Result<ServoType> {
+    let board = get_board_from_dependencies(dependencies).ok_or(anyhow::anyhow!(
+        "configuration for 'gpio' servo is missing the required board dependency"
+    ))?;
+    let servo_settings = GpioServoSettings::from_config(&cfg)?;
+    let pin = cfg.get_attribute::<i32>("pin").map_err(|err| match err {
+        AttributeError::ConversionImpossibleError => {
+            anyhow::anyhow!(
+                "could not convert pin attribute to integer when configuring gpio servo"
+            )
+        }
+        _ => anyhow::anyhow!("error configuring gpio servo: {:?}", err),
+    })?;
+    Ok(Arc::new(Mutex::new(GpioServo::<BoardType>::new(
+        board.clone(),
+        pin,
+        servo_settings,
+    )?)))
+}
+
+#[derive(Debug)]
+pub(crate) struct GpioServoSettings {
+    pub min_angle_deg: u32,
+    pub max_angle_deg: u32,
+    pub min_period_us: u32,
+    pub max_period_us: u32,
+    pub frequency: u32,
+    /// when 0, pwm_resolution is not considered when calculating the PWM duty cycle
+    /// necessary to move the servo to particular angular position
+    pub pwm_resolution: u32,
+}
+
+impl GpioServoSettings {
+    pub fn from_config(cfg: &ConfigType) -> anyhow::Result<Self> {
+        let min_angle_deg = cfg
+            .get_attribute::<u32>("min_angle_deg")
+            .unwrap_or(SAFE_ANGULAR_POSITION_LIMITS.0);
+        let max_angle_deg = cfg
+            .get_attribute::<u32>("max_angle_deg")
+            .unwrap_or(SAFE_ANGULAR_POSITION_LIMITS.1);
+        let min_period_us = cfg
+            .get_attribute::<u32>("min_width_us")
+            .unwrap_or(SAFE_PERIOD_WIDTH_LIMITS.0);
+        let max_period_us = cfg
+            .get_attribute::<u32>("max_width_us")
+            .unwrap_or(SAFE_PERIOD_WIDTH_LIMITS.1);
+        let frequency = cfg
+            .get_attribute::<u32>("frequency_hz")
+            .unwrap_or(SAFE_DEFAULT_FREQUENCY_HZ);
+        let pwm_resolution = cfg
+            .get_attribute::<u32>("pwm_resolution")
+            .unwrap_or_default();
+        Ok(Self {
+            min_angle_deg,
+            max_angle_deg,
+            min_period_us,
+            max_period_us,
+            frequency,
+            pwm_resolution,
+        })
+    }
+}
+
+pub struct GpioServo<B> {
+    board: B,
+    pin: i32,
+    min_angle_deg: u32,
+    max_angle_deg: u32,
+    min_period_us: u32,
+    max_period_us: u32,
+    frequency: u32,
+    pwm_resolution: u32,
+}
+
+impl<B> GpioServo<B>
+where
+    B: Board,
+{
+    pub(crate) fn new(board: B, pin: i32, settings: GpioServoSettings) -> anyhow::Result<Self> {
+        if settings.frequency == 0 {
+            return Err(anyhow::anyhow!(
+                "PWM frequency cannot be zero for 'gpio' servo"
+            ));
+        }
+        let mut res = Self {
+            board,
+            pin,
+            min_angle_deg: settings.min_angle_deg,
+            max_angle_deg: settings.max_angle_deg,
+            min_period_us: settings.min_period_us,
+            max_period_us: settings.max_period_us,
+            frequency: settings.frequency,
+            pwm_resolution: settings.pwm_resolution,
+        };
+        res.board.set_pwm_frequency(pin, res.frequency as u64)?;
+        Ok(res)
+    }
+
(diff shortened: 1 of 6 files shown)


src/common/gpio_servo.rs
```diff
@@ -0,0 +1,191 @@
...
+    )?)))
+}
+
+#[derive(Debug)]
+pub struct GpioServoSettings {
+    pub min_angle_deg: f64,
+    pub max_angle_deg: f64,
+    pub min_period_us: u32,
+    pub max_period_us: u32,
+    pub frequency: u32,
```
mattjperez, line 73: some docs here would be useful. Does `*_us` mean microseconds? Is there a (in)valid range for `*_angle_deg`?

src/common/gpio_servo.rs
```diff
@@ -0,0 +1,191 @@
...
+
+impl GpioServoSettings {
+    pub fn from_config(cfg: &ConfigType) -> anyhow::Result<Self> {
+        let min_angle_deg = cfg
+            .get_attribute::<f64>("min_angle_deg")
+            .unwrap_or_default();
+        let max_angle_deg = cfg.get_attribute::<f64>("max_angle_deg").unwrap_or(180.0);
+        let min_period_us = cfg.get_attribute::<u32>("min_width_us").unwrap_or(500);
+        let max_period_us = cfg.get_attribute::<u32>("max_period_us").unwrap_or(2500);
+        let frequency = cfg.get_attribute::<u32>("frequency_hz").unwrap_or(300);
```
mattjperez, line 87: Is there a reason for these defaults? Are they just general settings that would work for most gpio servos?
Are these settings visible in Viam or is there a way for the user to see what the implicit settings are after creation?

src/common/gpio_servo.rs
```diff
@@ -0,0 +1,191 @@
...
+            max_period_us: settings.max_period_us,
+            frequency: settings.frequency,
+            pwm_resolution: settings.pwm_resolution,
+        };
+        res.board.set_pwm_frequency(pin, res.frequency as u64)?;
+        Ok(res)
+    }
+
+    fn angle_to_duty_pct(&self, angle_deg: u32) -> f64 {
+        let period = 1.0 / (self.frequency as f64);
```
mattjperez, line 133: what happens if frequency is 0? I don't think I saw anything above that prevents that.

src/common/servo.rs
```diff
@@ -0,0 +1,39 @@
+use std::sync::{Arc, Mutex};
+
+use super::{actuator::Actuator, status::Status};
+
+pub static COMPONENT_NAME: &str = "motor";
+
+pub trait Servo: Status + Actuator {
+    /// Moves the servo to an angular position of `angle_deg` away
+    /// from the home position
```
mattjperez, line 9: Is there an (in)valid range?

src/common/gpio_servo.rs
gvaradarajan: I feel like `_us` very commonly means microseconds. There's no invalid range, because it gets clamped later

src/common/servo.rs
gvaradarajan: this is driver dependent, but I'll add a comment on the gpio servo implementation of the method to explain the clamping behavior

src/common/gpio_servo.rs
gvaradarajan: My mistake, only the period internally gets clamped. I'll add validation and comments

gvaradarajan: I'm going to clamp it and document the behavior

src/common/gpio_servo.rs
```diff
@@ -154,7 +154,14 @@ impl<B> Servo for GpioServo<B>
 where
     B: Board,
 {
+    /// this implementation of move_to clamps the angle to the range determined
+    /// by min_angle_deg and max_angle_deg
     fn move_to(&mut self, angle_deg: u32) -> anyhow::Result<()> {
+        let angle_deg = match angle_deg {
```
mattjperez, line 160: I think you can actually use the [clamp](https://doc.rust-lang.org/core/cmp/trait.Ord.html#method.clamp) here.

acmorrow (commented):
Looks pretty good, some small comments here and there. Per our lunch discussion I know you have more changes coming so I'm sending you the comments as I have them now and I'll take a look at an update when you have it ready.

src/common/servo.rs
```diff
@@ -0,0 +1,39 @@
...
+where
+    A: ?Sized + Servo,
+{
+    fn move_to(&mut self, angle_deg: u32) -> anyhow::Result<()> {
+        self.lock().unwrap().move_to(angle_deg)
+    }
+    fn get_position(&mut self) -> anyhow::Result<u32> {
+        self.lock().unwrap().get_position()
+    }
+}
```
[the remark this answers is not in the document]
gvaradarajan: I agree, it would be nice to have something like that, I'll look into it

src/common/gpio_servo.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+#[allow(dead_code)]
+pub(crate) fn from_config(
+    cfg: ConfigType,
+    dependencies: Vec<Dependency>,
+) -> anyhow::Result<ServoType> {
+    let board = get_board_from_dependencies(dependencies);
+    if board.is_none() {
```
npmenard, line 47: can you use `ok_or` and a shortcut?

src/common/gpio_servo.rs
```diff
@@ -0,0 +1,195 @@
...
+    })?;
+    Ok(Arc::new(Mutex::new(GpioServo::<BoardType>::new(
+        board_unwrapped,
+        pin,
+        servo_settings,
+    )?)))
+}
+
+#[derive(Debug)]
+pub struct GpioServoSettings {
```
npmenard, line 68: should it be pub(crate)

src/common/gpio_servo.rs
```diff
@@ -0,0 +1,195 @@
...
+            pwm_resolution: settings.pwm_resolution,
+        };
+        res.board.set_pwm_frequency(pin, res.frequency as u64)?;
+        Ok(res)
+    }
+
+    fn angle_to_duty_pct(&self, angle_deg: u32) -> f64 {
+        let period = 1.0 / (self.frequency as f64);
+        let angle_range = self.max_angle_deg - self.min_angle_deg;
+        let period_range = self.max_period_us - self.min_period_us;
```
npmenard, line 135: can just force the type to be f64

npmenard (commented):
done with first pass

src/common/servo.rs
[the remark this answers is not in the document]
mattjperez: A pattern I've seen elsewhere, but would require a substantial amount of changes, is having the `Arc<Mutex<A>>` be a field internal to a struct,
`struct MyStruct(Arc<Mutex<Inner>>)`. 

Maybe a macro? @npmenard might know how viable that is.

src/common/gpio_servo.rs
```diff
@@ -0,0 +1,195 @@
...
+    actuator::Actuator,
+    board::{Board, BoardType},
+    config::{AttributeError, ConfigType},
+    registry::{get_board_from_dependencies, ComponentRegistry, Dependency},
+    servo::{Servo, ServoType},
+    status::Status,
+};
+
+pub(crate) fn register_models(registry: &mut ComponentRegistry) {
+    if registry.register_servo("gpio", &from_config).is_err() {
```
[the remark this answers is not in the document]
gvaradarajan: it happens automatically but where it happens depends on which part of the namespace we're talking about. This will have the API namespace "rdk:component:servo" and the model namespace "rdk:builtin:gpio". The "rdk" part of the API namespace is implicitly set in the component config's "namespace" key by app (this is temporary until the work for the new robot details page gets underway, long story that I can explain separately). The "rdk:builtin" portion is also prefixed to the model name by app. "servo" is added by the `register_servo` function and the `insert_resources` function on robot depending on whether you're registering the constructor or the actual component

src/common/gpio_servo.rs
```diff
@@ -0,0 +1,195 @@
...
+            max_angle_deg,
+            min_period_us,
+            max_period_us,
+            frequency,
+            pwm_resolution,
+        })
+    }
+}
+
+pub struct GpioServo<B> {
```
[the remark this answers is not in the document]
gvaradarajan: I'm honestly not sure, it's been the convention for when the type parameter has been board related since when I started contributing

src/common/gpio_servo.rs
```diff
@@ -0,0 +1,195 @@
...
+    board: B,
+    pin: i32,
+    min_angle_deg: u32,
```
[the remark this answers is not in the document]
gvaradarajan: It could, I didn't do this because the attributes are often referenced in the actual logic for a lot of the functions. I found that I either had to implement getters or keep writing `.servo_setttings.<attr>` for each of these things

The reason `GpioServoSettings` even exists is because clippy doesn't like more than 7 positional arguments to a function

src/common/gpio_servo.rs
```diff
@@ -0,0 +1,195 @@
...
+    /// necessary to move the servo to particular angular position
+    pub pwm_resolution: u32,
+}
+
+impl GpioServoSettings {
+    pub fn from_config(cfg: &ConfigType) -> anyhow::Result<Self> {
+        let min_angle_deg = cfg
+            .get_attribute::<u32>("min_angle_deg")
+            .unwrap_or_default();
+        let max_angle_deg = cfg.get_attribute::<u32>("max_angle_deg").unwrap_or(180);
```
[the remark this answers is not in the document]
gvaradarajan: yeah, I copied from RDK. I'd rather give actual reasons than refer to RDK so I'll ask around

src/common/gpio_servo.rs
```diff
@@ -0,0 +1,195 @@
...
+        let location_in_period = (pwm_width as u32) - self.min_angle_deg;
+        self.min_angle_deg + (location_in_period * angle_per_period)
+    }
+}
+
+impl<B> Servo for GpioServo<B>
+where
+    B: Board,
+{
+    /// this implementation of move_to clamps the angle to the range determined
```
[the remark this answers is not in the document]
gvaradarajan: oh maybe not? Honestly not sure what the convention is here

src/common/gpio_servo.rs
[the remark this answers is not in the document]
acmorrow: OK. I need to think about what the implications of that are for modular resources. Thanks for providing the details.

src/common/gpio_servo.rs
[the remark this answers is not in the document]
acmorrow: If it is `B` for `Board` then that's fine by me.

src/common/gpio_servo.rs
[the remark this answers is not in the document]
acmorrow: OK, that's reasonable, and I don't feel strongly about it.

> The reason GpioServoSettings even exists is because clippy doesn't like more than 7 positional arguments to a function.

I'm ... sort of with clippy on that one.

src/common/gpio_servo.rs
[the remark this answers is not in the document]
acmorrow: Would a user of our SDK ever benefit from reading this text in the documentation? If not, I'd downgrade to a regular comment.

npmenard (approved):
LGTM

src/common/gpio_servo.rs
```diff
@@ -0,0 +1,195 @@
...
+    if board.is_none() {
+        return Err(anyhow::anyhow!(
+            "actual board is required to be passed to configure a gpio servo"
+        ));
+    }
+    let board_unwrapped = board.unwrap();
+    let servo_settings = GpioServoSettings::from_config(&cfg)?;
+    let pin = cfg.get_attribute::<i32>("pin").map_err(|err| match err {
+        AttributeError::ConversionImpossibleError => {
+            anyhow::anyhow!("could not convert pin attribute format when configuring gpio servo")
```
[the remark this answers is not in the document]
acmorrow: @gvaradarajan - I won't insist on it, but if you feel like improving this error message before merging I think it would be nice for our users. Resolving this comment though, and no need to post an update should you chose to improve the error message.

src/common/gpio_servo.rs
acmorrow: @gvaradarajan - I'm interested in the answer to this.

src/common/gpio_servo.rs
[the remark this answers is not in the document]
acmorrow: Again, I wouldn't mind seeing this get done before merging if you think it can be non-doc comment, but no need to post an update for that.

acmorrow (approved):
LGTM mod some small tweaks that could be made before merge, and I'm still interested in an answer to Matt's question about dividing by zero.

src/common/gpio_servo.rs
gvaradarajan: I'll raise an error in the `new` function if the passed frequency is 0. I think past that, it's ok if this panics because someone would really have to go out of their way to make a `GpioServo` instance with a zero frequency

PR #112, opened 2023-11-08
RSDK-2980 - implement gpio AB motor

gvaradarajan:


src/common/motor.rs
```diff
@@ -192,10 +204,18 @@ impl TryFrom<&Kind> for MotorPinsConfig {
                 }
             },
         };
-        let pwm = value
-            .get("pwm")?
-            .ok_or_else(|| AttributeError::KeyNotFound("pwm".to_string()))?
-            .try_into()?;
+        let pwm = match value.get("pwm") {
+            Ok(opt) => match opt {
```
npmenard, line 208: would a Option::map work here?

src/common/gpio_motor.rs
```diff
@@ -437,3 +459,146 @@ where
...
+            b_pin,
+            max_rpm,
+            dir_flip,
+            is_on: false,
+            pwm_pin: a_pin,
+        };
+        // we start with this because we want to reserve a timer and PWM channel early
+        // for boards where these are a limited resource
+        res.board.set_pwm_frequency(a_pin, MOTOR_PWM_FREQUENCY)?;
+        res.board.set_pwm_duty(a_pin, 0.0)?;
```
npmenard, line 500: why not reserving a pwm pin for both?

src/common/gpio_motor.rs
```diff
@@ -117,9 +118,10 @@ pub(crate) fn gpio_motor_from_config(
 }
 
 // Motors generally don't care about the PWM frequency, so long as
-// it is in the order of magnitude 0f 10s of kHZ. For simplicity, we
-// just select 10kHz.
-const MOTOR_PWM_FREQUENCY: u64 = 10000;
```
npmenard, line 122: why did we move down to 1k?

src/common/gpio_motor.rs
```diff
@@ -437,3 +459,146 @@ where
...
+        } else {
+            (self.a_pin, self.b_pin)
+        };
+        if pwm_pin != self.pwm_pin {
+            self.board.set_pwm_frequency(pwm_pin, MOTOR_PWM_FREQUENCY)?;
+            self.board.set_pwm_frequency(self.pwm_pin, 0)?;
+        }
+        self.pwm_pin = pwm_pin;
+        self.board.set_gpio_pin_level(high_pin, true)?;
+        let power_pct = 1.0 - pct.abs();
```
npmenard, line 550: why do we need to do that?

npmenard (changes requested):
Done with first pass

acmorrow (approved):
I don't have anything to add beyond what @npmenard already identified, otherwise this LGTM once his concerns are addressed.

src/common/gpio_motor.rs
```diff
@@ -437,3 +459,146 @@ where
...
+
+impl<B> Status for AbMotor<B>
+where
+    B: Board,
+{
+    fn get_status(&self) -> anyhow::Result<Option<google::protobuf::Struct>> {
+        let mut hm = HashMap::new();
+        let pos = 0.0;
+        hm.insert(
+            "position".to_string(),
```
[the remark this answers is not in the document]
gvaradarajan: this is due to an RC page issue. If you don't supply a position key in Status for motor one of the following happens, but I don't remember which:

1. The motor card doesn't render
2. The position badge says NaN

I'll double-check. If it's the second, I'll just remove. If it's the first, I'll keep and file a bug with fleet

src/common/gpio_motor.rs
gvaradarajan: discussed offline

src/common/gpio_motor.rs
gvaradarajan: because we only ever need one PWM signal to run the motor (it's just that the pin the signal is running on changes), so I figured it was more economical to only reserve one timer

src/common/gpio_motor.rs
gvaradarajan: ok, I based this off rdk's example. I now see this was a mistake because RDK writes a single implementation for all three types of motors. I'll remove this and reverse the condition below

src/common/motor.rs
gvaradarajan: I wanted to do this but the problem I was having was dealing with the fact that `try_into` returns a result and not the actual value

gvaradarajan: I figured out how to do it, but do you find this clearer?
```rust
let pwm = value.get("pwm").map(
    |opt| opt.map(
        |val| i32::try_from(val)
    )
)?.transpose().or_else(|err| match err {
    AttributeError::KeyNotFound(_) => Ok(None),
    _ => {
        Err(err)
    }
})?;
```

npmenard (approved):
LGTM

PR #113, opened 2023-11-10
update webhook to use API key

npmenard:


src/esp32/webhook.rs
```diff
@@ -40,8 +40,10 @@ pub struct Webhook {
     fqdn: String,
     /// Endpoint that will be sent a GET request with credential information
     endpoint: Option<String>,
-    /// Location Secret used by the webhook's SDK script to connect
-    secret: Option<String>,
+    /// API key ID used by the webhook's SDK script to connect
+    api_key_id: Option<String>,
+    /// API key used by the webhook's SDK script to connect
+    api_key: Option<String>,
```
randhid, line 46: uhhhh I think fleet isn't doing this anymore. Can you check with emily and roxy?

randhid: unless this is the only secret that fleet 2 is exposing.

npmenard: i mean this is what the example code show since a couple a days. they are removing location secret

randhid: cool - I'm probably mistaken then.

PR #117, opened 2023-11-13
split would trigger reallocation so change it to split_off

npmenard:


acmorrow (approved):
LGTM per slack discussion

PR #120, opened 2023-11-16
RSDK-5725 - implement DoCommand support for all components

gvaradarajan:


src/common/grpc.rs
```diff
@@ -1042,6 +1156,27 @@ where
         self.encode_message(resp)
     }
 
+    fn power_sensor_do_command(&mut self, message: &[u8]) -> Result<(), ServerError> {
```
acmorrow, line 1159: It feels a little wrong to me to need 7 different implementations of this. Is there a way on robot that we could support uniformly getting the access to the `DoCommand` interface of these different types? As far as I can see the only actual difference between, say, `power_sensor_do_command` and `encoder_do_command` is the method called on `.robot`.

acmorrow (approved):
Overall this LGTM.

Per in-person discussion, it is a little unfortunate that the default needs to be manually injected into every leaf type, rather than attached at the trait level to e.g. `Sensor` and `Motor`. It feels unfortunate that we will push the requirement to adopt the default onto every module author. Perhaps @npmenard has some ideas about how to overcome the trait conflicts you encountered when you tried to do so.

Also, I'm curious about potential strategies that might reduce the need to re-implement the do command handling for each Trait-like that robot exposes.

npmenard:
Blanket implementation might do the trick here. But would likely mean we can specialize on a case by case basis, there might be a trick to it tough.
Otherwise this looks like a prime candidate for a Derive macro would make it more readable and easier less noisy on the code side.

trivial_generic/src/lib.rs
```diff
@@ -0,0 +1,29 @@
...
+        impl #impl_generics DoCommand for #name #ty_generics #where_clause {}
+    };
+    gen.into()
+}
+
+// #[cfg(test)]
+// mod tests {
+//     use super::*;
+
+//     #[test]
```
npmenard, line 24: cool if we can write a couple of tests

Cargo.toml
```diff
@@ -74,6 +74,7 @@ pin-project-lite = "0.2.9"
 url = "2.4.0"
 once_cell = "1.18.0"
 chrono = "0.4.31"
+trivial_generic = { path = "./trivial_generic" }
```
npmenard, line 77: can it live inside micro-rdk? maybe a separate macro folder?

src/common/adxl345.rs
```diff
@@ -12,6 +12,8 @@ use super::movement_sensor::MovementSensorType;
 use super::registry::{get_board_from_dependencies, ComponentRegistry, Dependency};
 use super::status::Status;
 
+use trivial_generic::TriviallyGeneric;
```
npmenard, line 15: probably needs a better name

npmenard (changes requested):
done with first pass

Cargo.toml
gvaradarajan: it can't, I think? proc-macros need to be their own crate

gvaradarajan: I interpreted this request as moving into src, which I was able to do, but I'm concerned it's no longer usable by modular drivers?

src/macros/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+use quote::quote;
+use syn;
+
+#[proc_macro_derive(DefaultDoCommand)]
+pub fn trivially_generic_derive(input: TokenStream) -> TokenStream {
+    let ast = syn::parse(input).unwrap();
+    impl_do_command(&ast)
+}
+
+fn impl_do_command(ast: &syn::DeriveInput) -> TokenStream {
```
acmorrow, line 11: What's the advantage of splitting this into two functions?

src/macros/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+use proc_macro::TokenStream;
+use quote::quote;
+use syn;
+
+#[proc_macro_derive(DefaultDoCommand)]
+pub fn trivially_generic_derive(input: TokenStream) -> TokenStream {
```
acmorrow, line 6: Is this name of this function meaningful? Why is it `trivial` / `generic`?

src/macros/src/lib.rs
```diff
@@ -0,0 +1,18 @@
...
+    let ast = syn::parse(input).unwrap();
+    impl_do_command(&ast)
+}
+
+fn impl_do_command(ast: &syn::DeriveInput) -> TokenStream {
+    let name = &ast.ident;
+    let (impl_generics, ty_generics, where_clause) = ast.generics.split_for_impl();
+    let gen = quote! {
+        impl #impl_generics DoCommand for #name #ty_generics #where_clause {}
```
acmorrow, line 15: Does `DoCommand` here need to be qualified somehow so that it is clear that it refers to the micro-rdk `DoCommand`? What would happen if I used this macro in a module?

src/macros/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+[lib]
+proc-macro = true
+
+[dependencies]
+quote = "1.0.33"
+syn = "2.0.39"
+
+[[test]]
+name = "do-command-tests"
+path = "tests/test.rs"
```
acmorrow, line 17: Please add a trailing newline here (and in the other file that is missing one).

src/macros/src/lib.rs
gvaradarajan: you have to import the DoCommand trait to use it, but I suppose I could replace `DoCommand` with `micro_rdk::common::generic::DoCommand`. Is that preferred?

acmorrow: I'm actually not sure if that is better or not. Because, at least in theory, the micro-rdk could actually be imported under a different name, with something like:
```
[dependencies]
micro-rdk = { git = "https://github.com/viamrobotics/micro-rdk.git", features = ["esp32"], package = "petit" }
```

 So maybe we don't actually know what the right top-level name is. Is there a way to use a relative name: `<magic>::common::generic::DoCommand` that indicates that we always want `common::generic::DoCommand` from our containing crate? `super`?

acmorrow: Yeah I think it is fine to require that `DoCommand` be imported. Not ideal, but fine. Resolving.

src/macros/tests/test.rs
```diff
@@ -0,0 +1,23 @@
...
+impl TestTrait for TestStruct {
+    fn test_fn(&self) -> u8 { 2 }
+}
+
+#[test]
+fn do_command_derive() {
+    let a = TestStruct {};
+    assert_eq!(a.do_command(), 1);
+    assert_eq!(a.test_fn(), 2);
+}
```
acmorrow, line 23: nit: missing newline

src/macros/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "macros"
```
acmorrow, line 2: Did you need to rename this so it didn't conflict at the top?

acmorrow (approved):
This LGTM, though I was expecting to see the `macros` crate renamed to `micro_rdk_macros` or similar.

src/common/adxl345.rs
```diff
@@ -6,11 +6,14 @@ use crate::google;
 
 use super::board::Board;
 use super::config::ConfigType;
+use super::generic::DoCommand;
 use super::i2c::I2CHandle;
 use super::movement_sensor::MovementSensorType;
 use super::registry::{get_board_from_dependencies, ComponentRegistry, Dependency};
 use super::status::Status;
 
+use macros::DefaultDoCommand;
```
npmenard, line 15: just DoCommand

src/macros/Cargo.toml
npmenard, line 2: I am going to be annoying, can it be at the top level? being a sub-crate inside src is weird.

npmenard, line 2: and can we name it micro-rdk-macros?

src/esp32/single_encoder.rs
```diff
@@ -23,6 +24,8 @@ use espsys::pcnt_evt_type_t_PCNT_EVT_H_LIM as pcnt_evt_h_lim;
 use espsys::pcnt_evt_type_t_PCNT_EVT_L_LIM as pcnt_evt_l_lim;
 use espsys::{esp, EspError, ESP_OK};
 
+use macros::DefaultDoCommand;
```
npmenard, line 27: can we make the macro gloabally available with #[macro_use] in lib.rs?

src/common/adxl345.rs
gvaradarajan: never mind, macro and trait usages never intersect so the compiler apparently infers appropriately

PR #122, opened 2023-11-27
RSDK-5729 - implement INA power sensor

gvaradarajan:
Hardware tested with INA219, driver supports [INA219](https://www.ti.com/lit/ds/symlink/ina219.pdf) and [INA226](https://www.ti.com/lit/ds/symlink/ina226.pdf)

### DIFF
--- a/src/common/ina.rs
+++ b/src/common/ina.rs
@@ -1,0 +1,230 @@
+/// Package ina implements ina power sensors to measure voltage, current, and power
+/// INA219 datasheet: https://www.ti.com/lit/ds/symlink/ina219.pdf
+/// INA226 datasheet: https://www.ti.com/lit/ds/symlink/ina226.pdf
+///
+/// The voltage, current and power can be read as
+/// 16 bit big endian integers from their given registers.
+/// This value is multiplied by the register LSB to get the reading in nanounits.
+///
+/// Voltage LSB: 1.25 mV for INA226, 4 mV for INA219
+/// Current LSB: maximum expected current of the system / (1 << 15)
+/// Power LSB: 25 * Current LSB for INA226, 20 * Current LSB for INA219
+///
+/// The calibration register is programmed to measure current and power properly.
+/// The calibration register is set to: calibratescale / (current_lsb * sense_resistor)
+use crate::common::i2c::I2CHandle;
+use core::fmt;
+use std::sync::{Arc, Mutex};
+
+use super::{
+    board::Board,
+    config::ConfigType,
+    generic::DoCommand,
+    i2c::I2cHandleType,
+    power_sensor::{Current, PowerSensor, PowerSensorType, PowerSupplyType, Voltage},
+    registry::{get_board_from_dependencies, ComponentRegistry, Dependency},
+    status::Status,
+};
+
+const DEFAULT_I2C_ADDRESS: u8 = 0x40;
+const DEFAULT_SHUNT_RESISTANCE_OHMS: f64 = 0.1;
+
+const INA_219_CALIBRATION_SCALE: f64 = 0.04096;
+const INA_226_CALIBRATION_SCALE: f64 = 0.00512;
+
+const DEFAULT_CONFIG_REGISTER_VALUE: u16 = 0x399F;
+const CALIBRATION_REGISTER: u8 = 0x05;
+const CONFIG_REGISTER: u8 = 0x00;
+const VOLTAGE_REGISTER: [u8; 1] = [0x02];
+const CURRENT_AMPERES_REGISTER: [u8; 1] = [0x04];
+const POWER_REGISTER: [u8; 1] = [0x03];
+
+pub(crate) fn register_models(registry: &mut ComponentRegistry) {
+    if registry
+        .register_power_sensor("ina219", &ina219_from_config)
+        .is_err()
+    {
+        log::error!("gpio model is already registered")
+    }
+    if registry
+        .register_power_sensor("ina226", &ina226_from_config)
+        .is_err()
+    {
+        log::error!("gpio model is already registered")
+    }
+}
+
+fn ina219_from_config(cfg: ConfigType, deps: Vec<Dependency>) -> anyhow::Result<PowerSensorType> {
+    Ok(Arc::new(Mutex::new(from_config(Model::Ina219, cfg, deps)?)))
+}
+
+fn ina226_from_config(cfg: ConfigType, deps: Vec<Dependency>) -> anyhow::Result<PowerSensorType> {
+    Ok(Arc::new(Mutex::new(from_config(Model::Ina226, cfg, deps)?)))
+}
+
+fn from_config(
+    model: Model,
+    cfg: ConfigType,
+    dependencies: Vec<Dependency>,
+) -> anyhow::Result<Ina<I2cHandleType>> {
+    let i2c_address = cfg
+        .get_attribute::<u8>("i2c_address")
+        .unwrap_or(DEFAULT_I2C_ADDRESS);
+
+    let default_max_current_amperes = match &model {
+        Model::Ina219 => 3.2,
+        Model::Ina226 => 20.0,
+    };
+    let max_current_amperes = cfg
+        .get_attribute::<f64>("max_current_amps")
+        .unwrap_or(default_max_current_amperes);
+    let max_current_nano_amperes = (max_current_amperes * 1e9) as i64;
+
+    let shunt_resistance_ohms = cfg
+        .get_attribute::<f64>("shunt_resistance")
+        .unwrap_or(DEFAULT_SHUNT_RESISTANCE_OHMS);
+    let shunt_resistance_nano_ohms = (shunt_resistance_ohms * 1e9) as i64;
+
+    let i2c_name = cfg
+        .get_attribute::<String>("i2c_bus")
+        .map_err(|_| anyhow::anyhow!("i2c_bus is a required config attribute for {}", model))?;
+    let board = get_board_from_dependencies(dependencies).ok_or(anyhow::anyhow!(
+        "a board must be configured to configure an {} sensor",
+        model
+    ))?;
+    let i2c_handle = board.get_i2c_by_name(i2c_name)?;
+
+    Ina::new(
+        model,
+        i2c_handle,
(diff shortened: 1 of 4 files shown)


src/common/ina.rs
```diff
@@ -0,0 +1,246 @@
...
+const INA_219_CALIBRATION_SCALE: f64 = 0.04096;
+const INA_226_CALIBRATION_SCALE: f64 = 0.00512;
+
+const DEFAULT_CONFIG_REGISTER_VALUE: u16 = 0x399F;
+const CONFIG_REGISTER: u8 = 0x00;
+const VOLTAGE_REGISTER: [u8; 1] = [0x02];
+const CURRENT_AMPERES_REGISTER: [u8; 1] = [0x04];
+const POWER_REGISTER: [u8; 1] = [0x03];
+
+pub(crate) fn register_models(registry: &mut ComponentRegistry) {
```
acmorrow, line 40: An annoying question, I'll admit, but: given that we now have rudimentary module support for micro-rdk, do the INA power sensors belong as builtins, or should they be in a separate module?

Cargo.toml
```diff
@@ -74,6 +74,7 @@ pin-project-lite = "0.2.9"
 url = "2.4.0"
 once_cell = "1.18.0"
 chrono = "0.4.31"
+ignore = "=0.4.20"
```
acmorrow, line 77: Just curious, but why the exact version pin?

src/common/mod.rs
```diff
@@ -24,6 +24,7 @@
 //! - [gpio_motor]
 //! - [adxl345]
 //! - [mpu6050]
+//! - [ina]
```
acmorrow, line 27: The other sections look alpha sorted, so this section probably should be too, even though it wasn't before.

src/common/power_sensor.rs
```diff
@@ -49,12 +49,12 @@ impl From<Current> for component::power_sensor::v1::GetCurrentResponse {
 }
 
 pub trait PowerSensor: Status {
-    fn get_voltage(&self) -> anyhow::Result<Voltage>;
+    fn get_voltage(&mut self) -> anyhow::Result<Voltage>;
```
acmorrow, line 52: What in the ina implementation forced your hand here?

src/common/ina.rs
```diff
@@ -0,0 +1,246 @@
...
+const DEFAULT_SHUNT_RESISTANCE_OHMS: f64 = 0.1;
+
+const INA_219_CALIBRATION_SCALE: f64 = 0.04096;
+const INA_226_CALIBRATION_SCALE: f64 = 0.00512;
+
+const DEFAULT_CONFIG_REGISTER_VALUE: u16 = 0x399F;
+const CONFIG_REGISTER: u8 = 0x00;
+const VOLTAGE_REGISTER: [u8; 1] = [0x02];
+const CURRENT_AMPERES_REGISTER: [u8; 1] = [0x04];
+const POWER_REGISTER: [u8; 1] = [0x03];
```
acmorrow, line 38: Why single element arrays for these registers?

src/common/ina.rs
```diff
@@ -0,0 +1,246 @@
...
+    }
+    if registry
+        .register_power_sensor("ina226", &ina226_from_config)
+        .is_err()
+    {
+        log::error!("gpio model is already registered")
+    }
+}
+
+pub(crate) fn ina219_from_config(
```
acmorrow, line 55: Do these two `x_from_config` functions actually need to be `pub` of any sort at all?

src/common/ina.rs
```diff
@@ -0,0 +1,246 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let model_str = match self {
+            Model::Ina219 => "INA219",
+            Model::Ina226 => "INA226",
+        };
+        write!(f, "{model_str}")
+    }
+}
+
+pub struct Ina<H> {
```
acmorrow, line 126: - Why is this generic over H if the only H that is impl'd is for `I2CHandle`?
- Does it need to be `pub`?

src/common/ina.rs
```diff
@@ -0,0 +1,246 @@
...
+        model,
+        i2c_handle,
+        i2c_address,
+        max_current_nano_amperes,
+        shunt_resistance_nano_ohms,
+    )
+}
+
+#[derive(Clone, Copy)]
+pub enum Model {
```
acmorrow, line 111: Does it need to be `pub`?

src/common/ina.rs
```diff
@@ -0,0 +1,246 @@
...
+                (INA_219_CALIBRATION_SCALE, power_lsb)
+            }
+            Model::Ina226 => (INA_226_CALIBRATION_SCALE, 25 * current_reading_lsb),
+        };
+        let calibration_scale = (unadjusted_calibration_scale
+            / ((current_reading_lsb as f64) * (shunt_resistance_nano_ohms as f64 * 1e-9)))
+            as i64;
+        if calibration_scale >= (1 << 16) {
+            anyhow::bail!(
+                "ina calibrate: calibration register value invalid {:?}",
```
acmorrow, line 159: Maybe `"ina calibrate: invalid calibration scale value {:?} exceeds limit 1 << 16"`. I'm not sure though, since it is a computed value, so it isn't immediately clear how the user should remedy it.

src/common/ina.rs
```diff
@@ -0,0 +1,246 @@
...
+                calibration_scale
+            );
+        }
+        // let address: [u8; 1] = [0x05];
```
acmorrow, line 163: Remove

src/common/ina.rs
```diff
@@ -0,0 +1,246 @@
...
+            power_reading_lsb,
+        })
+    }
+
+    fn calibrate(&mut self) -> anyhow::Result<()> {
+        // set scaling factor for current and power registers by writing adjusted
+        // calibration scale to the appropriate register
+        let mut calibration_scale_bytes: [u8; 2] = [0; 2];
+        calibration_scale_bytes.copy_from_slice(&self.calibration_scale_bytes);
+        self.write_to_register(0x05, &calibration_scale_bytes)?;
```
acmorrow, line 180: The register should probably be in a named constant.

src/common/ina.rs
```diff
@@ -0,0 +1,246 @@
...
+        // reset the config register to default values
```
acmorrow, line 181: This comment says what it is doing, but not why.

src/common/ina.rs
```diff
@@ -0,0 +1,246 @@
...
+    H: I2CHandle,
+{
+    fn new(
+        model: Model,
+        i2c_handle: H,
+        i2c_address: u8,
+        max_current_nano_amperes: i64,
+        shunt_resistance_nano_ohms: i64,
+    ) -> anyhow::Result<Self> {
+        let current_reading_lsb = max_current_nano_amperes / (1 << 15);
```
acmorrow, line 146: How carefully do you want me to review the math and logic that connects to the sensor specs? Ditto register usage, etc. Or was this largely cribbed from another ina sensor impl so you have good confidence in your translation to micro-rdk?

src/common/ina.rs
```diff
@@ -0,0 +1,246 @@
...
+        };
+        Ok(Voltage {
+            volts,
+            power_supply_type: PowerSupplyType::DC,
+        })
+    }
+
+    fn get_current(&mut self) -> anyhow::Result<Current> {
+        let current_reading_lsb = self.max_current_nano_amperes / (1 << 15);
+        self.calibrate()?;
```
acmorrow, line 213: Is it necessary to calibrate on every call to `get_current` and `get_power`?

gvaradarajan: potentially not, I cribbed this part from RDK as well. But RDK might have been doing this to accommodate reconfiguration of something like that, so I'll try removing and seeing what happens.

src/common/ina.rs
acmorrow: OK, so I will not go over the detailed logic here too carefully.

src/common/ina.rs
gvaradarajan: I think we should we keep as builtin drivers for now and move them out when we have an easier build process for modules?

src/common/ina.rs
```diff
@@ -123,19 +118,15 @@ impl fmt::Display for Model {
     }
 }
 
-pub struct Ina<H> {
+struct Ina<H: I2CHandle> {
```
acmorrow, line 121: I'm still a little puzzled here, and I think it is mostly my issue re rust comprehension,  but how can `H` ever end up being something other than `I2cHandleType` here? The only caller of `Ina<H>::new` is `from_config`, and it always calls it with the result of calling `Board::get_i2c_by_name`, which returns `pub type I2cHandleType = Arc<Mutex<dyn I2CHandle + Send>>;`, which happens to *also* be an `I2CHandle`. Why can't the `i2c_handle` field just be declared to be of type `I2cHandleType`?

src/common/ina.rs
npmenard: I don't think we have a clear view on whether or not we need these sensor as builtin, it would help having them since it make micro-rdk more versatile out of the box. As gautham said not having a process where people can generate an image just by adding modules makes it unlikely they will ever be used at this stage, plus I would like to have a stronger CI to protect against regression and module that start to fail to built against micro-rdk as versions get released. We are still changing interfaces and all after all.

acmorrow: @npmenard - OK, sounds good. Resolving this comment.

src/common/ina.rs
gvaradarajan: it can be, I wanted to be able to support any implementation of the `I2CHandle` trait, but now that nothing is public anymore you're correct.

gvaradarajan: resolved offline

acmorrow (approved):
LGTM, but I'd like to know @npmenard's thoughts on the `ignore` exact version pin.

npmenard:
> LGTM, but I'd like to know @npmenard's thoughts on the `ignore` exact version pin.

It's being pulled indirectly by other crates but it's not compatible with rust 1.67 we have to upgrade our minimum rust version before removing it

PR #123, opened 2023-11-27
RSDK-5795 - implement ResourceRPCSubtypes so the main robot part stops complaining

gvaradarajan:


acmorrow (approved):
Looks reasonable to me.

npmenard:
As far as I recall ResourceRPCSubtypes is used to track resource that a main server may not understand so that any RPC call are passed through to the resource. Has it changed?
If so why micro-rdk needs this function? If viam-server complains then it should be the burden of the viam server to interpret unimpleted as do not use this call again. Like Sessions are handled on our side

PR #125, opened 2023-11-30
Remove static config and Kind trait conversion

npmenard:
This covers RSDK-5875 and RSDK-2321, plus a small refactor of the config logic

### DIFF
--- a/examples/esp32/partitions.csv
+++ b/examples/esp32/partitions.csv
@@ -1,5 +1,5 @@
 # Name,   Type, SubType, Offset,  Size, Flags
 # Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
-nvs,      data, nvs, 0x9000    ,        0x8000,
-phy_init, data, phy, 0x11000    ,        0x1000,
-factory,  app,  factory, 0x30000,       0x3D0000,
+nvs,      data, nvs, 0x9000    ,        0x6000,
+phy_init, data, phy, 0xf000    ,        0x1000,
+factory,  app,  factory, 0x10000,       0x3effff,
--- a/src/common/config.rs
+++ b/src/common/config.rs
@@ -32,20 +32,7 @@ macro_rules! primitives
 {
     ( $($t:ty),* ) =>
     {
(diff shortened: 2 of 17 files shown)


src/common/config.rs
```diff
@@ -399,35 +294,26 @@ impl TryFrom<&ComponentConfig> for DynamicComponentConfig {
 }
 
 #[derive(Debug)]
-pub enum ConfigType {
-    Static(&'static StaticComponentConfig),
-    Dynamic(DynamicComponentConfig),
+pub enum ConfigType<'a> {
```
gvaradarajan, line 297: should we just remove the enum, use `DynamicComponentConfig` directly and rename to `ComponentConfig`?

examples/esp32/partitions.csv
```diff
@@ -1,5 +1,5 @@
 # Name,   Type, SubType, Offset,  Size, Flags
 # Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
-nvs,      data, nvs, 0x9000    ,        0x8000,
-phy_init, data, phy, 0x11000    ,        0x1000,
-factory,  app,  factory, 0x30000,       0x3D0000,
+nvs,      data, nvs, 0x9000    ,        0x6000,
+phy_init, data, phy, 0xf000    ,        0x1000,
```
gvaradarajan, line 4: [q] how often should we check that the updated compiled binary fits in the factory partition?

gvaradarajan (changes requested):
LGTM, I'm glad we're finally doing this. Just one adjustment I think worth doing

examples/esp32/partitions.csv
npmenard: We should be mindful we don't want the binary to bloat even in Debug mode

acmorrow (approved):
LGTM. I went through it pretty quickly because it is mostly a feature removal and associated simplifications.

src/common/config.rs
gvaradarajan: understood, I'm comfortable with that

gvaradarajan (approved):
LGTM

PR #127, opened 2023-12-01
RSDK-5707 - fix units bug in MPU6050 driver

gvaradarajan:


src/common/mpu6050.rs
```diff
@@ -30,6 +46,7 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
 
 const READING_START_REGISTER: u8 = 59;
 const STANDBY_MODE_REGISTER: u8 = 107;
+const MAX_U16: f64 = 32768.0;
```
acmorrow: Is there a rust intrinsic for this? Sort of like C++ has `std::numeric_limits<T>::max`?

src/common/mpu6050.rs
```diff
@@ -98,18 +115,17 @@ impl Drop for MPU6050 {
 }
 
 fn get_angular_velocity_from_reading(reading: &[u8; 14]) -> Vector3 {
-    let (x_bytes, y_z_bytes) = reading[8..14].split_at(size_of::<u16>());
-    let unscaled_x = u16::from_be_bytes(x_bytes.try_into().unwrap());
-    let (y_bytes, z_bytes) = y_z_bytes.split_at(size_of::<u16>());
-    let unscaled_y = u16::from_be_bytes(y_bytes.try_into().unwrap());
-    let unscaled_z = u16::from_be_bytes(z_bytes.try_into().unwrap());
+    let (x_bytes, y_z_bytes) = reading[8..14].split_at(size_of::<i16>());
```
acmorrow, line 118: Since it looks like there was also a signed-ness bug here, and not just a scaling bug, can you please add a test (one that fails without this change, but now passes).

acmorrow (approved):
LGTM but please add a test that covers the signed-ness issues too.

PR #128, opened 2023-12-04
RSDK-5740 - change esp32_wheeled_base to two_wheeled_base

gvaradarajan:
moved our wheeled_base into common and changed the model name to "micro_rdk_wheeled_base"

Will comment here when tested on hardware, but I didn't really change anything.

### DIFF
--- a/src/common/wheeled_base.rs
+++ b/src/common/wheeled_base.rs
@@ -13,22 +13,22 @@ use std::sync::{Arc, Mutex};
 pub(crate) fn register_models(registry: &mut ComponentRegistry) {
     if registry
         .register_base(
-            "micro_rdk_wheeled_base",
+            "two_wheeled_base",
             &WheeledBase::<MotorType, MotorType>::from_config,
         )
         .is_err()
     {
-        log::error!("micro_rdk_wheeled_base model is already registered")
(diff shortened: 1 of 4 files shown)


src/common/wheeled_base.rs
```diff
@@ -14,18 +13,18 @@ use std::sync::{Arc, Mutex};
 pub(crate) fn register_models(registry: &mut ComponentRegistry) {
     if registry
         .register_base(
-            "esp32_wheeled_base",
-            &Esp32WheelBase::<MotorType, MotorType>::from_config,
+            "micro_rdk_wheeled_base",
```
acmorrow, line 16: Why put the prefix in at all, rather than just "wheeled_base"? Or even just `wheeled` to match RDK?

acmorrow (approved):

src/common/wheeled_base.rs
gvaradarajan: you're right, I'm actually going to name it "two-wheeled base". Originally I was trying to potentially accommodate https://viam.atlassian.net/browse/RSDK-5412. But since we're hardcoding the models for that ticket it doesn't matter

acmorrow (approved):

PR #129, opened 2023-12-05
[RSDK-5643] return raw readings when reading analog values

npmenard:


src/esp32/analog.rs
```diff
@@ -19,22 +23,36 @@ where
...
+        self.raw_reading()
             .map_err(|e| anyhow::anyhow!(format!("error while reading analog reader {e}")))
     }
     fn inner_name(&self) -> String {
         self.name.clone()
     }
+    fn raw_reading(&mut self) -> anyhow::Result<u16> {
+        let unit = T::Adc::unit();
+        let channel = self.pin.0;
+        if unit == adc_unit_t_ADC_UNIT_1 {
```
acmorrow, line 45: Maybe invert this conditional to handle the error case first:

```
if unit != adc_unit ... {
    return Err(...)
}

let measurement = ...
```

src/esp32/analog.rs
```diff
@@ -19,22 +23,36 @@ where
...
+            let measurement = unsafe { adc1_get_raw(channel) };
```
acmorrow, line 46: Why doesn't `AdcDriver::read_raw` do what you need here?

acmorrow (commented):
In my own words, my best understanding of what is happening here is that we used to just use `AdcDriver::read` which returned a reading in millivolts against calibration settings, but per the ticket we really just want the raw reading to be consistent with RDK, so dropping down to `adc1_get_raw` directly. But I'm curious why `AdcDriver::read_raw` couldn't be used, as it seems to arrive at `adc1_get_raw` anyway: https://esp-rs.github.io/esp-idf-hal/src/esp_idf_hal/adc.rs.html#298.

npmenard:
> In my own words, my best understanding of what is happening here is that we used to just use `AdcDriver::read` which returned a reading in millivolts against calibration settings, but per the ticket we really just want the raw reading to be consistent with RDK, so dropping down to `adc1_get_raw` directly. But I'm curious why `AdcDriver::read_raw` couldn't be used, as it seems to arrive at `adc1_get_raw` anyway: https://esp-rs.github.io/esp-idf-hal/src/esp_idf_hal/adc.rs.html#298.
> 
> 

You're right I missed the function altogether -_-

src/esp32/analog.rs
npmenard: I withdraw my comment, we would need to update esp-idf-hal which will need us to update to rust 1.71 which snowballs into an image that will not fit anymore in the 4MB address space

acmorrow: Ah, that's unfortunate. Maybe leave a comment in the hopes that we can revisit later?

acmorrow (approved):

acmorrow (approved):

PR #130, opened 2023-12-07
RSDK-5787 - implement installer option to print monitor logs to file

gvaradarajan:
corresponding commit on fork of espflash here: https://github.com/viamrobotics/espflash/commit/8f8e58b8b1f5509d2c406a98e6e92a6a72f10140

### DIFF
--- a/micro-rdk-installer/Cargo.lock
+++ b/micro-rdk-installer/Cargo.lock
@@ -3,27 +3,6 @@
 version = 3
 
 [[package]]
-name = "CoreFoundation-sys"
-version = "0.1.4"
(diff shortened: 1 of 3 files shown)


npmenard (approved):
LGTM. Right not we are truncating the content of the logfile which I think we should avoid. Also could be add a separator line with date and time so we can track roughly when the logs where generated?

gvaradarajan:
Note: A follow-up [commit](url) on our espflash fork was required to address @npmenard 's requested changes

npmenard (approved):
LGTM

PR #133, opened 2023-12-14
Updating PWM logic

npmenard:
There were a couple of issues with the first iteration of the PWM code. Firstly the order of argument of ` ledc_bind_channel_timer` was incorrect (thanks C...)  The other issue came from the way the underlying Ledc driver select the clock source when  timer is created. It will most likely select a slow clock when PWM frequency < 1000 are set which prevent a user from switching to a higher frequency. Now we are rebuilding the timer when changing frequency

### DIFF
--- a/src/esp32/pwm.rs
+++ b/src/esp32/pwm.rs
@@ -257,16 +257,35 @@ impl<'a> LedcManager<'a> {
     }
 
     fn next_available_timer(&mut self, frequency_hz: u32) -> Result<usize, Esp32PwmError> {
+        // Timer with same frequency exist?
         if let Some(timer) = self.find_timer_by_frequency(frequency_hz) {
             return Ok(timer);
         }
-        let timer_number = self
+        // Free Timer?
+        let res = self
             .timer_allocation
             .iter()
             .enumerate()
             .find_map(|(i, t)| if t.count == 0 { Some(i) } else { None })
-            .ok_or(Esp32PwmError::NoTimersAvailable)?;
-        self.timer_allocation[timer_number].set_frequency(frequency_hz)?;
+            .ok_or(Esp32PwmError::NoTimersAvailable);
+        let timer_number = match res {
+            Ok(t) => {
+                self.timer_allocation[t].set_frequency(frequency_hz)?;
+                t
+            }
+            // if no timer are free then match with the nearest pwm frequency
+            Err(_) => self
+                .timer_allocation
+                .iter()
+                .enumerate()
+                .min_by(|(_, a), (_, b)| {
+                    (a.frequency as i32 - frequency_hz as i32)
+                        .abs()
+                        .cmp(&(b.frequency as i32 - frequency_hz as i32).abs())
+                })
+                .map(|(idx, _)| idx)
+                .unwrap(),
+        };
         Ok(timer_number)
     }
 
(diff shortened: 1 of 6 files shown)


src/esp32/pwm.rs
```diff
@@ -6,31 +7,21 @@ use esp_idf_hal::ledc::{
...
-    PwmDriver::new(pin, chan, starting_frequency_hz)
-}
-
 #[derive(Debug, Error)]
 pub enum Esp32PwmError {
     #[error("{0}")]
     EspError(EspError),
-    #[error("only 4 different PWM frequencies allowed, available freq: {0}, {1}, {2}, {3}")]
-    NoTimersAvailable(u32, u32, u32, u32),
+    #[error("all timer are used try a different frequency")]
```
gvaradarajan, line 23: are the currently used frequencies not valuable information when receiving this error?

gvaradarajan (approved):
wow this is a much better implementation than what was there before, I learned a lot from this PR

src/esp32/pwm.rs
```diff
@@ -209,123 +120,241 @@ impl<'a> PwmDriver<'a> {
...
-        if let Some(index) = timer_channels.iter().position(|&x| x == channel) {
-            timer_channels.swap_remove(index);
-            if self.timer_channels[timer_number].is_empty() {
-                self.timer_frequencies[timer_number] = None;
-            };
-        };
-    }
+        self.timer_allocation[timer_number].dec();
+        let res = self.timer_allocation[timer_number].set_frequency(frequency_hz);
+        self.timer_allocation[timer_number].inc();
```
acmorrow, line 286: Should this `inc` happen even if this function ultimately returns a `NoTimersAvailable` error, down on 299?

acmorrow (approved):
LGTM. One tiny nit and a question about something that looked maybe a little suspicious to me.

src/esp32/pwm.rs
mattjperez: Also, isn't it already being incremented within `bind_channel_to_timer` at Line 329?

src/esp32/pwm.rs
```diff
@@ -142,20 +79,15 @@ pub(crate) struct PwmDriver<'a> {
...
-    ) -> Result<PwmDriver<'a>, Esp32PwmError> {
+    pub fn new(pin: AnyIOPin, starting_frequency_hz: u32) -> Result<PwmDriver<'a>, Esp32PwmError> {
         let mut ledc_manager = LEDC_MANAGER.lock().unwrap();
-        let timer_number = ledc_manager.take_timer(starting_frequency_hz, channel)?;
-        let ledc_driver =
-            get_ledc_driver_by_channel(channel, timer_number, starting_frequency_hz, pin)?;
-
+        let channel = ledc_manager.allocate_pin(pin.pin(), starting_frequency_hz)?;
+        let timer = ledc_manager.get_configure_timer_instance(channel.1);
+        let ledc_driver = get_ledc_driver_by_channel(channel.0, timer, pin).unwrap();
```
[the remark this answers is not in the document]
mattjperez: I kinda like that it's separated out. This function is very readable as it is, and `get_ledc_driver_by_channel` is very descriptive. It also isolates a little bit of the unsafe calls to just there. 

We could also add `#[inline(always)]` if you want to strongly urge the compiler to do so.

src/esp32/pwm.rs
npmenard: It is indeed, however since in RDK (at least for the pi) we do not error out when configuring an infeasible PWM frequency then a user will most likely never see such error. I will add some logic to find the nearest frequency usable

src/esp32/pwm.rs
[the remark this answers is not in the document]
acmorrow: No, I definitely don't think it warrants force inlining. It just didn't seem to need to be its own function. I don't mind at all if it stays one though.

src/esp32/pwm.rs
npmenard: The logic here is to temporally decrement the timer count so we can find out if other channels are bind to it. 
So we have 3 Options:
1) One channel is bind so we update the frequency and the count remain unchanged
2) Another Timer is allocated and we decrement the old timer and increment the new one
3) No other timer can be allocated the channel remain bound to the previous timer and nothing happens

acmorrow: Thanks, with those cases listed out I see how it works. A comment summarizing what you've said here might be worthwhile for future us.

acmorrow (approved):

PR #134, opened 2023-12-19
fix Makefile build commands

gvaradarajan:


gvaradarajan:
Never mind, I didn't espflash locally. Oops

PR #136, opened 2024-01-05
RSDK-5903 Add micro-rdk support for HC-SR04 type ultrasonic sensors

acmorrow:


acmorrow:
This is working well enough that I thought it was worth getting a review started. There are a few small TODOs left, and a few other things I'll leave comments about. For now, this doesn't use the mcpwm subsystem, it just manually times the pulse based on digital interrupts. I'd suggest that if we want to move to a mcpwm based approach that we land this implementation first in order to unblock using this sensor with the micro-rdk.

src/esp32/hcsr04.rs
```diff
@@ -0,0 +1,289 @@
+// Support for HC-SR04 style ultrasonic ranging modules. See
+// https://cdn.sparkfun.com/datasheets/Sensors/Proximity/HCSR04.pdf.
```
acmorrow, line 2: I will add comments here covering the configuration parameters with an example config for the sensor.

src/esp32/hcsr04.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+}
+
+struct IsrSharedState {
+    // The state machine used to track interrupts and compute the
+    // length of the echo pulse. It holds one of the following values:
+    //
+    // 0: Starting state, ready to take a reading
+    // i64 > 0: Millisecond timestamp of first edge of echo signal
+    // -1 : Lockout state, further interrupts will be ignored
```
acmorrow, line 54: I'm not 100% convinced this state is needed, and managing it costs an extra compare_and_exchange in the isr. I added it while debugging state handling around timeouts, but there were other, subsequently resolved, issues that may have been the actual cause. I will re-test with it removed.

src/esp32/hcsr04.rs
```diff
@@ -0,0 +1,289 @@
...
+    timestamp: AtomicI64,
+
+    // The channel the ISR will use to communicate results back to waiters.
+    notifier: Arc<Notifier>,
+}
+
+#[derive(DoCommand)]
+pub struct HCSR04Sensor {
+    // We require a board in order to ensure that ISR setup has completed
```
acmorrow, line 63: This is related to the TODO down on 131. I'd like to move the ISR setup out of this file, remove what is currently being done in `pin.rs`, and start using the now `pub` interfaces from HAL to make ISR setup something that a `board` handles automatically. I'm definitely open to other approaches.

src/esp32/hcsr04.rs
```diff
@@ -0,0 +1,289 @@
...
+    _board: BoardType,
+
+    // The PinDriver to control the pin that triggers issuing a pulse.
+    trigger_pin: PinDriver<'static, AnyIOPin, Output>,
+
+    // The PinDriver used to listen for digital interrupts and measure
+    // the length of the echo pulse.
+    echo_interrupt_pin: PinDriver<'static, AnyIOPin, Input>,
```
acmorrow, line 71: This does not seem to work if this `PinDriver` is  made `InputOutput`, so I can't use the `Esp32GPIOPin` stuff, since that unconditionally makes all pins `InputOutput`.

src/esp32/hcsr04.rs
```diff
@@ -0,0 +1,289 @@
...
+        {
+            Ok(_) => {
+                // Initial edge: timestamp gets stored.
+            }
+            Err(-1) => {
+                // Locked out: ignore this interrupt. Sometimes it
+                // seems that there are spurious edges on timeouts.
+            }
+            Err(prior) => {
+                // Final edge: try set the lockout to debounce
```
acmorrow, line 192: Typo here, but I'm hoping that the 'lockout' state goes away so maybe this comment does too.

src/esp32/hcsr04.rs
```diff
@@ -0,0 +1,289 @@
...
+        enable_isr_service()?;
+
+        let notification = Notification::new();
+        let notifier = notification.notifier();
+
+        let mut sensor = Self {
+            _board: board,
+            trigger_pin: PinDriver::output(unsafe { AnyIOPin::new(trigger_pin) })?,
+            echo_interrupt_pin: PinDriver::input(unsafe { AnyIOPin::new(echo_interrupt_pin) })?,
+            timeout: timeout.unwrap_or(Duration::from_millis(50)),
```
acmorrow, line 142: The RDK defaults `timeout_ms` to 1000ms which is far too long. Also, we are blocking in `wait` for up to this duration, so it really doesn't make sense. The range of the sensor is 4m so 50ms is more than enough. I think I might consider clipping any configured value higher than, say, 100ms.

src/esp32/hcsr04.rs
```diff
@@ -0,0 +1,289 @@
...
+                Some(Self::subscription_interrupt),
+                Arc::as_ptr(&sensor.isr_shared_state) as *mut _,
+            ))?;
+        }
+
+        Ok(sensor)
+    }
+
+    #[inline(always)]
+    #[link_section = ".iram1.intr_srv"]
```
acmorrow, line 176: I copied this verbatim from how `pin.rs` is set up, but I'm not 100% sure whether it is required or not.

src/common/sensor.rs
```diff
@@ -35,7 +35,7 @@ pub trait Sensor: Readings + Status + DoCommand {}
 pub type SensorType = Arc<Mutex<dyn Sensor>>;
 
 pub trait SensorT<T>: Sensor {
-    fn get_readings(&self) -> anyhow::Result<TypedReadingsResult<T>>;
+    fn get_readings(&mut self) -> anyhow::Result<TypedReadingsResult<T>>;
```
npmenard, line 38: I see why we need this for ultrasonic sensor but I would like to avoid changing the signature to a &mut self if possible. Would we be able to get around this limitation with interior mutability?

src/esp32/hcsr04.rs
```diff
@@ -0,0 +1,289 @@
...
+        defer! {
+            let _ = trigger_pin.set_high();
+        }
+
+        // Wait (up to timeout) for a notification from the
+        // ISR. Convert any result from the notification into a
+        // distance.
+        match self
+            .interrupt_notification
+            .wait(TickType::from(self.timeout).as_millis_u32())
```
npmenard, line 267: this call will suspend the executor thread meaning  no future will be able to advance during `self.timeout` how hard would it be to get a local copy of the executor and use a blocking call instead?

src/esp32/hcsr04.rs
```diff
@@ -0,0 +1,289 @@
...
+    }
+
+    #[inline(always)]
+    #[link_section = ".iram1.intr_srv"]
+    unsafe extern "C" fn subscription_interrupt(arg: *mut core::ffi::c_void) {
+        let arg: &mut IsrSharedState = &mut *(arg as *mut _);
+        let when = esp_idf_sys::esp_timer_get_time();
+        match arg
+            .timestamp
+            .compare_exchange(0, when, Ordering::AcqRel, Ordering::Acquire)
```
npmenard, line 182: I am not sure you need to atomically store the timestamp result with the way the code is written since it can only be updated from the ISR and read by the thread after it was notified by the ISR.

src/esp32/hcsr04.rs
npmenard: this is just the section where to code is going , the important part is that it is going in  the iram (maybe not needed for low-frequency interrupt)

npmenard (changes requested):
Done with first pass

src/esp32/hcsr04.rs
gvaradarajan: ~can we use a different executor entirely just to do this wait? What are the consequences if you just instantiated an instance of `smol::Executor`?~ Never mind, this was dumb.

src/esp32/hcsr04.rs
gvaradarajan: Do we need this since the board is always configured first by default? And even if it wasn't, wouldn't it be fine just to define and register a dependency getter that requires the board?

src/esp32/hcsr04.rs
gvaradarajan: I had a different question here: why doesn't the atomic nature of the timestamp cover the case that the lockout state is trying to guard against? (I know you were considering removing lockout anyway, so if you do please disregard)

gvaradarajan: also @npmenard , doesn't `get_readings` try to access it outside of an ISR context (L. 251)? Or will that still not interfere?

gvaradarajan (changes requested):

src/common/sensor.rs
acmorrow: Sure, I'll give interior mutability a try. I wasn't exactly thrilled about needing to switch it to `&mut self` either.

Note though that `Readings::get_generic_readings` is `&mut self` too as of #126. If it is a design goal for the sensor API to avoid mutability for readings, perhaps that needs to be revisited?

src/esp32/hcsr04.rs
acmorrow: Yes, I'm clear on the link_section part; it was more the second part (whether it should be or is required to be in iram or not) that I was (and remain) unsure about.

src/esp32/hcsr04.rs
acmorrow: I brought this up in the micro-rdk team meeting a bit ago and we decided then that a blocking implementation was OK, at least initially. Of course we can change our minds about that and I'll see what I can do. However, I don't see an easy way to obtain access to the executor. Would you be OK merging with the blocking approach for now, and then filing a follow-up ticket to investigate how we can do it without blocking the executor? I know @gvaradarajan is also doing work in this space right now, so I'm interested to see what his investigation yields as well.

src/esp32/hcsr04.rs
acmorrow: @npmenard - My reasoning in favor of an atomic here is that the timestamp gets written by the first interrupt (rising edge of echo pin) and the value written then needs to be read by the second interrupt (falling edge of echo pin) in order to compute the delta between them. As far as I understand there needs to be a happens-before edge between the write and the read, and without the atomic there would not be. It also isn't clear to me that the notification enforces any global ordering such that a non-atomic write in the ISR would be seen by the read in `get_readings`, ditto the write of zero to "arm" the interrupt being seen in the ISR, as noted by @gvaradarajan.

@gvaradarajan - I'm about 90% sure that the lockout isn't needed, I just didn't want to hold up the review  until I've re-tested, since if it isn't actually needed its just a simplification. What I was observing was what looked like spurious edges when the sensor timed out. But things also weren't fully debugged at that point and it is entirely possible that those "spurious edges" were in fact my code misbehaving. Some testing should quickly make it clear.

src/esp32/hcsr04.rs
acmorrow: > Do we need this since the board is always configured first by default?

If that's the case, then no, we probably don't need it.


> And even if it wasn't, wouldn't it be fine just to define and register a dependency getter that requires the board?

I'm not entirely sure what you mean by "register a dependency getter"? Do you mean just doing `registry::get_board_from_dependencies` and then not retaining the result? I'd be fine with that.

Overall, I'd also just be fine with taking it out for now, though I expect it will creep back in later.

src/common/sensor.rs
acmorrow: Updated with interior mutability by wrapping the `PinDriver`s with `RefCell` and sprinkling `borrow_mut` as needed. I'm not sure if this is better or not. The sensor, for better or worse, really is changing the state of the outside world by driving pins and configuring interrupts. Requiring it to remain `&self` means any sort of stateful sensor will have little choice but to similarly adopt interior mutability. On the other hand, maybe the majority of sensors are passive. I think it comes down to what we want in terms of API design and implementation experience for module authors and how that intersects with idiomatic rust API design, neither of which is something I'm particularly expert in.

src/esp32/hcsr04.rs
acmorrow: I removed the lockout state, since after re-testing with it removed I no longer find it to be necessary, so this code is gone.

src/esp32/hcsr04.rs
```diff
@@ -0,0 +1,289 @@
...
+            }
+            Err(-1) => {
+                // Locked out: ignore this interrupt. Sometimes it
+                // seems that there are spurious edges on timeouts.
+            }
+            Err(prior) => {
+                // Final edge: try set the lockout to debounce
+                // (i.e. we observe `timestamp` to be still set to
+                // `prior`), and if so, return a reply.
+                if let Ok(_) =
```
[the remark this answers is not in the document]
acmorrow: Resolved as above.

src/esp32/hcsr04.rs
acmorrow: This is now clamped between 100us and 100ms since those are below and above the documented min and max echo response pulse times.

src/esp32/hcsr04.rs
acmorrow: The lockout has been removed, further testing indicated it was not required.

acmorrow:
Meanwhile, I'm getting a clippy error I don't totally understand:

```
error: unused `#[macro_use]` import
  --> src/lib.rs:12:13
   |
12 | #[macro_use(defer)]
   |             ^^^^^
   |
   = note: `-D unused-imports` implied by `-D warnings`

error: could not compile `micro-rdk` (lib) due to previous error
make: *** [Makefile:62: clippy-native] Error 101
Error: Process completed with exit code 2.
```

I really am using the `defer` macro though, so I'm not sure why it is complaining. However, if I try to move the
```
#[macro_use(defer)]
extern crate scopeguard;
```
block into `hcsr04.rs` it fails to compile:
```
error: could not compile `micro-rdk` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
error[E0468]: an `extern crate` loading macros must be at the crate root
  --> /host/micro-rdk/src/esp32/hcsr04.rs:44:1
   |
44 | extern crate scopeguard;
   | ^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0468`.
make: *** [Makefile:11: build-esp32-bin] Error 101
```

So, I'm not sure where I should be doing this (or doing something else).

src/esp32/hcsr04.rs
gvaradarajan: to clarify, I meant using [register_dependency_getter](https://github.com/viamrobotics/micro-rdk/blob/6c3b3a5c175fe21ff2c42f258ded231305b7d1ce/src/common/registry.rs#L287)

src/esp32/hcsr04.rs
```diff
@@ -0,0 +1,331 @@
...
+//     "timeout_ms" : "20",
+//   },
+//   "depends_on": [
+//     "board"
+//   ]
+// }
+//
+// Configuration details:
+//
+// The sensor requires a board dependency in the `depends_on`
```
gvaradarajan, line 22: I don't think we actually use the depends_on in micro-RDK. All of our dependencies are implicit and done by registering a [dependency getter](https://github.com/viamrobotics/micro-rdk/blob/6c3b3a5c175fe21ff2c42f258ded231305b7d1ce/src/common/registry.rs#L287)

acmorrow:
> Meanwhile, I'm getting a clippy error I don't totally understand:
> 
> ```
> error: unused `#[macro_use]` import
>   --> src/lib.rs:12:13
>    |
> 12 | #[macro_use(defer)]
>    |             ^^^^^
>    |
>    = note: `-D unused-imports` implied by `-D warnings`
> 
> error: could not compile `micro-rdk` (lib) due to previous error
> make: *** [Makefile:62: clippy-native] Error 101
> Error: Process completed with exit code 2.
> ```
> 
> I really am using the `defer` macro though, so I'm not sure why it is complaining. However, if I try to move the
> 
> ```
> #[macro_use(defer)]
> extern crate scopeguard;
> ```
> 
> block into `hcsr04.rs` it fails to compile:
> 
> ```
> error: could not compile `micro-rdk` (lib) due to 2 previous errors
> warning: build failed, waiting for other jobs to finish...
> error[E0468]: an `extern crate` loading macros must be at the crate root
>   --> /host/micro-rdk/src/esp32/hcsr04.rs:44:1
>    |
> 44 | extern crate scopeguard;
>    | ^^^^^^^^^^^^^^^^^^^^^^^^
> 
> error: aborting due to previous error
> 
> For more information about this error, try `rustc --explain E0468`.
> make: *** [Makefile:11: build-esp32-bin] Error 101
> ```
> 
> So, I'm not sure where I should be doing this (or doing something else).

@gvaradarajan pointed out to me on slack that I hadn't guarded the `macro(use)` with `#[cfg(feature = "esp32")]`, so that's fixed now.

gvaradarajan (approved):
Just wanted a minor change to the doc comment, but pending the executor conversation LGTM

src/common/sensor.rs
npmenard: I not an expert either with idiomatic Rust API design but I feels more natural to have get_reading be non mutable and have the sensor update it's state if need in a separate thread/future or worst case via interior mutability (in that case concurrent reading should yield a previous value) With respect to  `Readings::get_generic_readings` i forgot why we ended up with mut self here since it delegate all call to `get_readings`

src/esp32/hcsr04.rs
npmenard: Sorry I forgot in that case we should have a max for timeout_ms so someone cannot set a very large value accidentally and break things

src/esp32/hcsr04.rs
npmenard: I see, I was reasoning that since interrupts cannot be pre-empted then all writes would flush when exiting the interrupt context (since all registers would be swapped) and since measuring the with ultrasonic has strongly ordered steps then  you couldn't get where you read a variable  in one thread when an interrupt comes in. However I couldn't find anything backing the first assumption in the technical reference for Xtensa and Esp32 plus I missed that the variable is i64 so no atomic read/write by default so I think it is better to leave the implementation  as is

src/esp32/hcsr04.rs
npmenard: i don't think  it is since it is not time critical or will be often called

npmenard (approved):
LGTM

acmorrow:
I filed two new tickets and added TODO's referencing them:
- https://viam.atlassian.net/browse/RSDK-6278, for not blocking the executor
- https://viam.atlassian.net/browse/RSDK-6279, for refactoring Esp32GPIOPin such that the ultrasonic sensor can use it instead of using `PinDriver` directly.

PR #138, opened 2024-01-11
RSDK-5973 - ESP32 connection canary

gvaradarajan:
Testing the github actions on my own a little later before merging, but wanted to get input on the connectivity data published and the acceptability threshold in the meantime

Used the Viam Python SDK since the golang one has its own netcode that does not make use of rust-utils.

Currently implemented to get the following information every hour and save to our MongoDB instance as a database called "esp32_canary" under a collection called "hourly_results"

- could we connect to the robot?
- how long did it take to connect to the robot?
- how many attempts did it take to connect to the robot?
- could we get/set a GPIO pin on the ESP32?
- if there was a failure, we store the error as well

Additionally there is a daily summary compiled at 11:30PM UTC that records the following information and saved to the same database under a collection called "daily_summaries"

- how many successful runs?
- how many failures to connect to the robot initially?
- how many failures when making calls to the board API?
- how many attempts did it take to connect to the robot on average (when successful)?
- what was the average time it took to connect to the robot?

### DIFF
--- a/.github/workflows/canary_summary.yml
+++ b/.github/workflows/canary_summary.yml
@@ -1,0 +1,29 @@
+name: Daily ESP32 Connection Summary
+
+on:
+  workflow_call:
+    secrets:
+      MONGODB_TEST_OUTPUT_URI:
+        required: true
+  schedule:
+    - cron:  '30 23 * * *'
+
+env:
+  MONGODB_TEST_OUTPUT_URI: ${{ secrets.MONGODB_TEST_OUTPUT_URI }}
+
+jobs:
+  canary:
+    runs-on: x64
+    defaults: 
+      run:
+        working-directory: ./canary
+        shell: bash
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+      - name: Run Canary
+        run: |
+          python -m pip install -r requirements.txt
+          python canary_summary.py
--- a/canary/canary_test.py
+++ b/canary/canary_test.py
@@ -21,17 +21,19 @@ async def connect(robot_address: str, api_key: str, api_key_id: str) -> Coroutin
 async def main():
     mongo_connection_str = os.environ["MONGODB_TEST_OUTPUT_URI"]
     db_client = MongoClient(mongo_connection_str)
-    db = db_client["esp32_canary"]
-    coll = db["hourly_results"]
+    db = db_client["micrordk_canary"]
+    coll = db["raw_results"]
 
     timestamp = datetime.datetime.now()
 
     default_item = {
         "_id": timestamp,
         "connection_success": False,
-        "board_api_success": False,
-        "error": "",
-        "connection_latency_ms": 0
+        "connection_error": "",
+        "connection_latency_ms": 0,
+        "board_api_successes": 0,
+        "board_api_failures": 0,
+        "connection_attempts": 5
     }
 
     robot_address = os.environ["ESP32_CANARY_ROBOT"]
@@ -41,7 +43,6 @@ async def main():
     print(f"connecting to robot at {robot_address} ...")
 
     start = time.time()
-    connection_attempts = 5
     for i in range(5):
         try:
             robot = await connect(robot_address, api_key, api_key_id)
@@ -49,7 +50,7 @@ async def main():
             break
         except Exception as e:
             if i == 4:
-                default_item["error"] = str(e)
+                default_item["connection_error"] = str(e)
                 coll.insert_one(default_item)
                 raise e
             print(e)
@@ -60,19 +61,25 @@ async def main():
     default_item["connection_latency_ms"] = connectivity_time
     default_item["connection_attempts"] = connection_attempts
 
-    try:
-        board = Board.from_robot(robot, "board")
-        board_return_value = await board.gpio_pin_by_name("32")
-        _ = await board_return_value.get()
-        await board_return_value.set(True)
-        value = await board_return_value.get()
-        if not value:
-            raise ValueError("Pin not set to high successfully")
-        default_item["board_api_success"] = True
-    except Exception as e:
-        default_item["error"] = str(e)
-        coll.insert_one(default_item)
-        raise e
+    board_api_successes = 0
+    board_api_failures = 0
+
+    board = Board.from_robot(robot, "board")
+    board_return_value = await board.gpio_pin_by_name("32")
+    for _ in range(20):
+        try:
+            _ = await board_return_value.get()
+            await board_return_value.set(True)
+            value = await board_return_value.get()
+            if not value:
+                raise ValueError("Pin not set to high successfully")
+            board_api_successes += 1
+        except Exception as e:
+            board_api_failures += 1
+            default_item["connection_error"] = str(e)
+
+    default_item["board_api_successes"] = board_api_successes
+    default_item["board_api_failures"] = board_api_failures
     
     coll.insert_one(default_item)
 
--- a/canary/canary_test.py
+++ b/canary/canary_test.py
@@ -1,5 +1,6 @@
 import asyncio
 import datetime
+import numpy as np
 import os
 import time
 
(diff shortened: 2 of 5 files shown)


npmenard (commented):
LGTM but with a couple of questions,

Is there a way to capture which version of micro-rdk was being used for the test? When we release a new version do you have to manually update it? Is there a script that can't reset the esp32 if the test fail? (Like if we have a test failure reset the esp32 and redo the test to see if this time it passes?)

.github/workflows/canary_summary.yml
```diff
@@ -0,0 +1,29 @@
...
+        working-directory: ./canary
+        shell: bash
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 2
+      - name: Run Canary
+        run: |
+          pip install -r requirements.txt
```
acmorrow, line 28: Ever so slight preference for `python -m pip install ...` here.

canary/canary_test.py
```diff
@@ -0,0 +1,75 @@
...
+import os
+import time
+
+from pymongo import MongoClient
+from typing import Coroutine, Any
+from viam.robot.client import RobotClient
+from viam.components.board import Board
+
+async def connect(robot_address: str, api_key: str, api_key_id: str) -> Coroutine[Any, Any, RobotClient]:
+    opts = RobotClient.Options.with_api_key(
```
acmorrow, line 12: There were some recommended connection settings for python clients identified in https://viam.atlassian.net/browse/RSDK-6144. Not sure if those are 1) still relevant after @npmenard's recent improvements, or 2) something we want to do here if we are trying to measure "default" quality.

canary/canary_test.py
```diff
@@ -0,0 +1,75 @@
...
+            if i == 4:
+                default_item["error"] = str(e)
+                coll.insert_one(default_item)
+                raise e
+            print(e)
+        time.sleep(0.5)
+
+    connectivity_time = (time.time() - start) * 1000
+    default_item["connection_success"] = True
+    default_item["connection_latency_ms"] = connectivity_time
```
acmorrow, line 56: This will end up reporting the total connection time over multiple connection attempts, if such is necessary. Nothing wrong with that, but it could be misleading / noisier than it needs to be. Maybe report both? The total time required to connect across retries, and the latency of the succeeding connection?

canary/canary_test.py
```diff
@@ -0,0 +1,75 @@
...
+
+    try:
+        board = Board.from_robot(robot, "board")
+        board_return_value = await board.gpio_pin_by_name("32")
+        _ = await board_return_value.get()
+        await board_return_value.set(True)
+        _ = await board_return_value.get()
```
acmorrow, line 63: Consider checking that the return here is still `True`?

canary/canary_test.py
```diff
@@ -0,0 +1,75 @@
...
+        default_item["error"] = str(e)
+        coll.insert_one(default_item)
+        raise e
+    
+    coll.insert_one(default_item)
+
+    await robot.close()
+
+if __name__ == '__main__':
+    asyncio.run(main())
```
acmorrow, line 75: Can you add some closing newlines for this file and others that lack it?

acmorrow (approved):
LGTM with some small and optional suggestions

gvaradarajan:
> LGTM but with a couple of questions,
> 
> Is there a way to capture which version of micro-rdk was being used for the test? When we release a new version do you have to manually update it? Is there a script that can't reset the esp32 if the test fail? (Like if we have a test failure reset the esp32 and redo the test to see if this time it passes?)

So I originally had code that stored the micro-rdk version as well, but currently there is nothing in the robot API for either rdk or micro-rdk to report version because we didn't want people to write application code that depended on version number. This leaves only two options for retrieving it (neither of which seemed very good): 

- I could read it from Cargo.toml (but this wouldn't necessarily represent the version actually running on the canary)
- I could get it from the robot config returned by app (another connection to app aside from the robot feels like another thing that could fail to connect that isn't necessarily relevant to whether we could connect to the canary)

I was going to have the ESP32 canary connected to a RaspberryPi as a followup and then have a process on the pi that re-flashes when a new version is released. 

I don't think we want to reset and see if it passes, because it might obscure a real problem that happens to be fixed on reset.

canary/canary_test.py
gvaradarajan: Does it functionally matter what the latency of the last attempt is versus the overall time to someone interacting via the SDK who has to try multiple times anyway? Would saving the number of attempts instead be useful?

acmorrow: Functionally, no, but I think we might be interested to know if we suddenly went from a situation where it mostly connects on first attempt and takes 100ms, to a situation where it typically takes 4 failing tries at 10ms each, and then connects taking 60ms. As things stand we wouldn't see that transition because the average wouldn't change. Reporting the number of attempts required to connect seems like a useful way to disambiguate those cases, and if that's easier to implement then sure, I'm happy with that approach too. I also don't feel strongly about it, so I'm fine moving forward with just the average if you prefer.

.github/workflows/canary_summary.yml
acmorrow: Ping?

acmorrow (approved):
LGTM

canary/canary_test.py
stevebriskin: I agree with @acmorrow, this metric can be clearer -- connection_latency_ms - latency of the successful connection, if available.

acmorrow (approved):
LGTM!

canary/canary_test.py
```diff
@@ -0,0 +1,82 @@
...
+        attempt_reconnect_interval=0,
+        disable_sessions=True,
+        dial_options=DialOptions.with_api_key(api_key_id=api_key_id,api_key=api_key)
+    )
+    return await RobotClient.at_address(robot_address, opts)
+
+async def main():
+    mongo_connection_str = os.environ["MONGODB_TEST_OUTPUT_URI"]
+    db_client = MongoClient(mongo_connection_str)
+    db = db_client["esp32_canary"]
```
stevebriskin, line 24: NIT: micrordk_canary

canary/canary_test.py
```diff
@@ -0,0 +1,82 @@
...
+    coll = db["hourly_results"]
```
stevebriskin, line 25: NIT: something raw_results will be better if we decide to run it more or less frequently

canary/daily_summary.py
```diff
@@ -0,0 +1,65 @@
...
+    
+    avg_connection_latency_ms = latency_sum / successes
+    avg_connection_attempts = connection_attempts / successes
+
+    total_runs = successes + connection_failures + board_api_failures
+    if total_runs == 0:
+        raise Exception(f"no hourly canary results found for {today}")
+    
+
+    coll2 = db["daily_summaries"]
```
stevebriskin, line 48: Personally, I don't see much value in persisting these aggregated stats in a table since they can all be easily derived by aggregating raw data as-needed. The downside with persisting is that it's one more place to change when we add fields to the hourly docs. I don't feel strongly about this at all.

canary/canary_test.py
```diff
@@ -0,0 +1,82 @@
...
+                raise e
+            print(e)
+        time.sleep(0.5)
+
+    connectivity_time = (time.time() - start) * 1000
+    default_item["connection_success"] = True
+    default_item["connection_latency_ms"] = connectivity_time
+    default_item["connection_attempts"] = connection_attempts
+
+    try:
```
stevebriskin, line 63: 1. capturing the latency of this call would be useful
2. it'd be cool if we did n (e.g. 10-100) of these calls and recorded # of successes, failures, and mean and stddev latency times.

stevebriskin (approved):
minor comments. lgtm

canary/canary_test.py
gvaradarajan: with regards to 2, I figured this is what we were basically doing by running this every hour?

gvaradarajan: with regards to 1, do you mean for each call (getting the board, getting pin level, setting pin level)

stevebriskin: What I'm suggesting is based on that assumption that a typical client will connect once and issue many calls. Some calls will be latency-sensitive, some will not. One early use case we targeted was sensor readings for a control loop. That's what I'm trying to imitate.

1. call = pin level. in my mind you'd get board once, and do many pin calls. that should how people will typically interact with it.
2. i think comparing means/stddev over separate runs will be apples and oranges since each connection can take a different route.

npmenard (approved):

PR #139, opened 2024-01-12
RSDK-6310 Obtain hal and sys via esp-idf-svc

acmorrow:
Pretty much a purely mechanical change, but it sets us up to:
1) Potentially reduce the dependencies in various `Cargo.toml` files to just `esp-idf-svc`, if we want to, and
2) Re-export `esp-idf-svc` (or `esp-idf-[svc,hal,sys]` if we don't consolidate on `esp-idf-svc`) from Micro-rdk so module or project authors don't need to explicitly import them. We can consider doing the same for the `embedded-[svc,hal]` packages too.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ repository = "https://github.com/viamrobotics/micro-rdk"
 resolver = "2"
 version = "0.1.6"
 rust-version = "1.71"
+links = "micro_rdk"
 
 [lib]
 crate-type = ["lib"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ async-io = {git = "https://github.com/viamrobotics/async-io.git"}
 [features]
 default = []
 camera = []
-esp32 = ["dep:esp-idf-hal", "dep:esp-idf-svc", "dep:esp-idf-sys", "dep:embedded-svc", "dep:embedded-hal"]
+esp32 = ["dep:esp-idf-svc", "dep:embedded-svc", "dep:embedded-hal"]
 native = ["dep:rustls", "dep:webpki-roots", "dep:rustls-pemfile", "dep:mdns-sd", "dep:local-ip-address", "dep:openssl", "dep:rcgen", "dep:async-std-openssl"]
 
 [dev-dependencies]
@@ -36,9 +36,7 @@ bytes = "1.2.1"
 either = "1.8.0"
 embedded-hal = { version = "0.2.7", features = ["unproven"], optional = true }
 embedded-svc = { version = "0.26.4", optional = true }
-esp-idf-hal = { version = "0.42.5", optional = true }
 esp-idf-svc = { version = "0.47.3",  optional = true }
-esp-idf-sys = { version = "0.33.7", features = ["binstart"],  optional = true }
 futures = "0.3.28"
 futures-lite = "1.12.0"
 h2 = "0.3.22"
@@ -81,7 +79,7 @@ scopeguard = "1.2.0"
 [build-dependencies]
 anyhow = "1"
 const-gen = "1.3.0"
-embuild = "0.29"
+embuild = "0.31.3"
 
 [template]
 sub_templates = ["templates/project", "templates/module"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ bytes = "1.2.1"
 either = "1.8.0"
 embedded-hal = { version = "0.2.7", features = ["unproven"], optional = true }
 embedded-svc = { version = "0.26.4", optional = true }
-esp-idf-svc = { version = "0.47.3",  optional = true }
+esp-idf-svc = { version = "= 0.47.3",  optional = true }
 futures = "0.3.28"
 futures-lite = "1.12.0"
 h2 = "0.3.22"
--- a/build.rs
+++ b/build.rs
@@ -1,9 +1,6 @@
 fn main() -> anyhow::Result<()> {
     if std::env::var("TARGET").unwrap() == "xtensa-esp32-espidf" {
-        embuild::build::CfgArgs::output_propagated("ESP_IDF")?;
-        embuild::build::LinkArgs::output_propagated("ESP_IDF")?;
-        let cfg = embuild::build::CfgArgs::try_from_env("ESP_IDF")?;
-        cfg.output();
+        embuild::espidf::sysenv::output();
     }
 
     Ok(())
--- a/build.rs
+++ b/build.rs
@@ -1,6 +1,12 @@
 fn main() -> anyhow::Result<()> {
     if std::env::var("TARGET").unwrap() == "xtensa-esp32-espidf" {
-        embuild::espidf::sysenv::output();
+        let cfg_args = embuild::build::CfgArgs::try_from_env("ESP_IDF_SVC")?;
+        cfg_args.output();
+        cfg_args.propagate();
+
+        let link_args = embuild::build::LinkArgs::try_from_env("ESP_IDF_SVC")?;
+        link_args.output();
+        link_args.propagate();
     }
 
     Ok(())
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -37,7 +37,6 @@ anyhow = {version = "1", features = ["backtrace"]}
 embedded-svc = "0.26.4"
 embedded-hal = { version = "0.2.7", features = ["unproven"]}
 micro-rdk = { path = "../", default-features = false, features = ["esp32"] }
-esp-idf-svc = { version = "= 0.47.3", features = ["experimental"] }
 log = "0.4"
 async-channel = "1.8.0"
 smol = "1.2"
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -36,8 +36,6 @@ anyhow = { version = "1", features = ["backtrace"] }
 anyhow = {version = "1", features = ["backtrace"]}
 embedded-svc = "0.26.4"
 embedded-hal = { version = "0.2.7", features = ["unproven"]}
-esp-idf-svc = { version = "= 0.47.3" }
-
 micro-rdk = { path = "../", default-features = false, features = ["esp32"] }
 log = "0.4"
 async-channel = "1.8.0"
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -37,7 +37,7 @@ anyhow = {version = "1", features = ["backtrace"]}
 embedded-svc = "0.26.4"
 embedded-hal = { version = "0.2.7", features = ["unproven"]}
 micro-rdk = { path = "../", default-features = false, features = ["esp32"] }
-esp-idf-svc = { version = "0.47.3", features = ["experimental"] }
+esp-idf-svc = { version = "= 0.47.3", features = ["experimental"] }
 log = "0.4"
 async-channel = "1.8.0"
 smol = "1.2"
--- a/examples/Cargo.toml
+++ b/examples/Cargo.toml
@@ -33,13 +33,11 @@ anyhow = { version = "1", features = ["backtrace"] }
 
 
 [target.'cfg(target_os="espidf")'.dependencies]
-esp-idf-sys = { version = "0.33.7", features = ["binstart"] }
 anyhow = {version = "1", features = ["backtrace"]}
 embedded-svc = "0.26.4"
 embedded-hal = { version = "0.2.7", features = ["unproven"]}
 micro-rdk = { path = "../", default-features = false, features = ["esp32"] }
 esp-idf-svc = { version = "0.47.3", features = ["experimental"] }
-esp-idf-hal = "0.42.5"
 log = "0.4"
 async-channel = "1.8.0"
 smol = "1.2"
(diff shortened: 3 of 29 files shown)


acmorrow:
> Where do you re-export the packages? I couldn't find it :P

Haven't done it yet. This is just using the new names.

npmenard:
I see, I misread. In terms of versioning does that mean we can only pin one version of esp-idf-svc for all? Also i remember you mentioned you still had to include esp-idf-sys in the dependency for it to compile properly if this is the case do you think we should use it instead of `esp-idf-svc::sys`?

acmorrow:
> I see, I misread. In terms of versioning does that mean we can only pin one version of esp-idf-svc for all?

I think that if we did remove our explicit dependencies on `esp-idf-[sys,hal]` from the micro-rdk `Cargo.toml` then the answer is that yes, we would get whatever version of sys and hal `esp-idf-svc` brought in. What I don't yet know, and want to determine, is what would happen if we retained our explicit micro-rdk Cargo.toml dependency declarations. That might allow us to retain control, as long as we selected versions of `esp-idf-[sys,hal]` that were compatible with `esp-idf-svc`. I need to do a little more research into how dependency resolution works to know if that is actually true.

> Also i remember you mentioned you still had to include esp-idf-sys in the dependency for it to compile properly if this is the case do you think we should use it instead of `esp-idf-svc::sys`?

That's correct, and I haven't yet identified exactly why that is. It may be that, per your above comment, moving to the indirect use of `sys` and `hal` by way of `svc` has advanced us to a newer version of those packages such that our build no longer works in some subtle way. Or, it may be that having the explicit dependencies is necessary to surface some information into the build phase. I plan to investigate that as well.

In either case, I don't plan to merge this until I have complete answers to those questions. Potentially, those answers would also bring changes to `Cargo.toml` in an update to this review.

Overall though, do we like this approach of routing all ESP-IDF references via `svc` if we can resolve the above questions in a satisfactory way?

npmenard:
Yes i like this approach I think ideally I would re-export esp-idf-sys via svc just to keep things clean. Since esp-idf-sys only generates the bindings we shouldn't to keep them separate we would just need to pin esp-idf-svc to this exact version to prevent accidental upgrades. 
The issue you ran into is probably just related to the way we instruct embuild the generates the bindings (we set that up a while ago and haven't changed it since then while they did) I based the code on `https://github.com/ivmarkov/rust-esp32-std-demo` and quickly looking at it it appears they don't have an explicit dependency to esp-idf-sys in their Cargo.toml might be as easy and upgrading embuild to their version?

gvaradarajan (approved):
LGTM, I generally agree that this approach to the packages is an improvement over what we've been doing so far. The one thing I would want to test extensively is the behavior of the built-in drivers that make direct use of esp32 libraries (that would the board, encoder, and the hcsr04 ultrasonic sensor)

build.rs
```diff
@@ -1,9 +1,6 @@
 fn main() -> anyhow::Result<()> {
     if std::env::var("TARGET").unwrap() == "xtensa-esp32-espidf" {
-        embuild::build::CfgArgs::output_propagated("ESP_IDF")?;
-        embuild::build::LinkArgs::output_propagated("ESP_IDF")?;
-        let cfg = embuild::build::CfgArgs::try_from_env("ESP_IDF")?;
-        cfg.output();
+        embuild::espidf::sysenv::output();
```
acmorrow, line 3: Confidence in this change is ... uncertain. I'm taking this from what the esp-idf template does: https://github.com/esp-rs/esp-idf-template/blob/master/cargo/build.rs#L2.

See below for my comment on where things currently break. Perhaps more of what is done in https://github.com/ivmarkov/rust-esp32-std-demo/blob/main/build.rs is required?

examples/Cargo.toml
```diff
@@ -33,13 +33,10 @@ anyhow = { version = "1", features = ["backtrace"] }
 
 
 [target.'cfg(target_os="espidf")'.dependencies]
-esp-idf-sys = { version = "0.33.7", features = ["binstart"] }
 anyhow = {version = "1", features = ["backtrace"]}
 embedded-svc = "0.26.4"
 embedded-hal = { version = "0.2.7", features = ["unproven"]}
 micro-rdk = { path = "../", default-features = false, features = ["esp32"] }
-esp-idf-svc = { version = "0.47.3", features = ["experimental"] }
```
acmorrow, line 41: Note: this removal breaks building the examples, and I am not sure why. The error is:
```
  = note: [ldproxy] Running ldproxy
          thread 'main' panicked at /usr/local/cargo/registry/src/index.crates.io-6f17d22bba15001f/ldproxy-0.3.3/src/main.rs:44:13:
          Cannot locate argument '--ldproxy-linker <linker>'
```
I'm investigating.

examples/esp32-with-cred/esp32-server-with-cred.rs
```diff
@@ -1,10 +1,10 @@
 use log::*;
 use thiserror::Error;
 
-use esp_idf_svc::eventloop::EspSystemEventLoop;
-use esp_idf_svc::nvs::{EspDefaultNvs, EspDefaultNvsPartition, EspNvs};
-use esp_idf_sys::EspError;
-use esp_idf_sys::{g_wifi_feature_caps, CONFIG_FEATURE_CACHE_TX_BUF_BIT};
+use micro_rdk::esp_idf_svc::eventloop::EspSystemEventLoop;
```
acmorrow, line 4: Within examples, which don't really form part of micro-rdk, we obtain `esp-idf-svc` via the `micro-rdk` re-export.

src/esp32/analog.rs
```diff
@@ -1,8 +1,8 @@
 #![allow(dead_code)]
 use crate::common::analog::AnalogReader;
+use crate::esp_idf_svc::hal::adc::{AdcChannelDriver, AdcDriver};
```
acmorrow, line 3: Within the micro-rdk proper, we obtain `esp-idf-svc` and its re-exported `sys` and `hal` crates via `crate::`.

src/esp_idf_svc.rs
```diff
@@ -0,0 +1,2 @@
+#[cfg(feature = "esp32")]
```
acmorrow, line 1: Here is where we re-export `esp-idf-svc`. Note that the `feature = "esp32"` condition is duplicated between this file and the `pub mod` statement in `src/lib.rs`. I wasn't sure whether it was better to make the `pub mod` conditional, or this conditional, or both. I opted for both, but please let me know if you have a preference for removing one or the other.

acmorrow (commented):
@npmenard and @gvaradarajan - Updated, PTAL. Note that this isn't *quite* working. It does work if I revert https://github.com/viamrobotics/micro-rdk/pull/139/commits/537aaa97ad873e418eb5a5d6e1fd92165a8c80b1, and I'm investigating why. I'd like to solve that, because I'd prefer that projects using `micro-rdk` don't need to name any of `esp-idf-[svc,sys,hal]` directly in `Cargo.toml` and just obtain it from `micro-rdk`.

acmorrow:
Until we are happy with this (or decide against it) I've held off on the compensating changes that would be required in:
 - https://github.com/viamrobotics/micro-rdk-robot-template
 - https://github.com/viam-labs/micro-rdk-modular-driver-examples
 - https://github.com/viam-labs/micro-rdk-esp32-sensor-examples

Overall, they are quite mechanical so I don't expect it will be difficult.

Cargo.toml
```diff
@@ -36,9 +36,7 @@ bytes = "1.2.1"
 either = "1.8.0"
 embedded-hal = { version = "0.2.7", features = ["unproven"], optional = true }
 embedded-svc = { version = "0.26.4", optional = true }
-esp-idf-hal = { version = "0.42.5", optional = true }
-esp-idf-svc = { version = "0.47.3",  optional = true }
-esp-idf-sys = { version = "0.33.7", features = ["binstart"],  optional = true }
+esp-idf-svc = { version = "= 0.47.3",  optional = true }
```
acmorrow, line 39: Also also, note that `esp-idf-svc` is now explicitly pinned so we don't get unexpected upgrades.

src/esp_idf_svc.rs
npmenard: if you move the file to esp32 does that help? If so then we can keep them  under crate::esp32 namespace

build.rs
acmorrow: Using `embuild::espidf::sysenv::output()` with an updated `embuild` seems to necessary, so I'm keeping that change.

So far though, I haven't been able to remove the link error when I drop the `esp-idf-svc` dependency from `examples/Cargo.toml`. So, I'm going to restore that line for now so we can lock in the gains here, and then investigate how we can remove it once this is merged.

examples/Cargo.toml
acmorrow: Pre above, I'm going to back out this removal for now.

src/esp_idf_svc.rs
acmorrow: Sure, I can move it down a layer, though it will result in a somewhat unwieldy name like `crate::esp32::esp_idf_svc::[sys|hal]::`, or `micro_rdk::esp32::esp_idf_svc::[sys|hal]::`. If we wanted to remove the redundant `esp32::esp` I could probably re-export it as `idf_svc`? So it'd be `[crate|micro_rdk]::esp32::idf_svc::[sys|hal]::`. On the other hand, I feel like renaming the crate is iffy. Happy to do it any way, just let me know.

acmorrow:
@npmenard: updated, PTAL. This was rebased above @gvaradarajan's template migration, so the interdiff won't prove useful. Important changes since last round of review:

- I restored `esp-idf-svc` as a `[dependency]` in `examples/Cargo.toml`, and retained it in `templates/project/Cargo.toml` to work around the linker issue, which I will continue to pursue after this merges.

- The templates have been updated to use the new names.

- I removed the `import ...::sys as _` from several files, as it was creating an unused import warning. I'm really not sure what purpose it served originally, so if it was there for a good reason I can always go the other way and suppress the warning.

As far as I know, the only remaining question is whether to adjust the export of `esp_idf_svc` to happen under the `esp32` directory, and, if so, under what name.

src/esp32/utils.rs
```diff
@@ -6,7 +6,9 @@ macro_rules! esp32_print_heap_summary {
     () => {
         #[cfg(debug_assertions)]
         {
-            use esp_idf_sys::{heap_caps_get_free_size, heap_caps_get_total_size, MALLOC_CAP_8BIT};
+            use $crate::esp_idf_svc::sys::{
```
acmorrow, line 9: clippy requested this change

npmenard (approved):

Cargo.toml
```diff
@@ -8,6 +8,7 @@ repository = "https://github.com/viamrobotics/micro-rdk"
 resolver = "2"
 version = "0.1.6"
 rust-version = "1.71"
+links = "micro_rdk"
```
acmorrow, line 11: This needs to be here so that `embuild` will let us get use `MICRO_RDK` as an argument to `output_propagated`.

acmorrow (commented):
@npmenard - 

I've found a way to make it so that the examples (and projects/modules) do not need to declare a dependency on `esp-idf-svc`, and pushed that change here: https://github.com/viamrobotics/micro-rdk/pull/139/commits/946c3aded2a49075c0a5b1f71650cb4ea577ad00.

This seems to work, but I'll admit that adding the `links` part to the top-level `Cargo.toml` is troubling. If we don't like this, I will go back to requiring the examples and project template to declare an explicit dependency on `esp-idf-svc`.

gvaradarajan (approved):
LGTM, that links stuff is wild

Cargo.toml
npmenard, line 11: This argument is meant for linking native libraries at least according to the doc. Is that usage explain somewhere? feels weird to have to link against ourselves or am I missing something?

acmorrow: Yes, I should have said more in my comment above.

That is the documented purpose of this field per the cargo docs. But it appears that `embuild` uses this field somewhat differently, as a key to understand how and when to propagate toolchain settings. See for instance, the rather cryptic comment here: https://github.com/esp-rs/embuild/blob/master/src/build.rs#L392. It took me a while to figure out that they were talking about https://doc.rust-lang.org/cargo/reference/manifest.html#the-links-field, which is a little thin on details w.r.t. the effects of specifying it. What gives me some hope that this is not, in fact, as weird as it looks (or is at least no weirder than what is already going on in the esp-idf ecosystem), is that this is what `esp-idf-svc` itself does, per https://github.com/esp-rs/esp-idf-svc/blob/master/Cargo.toml#L13, in order to propagate its settings / args. As far as I can see there is no actual `esp_idf_svc` native lib in play: the `esp_idf_svc` project is entirely rust. Perhaps it would have been better if `embuild` had used custom package metadata to do this, like we did for micro-rdk modules: https://github.com/viam-labs/micro-rdk-esp32-sensor-examples/blob/main/Cargo.toml#L17-L18.

In any event, if you think this is a step to far (and I won't blame you in the least), we can back down to requiring that projects retain an explicit dependency on `esp-idf-svc` in their `[dependencies]` section.

acmorrow: Also, I'd like to add that all of the above is per my best current understanding of how these things work and what `embuild` is doing. It is entirely possible I've got it wrong or that there is another better way.

npmenard: I don't think we should do this it feels incorrect. We should be able to propagate the proper env with embuild.

npmenard (approved):

PR #140, opened 2024-01-17
RSDK-5805 - support remembering serial port for espflash

gvaradarajan:
Not yet tested, opening PR somewhat prematurely


npmenard (approved):
LGTM. I tested with and without the commit. With it asks me if I want to save the port without it doesn't

PR #141, opened 2024-01-17
Adding coredump

npmenard:
This PR introduce couple of changes to try to expose a bug where the esp32 stops responding to signaling request and remains in a waiting state forever.
Instead of relying a block_on to run the executor state machine we implement our own Driver (which just polls the executor run future). We don't use the Parker/Unparker mechanism to relinquish control to another thread but instead let the task sleep for a few tick before waking up again. This should avoid potential scenario where we park the thread without any means to wake it up.
I also added a timeout of 3 minutes on signaling request.

Lastly we setup a task watchdog on the task running the executor if the task doesn't clear the watchdog for 5 minutes then it will panic and produce a coredump for post-mortem analysis.


src/esp32/entry.rs
```diff
@@ -128,5 +139,59 @@ pub fn serve_web(
...
+    fn wake_by_ref(self: &Arc<Self>) {}
+}
+pub fn serve_web(
+    app_config: AppClientConfig,
+    tls_server_config: Esp32TlsServerConfig,
+    repr: RobotRepresentation,
+    _ip: Ipv4Addr,
+    webrtc_certificate: WebRtcCertificate,
+) {
+    // set the TWDT to expire after 5 minutes
```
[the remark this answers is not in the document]
gvaradarajan: resolved offline, these are different timeouts

gvaradarajan (approved):
LGTM

examples/sdkconfig.defaults
```diff
@@ -50,3 +50,12 @@ CONFIG_ESP32_WIFI_TASK_PINNED_TO_CORE_0=y
...
 CONFIG_MBEDTLS_DEBUG_LEVEL_WARN=n
+
+CONFIG_ESP_COREDUMP_ENABLE_TO_UART=y
+CONFIG_ESP_COREDUMP_DATA_FORMAT_ELF=y
+CONFIG_ESP_COREDUMP_CHECKSUM_CRC32=y
+CONFIG_ESP_COREDUMP_ENABLE=y
+CONFIG_ESP_COREDUMP_MAX_TASKS_NUM=64
+CONFIG_ESP_COREDUMP_UART_DELAY=0
+CONFIG_ESP_COREDUMP_DECODE_INFO=y
+CONFIG_ESP_COREDUMP_DECODE="disable"
```
acmorrow, line 61: Looking at the docs it looks like maybe `CONFIG_ESP_COREDUMP_DECODE_INFO` is supposed to be a value to be set for `CONFIG_ESP_COREDUMP_DECODE`?

Please disregard if I've got that wrong or if what you have done here is equivalent, but as written, I can't tell if you are intending to decode or not.

src/common/grpc_client.rs
```diff
@@ -162,21 +162,20 @@ pub struct GrpcClient<'a> {
 
 impl<'a> Drop for GrpcClient<'a> {
     fn drop(&mut self) {
-        if let Some(task) = self.http2_task.take() {
-            //(TODO(RSDK-3061)) avoid blocking on if possible
-            block_on(self.executor.run(async {
-                task.cancel().await;
-            }));
-        }
+        let _ = self.http2_task.take();
```
acmorrow, line 165: Why is it no longer necessary to wait on task cancelation here?

acmorrow (approved):
I think I generally like seeing more `async` functions and fewer explicit `block_on` calls, anyway. I, like @gvaradarajan, have some questions about the exact mechanics of how this works, but I think it is more important to get it merged in the hopes of finding the stall than it is to bring everyone up to complete understanding in the review.

examples/sdkconfig.defaults
npmenard: This is for esp-idf monitor to automatically decode base64 encoded coredump (we don't support that for now)

src/common/grpc_client.rs
npmenard: We don't need it really (because the TCP layer will close the connection gracefully) But better to add so we close the HTTP2 connection a bit more gracefully

PR #144, opened 2024-01-23
improvements

npmenard:
Just some minor changes to make the esp32 running micro-rdk more stable overtime


examples/sdkconfig.defaults
```diff
@@ -1,5 +1,5 @@
 # Rust often needs a bit of an extra main task stack size compared to C (the default is 3K)
-CONFIG_ESP_MAIN_TASK_STACK_SIZE=26000
+CONFIG_ESP_MAIN_TASK_STACK_SIZE=22528
```
npmenard, line 2: since 0.1.6 we use a bit less stack so lets claim it back :)

examples/sdkconfig.defaults
```diff
@@ -8,6 +8,7 @@ CONFIG_FREERTOS_HZ=1000
 CONFIG_ESP32S2_DEFAULT_CPU_FREQ_240=y
 CONFIG_ESP32S3_DEFAULT_CPU_FREQ_240=y
 CONFIG_ESP_SYSTEM_EVENT_TASK_STACK_SIZE=4096
+CONFIG_PTHREAD_TASK_STACK_SIZE_DEFAULT=4096
```
npmenard, line 11: A task is created to manage the timers we spawn, it would often time exceed it's stack limit and overwrite which would potentially lock the esp32 up.

examples/sdkconfig.defaults
```diff
@@ -31,6 +32,7 @@ CONFIG_SPIRAM_MALLOC_ALWAYSINTERNAL=4090
 
 #CONFIG_MBEDTLS_DEBUG=y
 CONFIG_MBEDTLS_SSL_PROTO_DTLS=y
+CONFIG_MBEDTLS_DEFAULT_MEM_ALLOC=y
```
npmenard, line 35: For security reason MBEDTLS by default only allocates in the internal ram. This option lets it allocate on the SPIRAM the security risk is that someone may reads its content viw the spi bus

src/esp32/entry.rs
```diff
@@ -131,7 +129,6 @@ pub async fn serve_web_inner(
             Box::new(
                 ViamServerBuilder::new(mdns, cloned_exec, client_connector, app_config)
                     .with_webrtc(webrtc)
-                    .with_http2(tls_listener, 12346)
```
npmenard, line 134: I removed HTTP2 listener since we are never using it

examples/sdkconfig.defaults
acmorrow: Usually, I'd say that if you have physical access to a device it is probably game over already, so fine. But, in this case, since the esp32 is maybe really out there in the world, physical access is expected. On the other hand, it looks like there are options to encrypt SPIRAM on some esp32 platforms (https://www.esp32.com/viewtopic.php?t=6513 and https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/api-guides/external-ram.html#encryption) and use eFuse to disable JTAG.

Let's say that wasn't done, and someone did pull TLS info out via SPI. What would that permit?

src/esp32/entry.rs
```diff
@@ -32,15 +34,15 @@ use futures_lite::Future;
...
+    _tls_server_config: Esp32TlsServerConfig,
     repr: RobotRepresentation,
     _ip: Ipv4Addr,
     webrtc_certificate: WebRtcCertificate,
     exec: Esp32Executor<'_>,
 ) {
     let (mut srv, robot) = {
         let mut client_connector = Esp32Tls::new_client();
-        let mdns = Esp32Mdns::new("".to_string()).unwrap();
+        let mdns = NoMdns {};
```
acmorrow, line 45: It looks like `Esp32Mdns` has no remaining usage after this change. Should it be removed?

src/common/conn/server.rs
```diff
@@ -381,11 +381,18 @@ where
...
                 .map_or(Err(ServerError::ServerConnectionTimeout), |r| r);
 
-            if connection.is_err() {
-                let _ = self.app_client.take();
-                continue;
-            }
-            if let Err(e) = match connection.unwrap() {
+            let connection = match connection {
+                Ok(c) => c,
+                Err(ServerError::ServerWebRTCError(_)) => {
```
acmorrow, line 386: I think a comment here is a good idea to explain both why `ServerWebRTCError` is special cased to not destroy the `app_client`, and why the `app_client` gets destroyed for other errors, as I think these error cases form an important part of the state machine.

src/esp32/entry.rs
```diff
@@ -95,10 +97,6 @@ pub async fn serve_web_inner(
             (cfg_response, robot)
         };
 
-        let address: SocketAddr = "0.0.0.0:12346".parse().unwrap();
-        let tls = Box::new(Esp32Tls::new_server(&tls_server_config));
-        let tls_listener = Esp32Listener::new(address.into(), Some(tls)).unwrap();
```
acmorrow, line 100: Is `Esp32Listener` now unused? Any other things that become unused if that thread is pulled?

acmorrow (approved):
LGTM; some small questions and a request for a comment, and want to think about the security implications a bit.

examples/sdkconfig.defaults
npmenard: they could get the client random and the session key thus decrypting all tls packets potentially for either the webrtc connection or the app connection. We could mitigate this with SPIRAM encryption but not all esp32 chip supports it

src/esp32/entry.rs
npmenard: no we want to keep it if folks want to use gRPC/HTTP2

src/esp32/entry.rs
npmenard: not really the code is shared with the client socket

acmorrow (approved):
LGTM

PR #147, opened 2024-01-29
RSDK-6421 Add support for sccache to micro-rdk canon images

acmorrow:
After this change, running `canon --profile ... env SCCACHE=1 ...` will enable `sccache` in the canon environment by setting `RUSTC_WRAPPER=/usr/bin/sccache` and `SCCACHE_DIR=/host/.sccache`, by analogy with the existing `.cargo-docker-registry` host-based cargo cache.

That makes it opt-in. It'd be nice, perhaps, if canon supported setting `env: ...` in your canon config file so you could opt-in to it persistently under a name, but it doesn't seem to. 

I'll confess to being very tempted to introduce as part of this change a new `.micro-rdk-canon-shared-state` or similarly named directory and push the `.cargo-docker-registry` and `.sccache` directories under it, without the leading `.`s. Let me know if you feel that's worthwhile, as it would be a pretty minimal change above what is here.

This currently comes with no controls over the cache size, but that can easily be added later if found to be needed.

### DIFF
--- a/etc/docker/Dockerfile
+++ b/etc/docker/Dockerfile
@@ -4,8 +4,8 @@ RUN apt update && apt-get install -y --no-install-recommends git libudev-dev mak
 RUN curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash
 RUN cargo binstall cargo-espflash --install-path /usr/ -y
 RUN cargo binstall espup --install-path /usr/ -y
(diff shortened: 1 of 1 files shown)


npmenard (approved):
LGTM, agree on a shared directory easier for a user when they want to clear build artifact (makes me think make clean doesn't clean canon artifacts !)

acmorrow:
@npmenard - Updated; cargo registry and `sccache` directory are now in `.micro-rdk-docker-caches/{cargo-registry,sccache}`, respectively.

PR #149, opened 2024-02-02
RSDK-6483 Workspace-ify micro-rdk

acmorrow:
This is a draft because I'd like to share what I have before investing more time. I've made it a draft because right now it doesn't quite work due to some issues (I believe) around `sdkconfig.defaults`, but I expect those are solvable. There are probably other rough edges to sand down as well before this is something I'd want to merge.

If we do like this direction, we can either stop here, or continue with splitting up `micro-rdk` into a few crates. I'm envisioning something like:
- `micro-rdk-api`: The code generated by `buf` from the `api` repo
- `micro-rdk-core`: Common type definitions and support code. Modules would build against this.
- `micro-rdk-resources` or `micro-rdk-builtins`: Implementations for builtins like the ultrasonic sensor, ina*, etc.
- `micro-rdk-server`: The binary we make available for download.


Cargo.toml
```diff
@@ -1,86 +1,34 @@
...
-resolver = "2"
-version = "0.1.6"
 rust-version = "1.71"
-links = "micro_rdk"
-
-[lib]
-crate-type = ["lib"]
+version = "0.1.6"
 
 [patch.crates-io]
```
npmenard, line 20: I think it make sense for now to keep the patches here for now since examples & micro-rdk needs them when building thier respective targets.

Cargo.toml
```diff
@@ -1,86 +1,34 @@
...
+[workspace.package]
 authors = ["Viam, inc. <contact@viam.com>"]
 description = "Viam RDK for microcontroller"
 edition = "2021"
 license = "AGPL-3.0"
-name = "micro-rdk"
 repository = "https://github.com/viamrobotics/micro-rdk"
-resolver = "2"
-version = "0.1.6"
 rust-version = "1.71"
```
npmenard, line 17: This will be shared with all crates in the worksapce? Can it be overwritten if we wanted to? (ex may micro-rdk-installer will need a more recent version of rust than the one used by micro-rdk?)

Cargo.toml
```diff
@@ -1,86 +1,34 @@
-[package]
+[workspace]
+resolver = "2"
+
+members = [
+        "examples",
```
npmenard, line 5: nit rename examples to micro-rdk-examples

npmenard (commented):
I like this approach seems more readable and easier to navigate for a newcomer.

I don't think we need to split micro-rdk further at least for this PR, do you think it will be much work if we do it later?

Also we need to keep the Cargo.lock at least for examples & micro-rdk-installer so build can be reproduce easily can we keep them?

acmorrow:
> I like this approach seems more readable and easier to navigate for a newcomer.

Great!

> 
> I don't think we need to split micro-rdk further at least for this PR, do you think it will be much work if we do it later?
> 

No, I think we should get the work from this review across the line and make sure we are happy living in a workspace'd world. Then we can plan out the next step of splitting up the `micro-rdk` crate as follow-on work.


> Also we need to keep the Cargo.lock at least for examples & micro-rdk-installer so build can be reproduce easily can we keep them?

Within a workspace there is only one top-level `Cargo.lock`. It is a bit of a puzzle to me, actually, how this is supposed to work for a workspace that has some libraries (lockfile should not be commited) and some programs (lockfile should be committed). I'll look into that and follow up.

gvaradarajan (commented):
I agree with this layout conceptually (provided the macros can still be imported properly for modules and the installer is able to build) and I'm eager to have the crate splitting as a follow-up. The single Cargo.lock file does give me pause, but maybe it's ok if we don't have a lock file at all?

examples/Cargo.toml
```diff
@@ -67,3 +67,9 @@ path = "esp32/esp32-server.rs"
 [[bin]]
 name = "esp32-server-with-cred"
 path = "esp32-with-cred/esp32-server-with-cred.rs"
+
+[package.metadata.esp-idf-sys]
+esp_idf_tools_install_dir = "global"
+#esp_idf_tools_install_dir = "global"
+#esp_idf_sdkconfig = "examples/sdkconfig.defaults"
```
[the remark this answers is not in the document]
npmenard: done

micro-rdk/src/esp32/i2c.rs
```diff
@@ -51,7 +51,7 @@ impl TryFrom<&Kind> for Esp32I2cConfig {
         }
         let mut baudrate_hz: u32 = 1000000;
         if value.contains_key("baudrate_hz")? {
-            baudrate_hz = value.get("baudrate")?.unwrap().try_into()?;
+            baudrate_hz = value.get("baudrate_hz")?.unwrap().try_into()?;
```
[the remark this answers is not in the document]
npmenard: Nope @gvaradarajan what is the right syntax?

npmenard:
> @npmenard - This LGTM. One or two small comments.
> 
> How do you want to proceed? If you like, I can take on getting this rebased (there are conflicts) and merged.
> 
> I think you will need to hit the approve button though, since I can't approve my own PR...

I just merged the diff still need to check that changes to the CI are correct. You can merge without approval :)

acmorrow:
> > @npmenard - This LGTM. One or two small comments.
> > How do you want to proceed? If you like, I can take on getting this rebased (there are conflicts) and merged.
> > I think you will need to hit the approve button though, since I can't approve my own PR...
> 
> I just merged the diff still need to check that changes to the CI are correct. You can merge without approval :)

Great, I'll merge if/when the CI checks pass.

acmorrow:
@npmenard - Looks like the clippy run didn't quite work: https://github.com/viamrobotics/micro-rdk/actions/runs/8158406229/job/22300277344?pr=149

npmenard (approved):
LGTM

PR #150, opened 2024-02-06
fix incorrect key used for baudrate in i2c configuration

gvaradarajan:
really should have caught this earlier


npmenard (approved):
:P LGTM

PR #151, opened 2024-02-07
Update protos

npmenard:
Bump the protos

### DIFF
--- a/src/common/robot.rs
+++ b/src/common/robot.rs
@@ -1058,6 +1058,7 @@ mod tests {
                     },
                 )]),
             }),
+            ..Default::default()
         };
         component_cfgs.push(comp);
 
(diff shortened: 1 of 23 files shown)


src/common/robot.rs
```diff
@@ -1058,6 +1058,7 @@ mod tests {
                     },
                 )]),
             }),
+            ..Default::default()
```
acmorrow, line 1061: This syntax is new to me, could you explain?

acmorrow (approved):
LGTM. I did not inspect the changes under the `gen` directory. I did have one question for a bit of syntax I don't recognize.

src/common/robot.rs
npmenard: this will set all fields that are not explicitly set to their default value

PR #154, opened 2024-03-06
Workspace based dependencies

acmorrow:
This had been something I'd hoped to do as part of converting to a workspace per #149, but since I didn't finish the work for that PR it wasn't included. It's maybe controversial? I'm not sure.

Anyway, this change moves us to use workspace dependency inheritance for all dependencies in the workspace crates.

The only downside to doing this that I'm aware of is that `cargo add` doesn't currently have support for adding directly at the workspace level, so if we `cargo add` a new dependency, we will need to manually pull it up to the top-level `Cargo.toml`, at least until https://github.com/rust-lang/cargo/issues/10608 gets resolved one way or another.

On the other hand, with this change in place, if we do `cargo add` of an existing dependency in a new location, it will import it with `.workspace = true`, which is nice.

Overall, I think this will make it easier to see what dependencies we are declaring at a glance, and easier to make upgrades in one place for crates used in multiple workspace sub-crates.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,5 +37,78 @@ opt-level = "s"
 debug = true # Symbols are nice and they don't increase the size on Flash
 opt-level = "z"
 
+[workspace.dependencies]
+anyhow = "1.0.71"
+async-channel = "1.8.0"
+async-io = "=1.12.0"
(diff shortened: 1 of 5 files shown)


npmenard (approved):
LGTM

PR #155, opened 2024-03-08
[RSDK-6857] support multiple connections

npmenard:


micro-rdk/src/common/conn/server.rs
```diff
@@ -561,3 +561,75 @@ where
...
+        let _ = self.task.replace(task);
+        let _ = self.prio.replace(prio);
+    }
+    fn is_finished(&self) -> bool {
+        if let Some(task) = self.task.as_ref() {
+            return task.is_finished();
+        }
+        true
+    }
+    // fn last_error(&mut self) -> Option<ServerError> {
```
acmorrow, line 582: Remove

micro-rdk/src/native/entry.rs
```diff
@@ -100,7 +100,7 @@ pub async fn serve_web_inner(
         exec.clone(),
     ));
 
-    let mut srv = ViamServerBuilder::new(mdns, cloned_exec, client_connector, app_config)
+    let mut srv = ViamServerBuilder::new(mdns, cloned_exec, client_connector, app_config, 3)
```
acmorrow, line 103: Below, does it matter that the ESP32 and native `serve_web` functions have diverged, since the ESP32 one now takes the maximum as a parameter and the native one does not?

acmorrow, line 103: Should `native` be permitted more than 3? Or do you want to keep it the same with the idea that it should be a fairly faithful reproduction of how an ESP32 micro-rdk behaves?

acmorrow (approved):
LGTM mod some commented out code, and one unimportant question.

micro-rdk/src/native/entry.rs
npmenard: I think its fine as long as we don't have a use case for native beyond testing we don't need to polish it too much

npmenard: same comment as above

gvaradarajan (approved):
LGTM as well

PR #158, opened 2024-03-13
add functionality to import micro-RDK as a library

gvaradarajan:
This changes things so that, by default, micro-RDK will be compiled as a library dependency in an external module or project (in other words, esp-idf-svc will not be compiled with the `binstart` feature in micro-RDK)

For all projects meant to be the main application on an esp32, the `binstart` feature will have to be specified for the micro-RDK dependency (see the changes in `examples/Cargo.toml` for an example)

This PR also exposes the `libstart` feature on esp-idf-svc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ Cargo.lock
 /examples/viam.json
 /canary/.venv/*
 .vscode
+components_esp32.lock
--- a/components_esp32.lock
+++ b/components_esp32.lock
@@ -1,0 +1,9 @@
+dependencies:
+  idf:
+    component_hash: null
+    source:
+      type: idf
+    version: 4.4.4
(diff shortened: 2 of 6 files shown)


components_esp32.lock
```diff
@@ -0,0 +1,9 @@
+dependencies:
+  idf:
+    component_hash: null
+    source:
```
npmenard, line 4: don't comit this file

.gitignore
```diff
@@ -4,3 +4,4 @@ Cargo.lock
 /examples/viam.json
 /canary/.venv/*
 .vscode
+components_esp32.lock
```
acmorrow, line 7: Thanks for adding this. Do we know what this file does?

acmorrow (approved):
LGTM

.gitignore
gvaradarajan: It looks like it's an auto-generated list of components for a project?

PR #159, opened 2024-03-14
[RSDK-6943] Remove anyhow from webrtc

npmenard:


micro-rdk/src/common/webrtc/candidates.rs
```diff
@@ -6,6 +6,19 @@ use std::{
 };
 
 use stun_codec::TransactionId;
+use thiserror::Error;
+
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum CandidateError {
+    #[error("cannot parse candidate")]
```
gvaradarajan, line 13: can we keep the printing of the candidate string in the error?

gvaradarajan (approved):

micro-rdk/src/common/webrtc/candidates.rs
npmenard: I rather not allocate a String for that it happens somewhat often and is not a critical failure

PR #160, opened 2024-03-18
RSDK-6940 - feature-gate builtin drivers

gvaradarajan:


micro-rdk/Cargo.toml
```diff
@@ -21,6 +21,7 @@ libstart = ["esp-idf-svc/libstart"]
 camera = []
 esp32 = ["dep:esp-idf-svc", "dep:embedded-svc", "dep:embedded-hal"]
 native = ["dep:rustls", "dep:webpki-roots", "dep:rustls-pemfile", "dep:mdns-sd", "dep:local-ip-address", "dep:openssl", "dep:rcgen", "dep:async-std-openssl"]
+builtin = []
```
npmenard, line 24: should be a default feature

micro-rdk/src/builtin/fake.rs
```diff
@@ -0,0 +1,569 @@
+use std::collections::HashMap;
```
npmenard, line 1: I don't think we should put all fake components in the same file

micro-rdk/src/builtin/mod.rs
```diff
@@ -0,0 +1,15 @@
+pub mod adxl345;
+#[cfg(feature = "esp32")]
+pub mod esp32_encoder;
+pub mod fake;
+pub mod gpio_motor;
+pub mod gpio_servo;
+#[cfg(feature = "esp32")]
```
npmenard, line 7: I would like to keep esp32 specific components in the esp32 folder

micro-rdk/Cargo.toml
npmenard: An name it builtin-components

npmenard (changes requested):
Done with first pass. I don't think we should reorganize the code in this PR we have a lot of changes in the pipeline and I don't want us to spend time rebasing. let's add cfg guard in the files it's not pretty but we can also expand them easily as we go until we have a better idea for the path forward

micro-rdk/src/common/motor.rs
```diff
@@ -1,32 +1,49 @@
...
+#[cfg(feature = "builtin-components")]
+use std::collections::HashMap;
 
 pub static COMPONENT_NAME: &str = "motor";
 
 #[derive(Error, Debug)]
 pub enum MotorError {
     #[error("invalid motor configuration")]
     InvalidMotorConfig,
+    #[cfg(feature = "builtin-components")]
```
npmenard, line 43: Why did you have to feature gates the errors?

micro-rdk/src/esp32/mod.rs
```diff
@@ -6,16 +6,20 @@ pub mod board;
 pub mod camera;
 pub mod certificate;
 pub mod dtls;
+#[cfg(feature = "builtin-components")]
```
npmenard, line 9: can you group them in brackets?

micro-rdk/src/esp32/pulse_counter.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
npmenard, line 1: what was the complain form clippy here?

npmenard (commented):
LGTM just a couple of nits here and there :

npmenard (approved):

micro-rdk/src/esp32/pulse_counter.rs
gvaradarajan: I think it was the unused code problem, but there was something I was doing weird that I fixed so this might not be an issue anymore. I'll check

micro-rdk/src/esp32/mod.rs
gvaradarajan: I actually tried, it got mad at me

gvaradarajan: to clarify, this yields a syntax error
```rust
#[cfg(feature = "builtin-components")]
{
    pub mod encoder;
    pub mod hcsr04;
    pub mod single_encoded_motor;
    pub mod single_encoder;
}
```
and I'm struggling to find another way to do this

acmorrow: I went looking for a way to do this too and failed. So much for `M-x sort-lines` on this block!

acmorrow (approved):
LGTM

PR #161, opened 2024-03-18
RSDK-6860 - wrap AnalogReaders in Arc<Mutex> so Board implements Send

gvaradarajan:
PR untested, will test that analogs still work on physical hardware as code is being reviewed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,7 +94,7 @@ build-esp32-with-cred-bin:
 
 flash-esp32-bin:
 ifneq (,$(wildcard ./target/xtensa-esp32-espidf/esp32-server.bin))
-	espflash write-bin 0x0 ./examples/target/esp32-server.bin -b 460800  && sleep 2 && espflash monitor
+	espflash write-bin 0x0 ./target/xtensa-esp32-espidf/esp32-server.bin -b 460800  && sleep 2 && espflash monitor
 else
 	$(error esp32-server.bin not found, run make build-esp32-bin first)
 endif
--- a/micro-rdk/src/common/analog.rs
+++ b/micro-rdk/src/common/analog.rs
@@ -1,6 +1,7 @@
 #![allow(dead_code)]
 
 use super::config::{AttributeError, Kind};
+use std::sync::{Arc, Mutex};
 use thiserror::Error;
 
 #[derive(Error, Debug)]
@@ -42,6 +43,19 @@ pub trait AnalogReader<Word> {
     fn name(&self) -> String;
 }
 
+impl<A, Word> AnalogReader<Word> for Arc<Mutex<A>>
+where
+    A: ?Sized + AnalogReader<Word>,
+{
+    type Error = A::Error;
+    fn read(&mut self) -> Result<Word, Self::Error> {
+        self.lock().unwrap().read()
+    }
+    fn name(&self) -> String {
+        self.lock().unwrap().name()
+    }
+}
+
 pub(crate) struct AnalogReaderConfig {
     pub(crate) name: String,
     pub(crate) pin: i32,
@@ -61,6 +75,8 @@ impl TryFrom<&Kind> for AnalogReaderConfig {
         Ok(Self { name, pin })
     }
 }
+
+pub type AnalogReaderType<W, E = AnalogError> = Arc<Mutex<dyn AnalogReader<W, Error = E>>>;
 
 #[cfg(test)]
 mod tests {
--- a/micro-rdk/src/common/board.rs
+++ b/micro-rdk/src/common/board.rs
@@ -7,12 +7,11 @@ use crate::{
     proto::{common, component},
 };
 
-use core::cell::RefCell;
 use log::*;
-use std::{collections::HashMap, rc::Rc, sync::Arc, sync::Mutex, time::Duration};
+use std::{collections::HashMap, sync::Arc, sync::Mutex, time::Duration};
 
 use super::{
-    analog::{AnalogError, FakeAnalogReader},
+    analog::{AnalogReaderType, FakeAnalogReader},
     config::ConfigType,
     generic::DoCommand,
     i2c::{FakeI2CHandle, FakeI2cConfig, I2CErrors, I2CHandle, I2cHandleType},
@@ -64,10 +63,7 @@ pub trait Board: Status + DoCommand {
     fn get_gpio_level(&self, pin: i32) -> Result<bool, BoardError>;
 
     /// Get an [AnalogReader] by name
-    fn get_analog_reader_by_name(
-        &self,
-        name: String,
-    ) -> Result<Rc<RefCell<dyn AnalogReader<u16, Error = AnalogError>>>, BoardError>;
+    fn get_analog_reader_by_name(&self, name: String) -> Result<AnalogReaderType<u16>, BoardError>;
 
     /// Set the board to the indicated [PowerMode](component::board::v1::PowerMode)
     fn set_power_mode(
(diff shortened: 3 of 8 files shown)


micro-rdk/src/common/board.rs
```diff
@@ -180,7 +178,7 @@ impl Board for FakeBoard {
             digital_interrupts: HashMap::new(),
         };
         self.analogs.iter().for_each(|a| {
-            let mut borrowed = a.borrow_mut();
+            let mut borrowed = a.clone();
```
acmorrow, line 181: I suspect `borrowed` no longer accurately reflects what is happening here.

micro-rdk/src/common/board.rs
```diff
@@ -199,8 +197,12 @@ impl Board for FakeBoard {
     fn get_analog_reader_by_name(
         &self,
         name: String,
-    ) -> Result<Rc<RefCell<dyn AnalogReader<u16, Error = AnalogError>>>, BoardError> {
-        match self.analogs.iter().find(|a| a.borrow().name() == name) {
+    ) -> Result<Arc<Mutex<dyn AnalogReader<u16, Error = AnalogError>>>, BoardError> {
+        match self
+            .analogs
+            .iter()
+            .find(|a| a.lock().unwrap().name() == name)
```
acmorrow, line 204: [Why] do you still need the explicit `lock().unwrap()` here?

micro-rdk/src/common/board.rs
```diff
@@ -253,7 +255,7 @@ impl Status for FakeBoard {
         let mut hm = HashMap::new();
         let mut analogs = HashMap::new();
         self.analogs.iter().for_each(|a| {
-            let mut borrowed = a.borrow_mut();
+            let mut borrowed = a.clone();
```
acmorrow, line 258: Same thoughts re `borrowed`.

micro-rdk/src/common/analog.rs
```diff
@@ -42,6 +43,19 @@ pub trait AnalogReader<Word> {
...
+{
+    type Error = A::Error;
+    fn read(&mut self) -> Result<Word, Self::Error> {
+        self.lock().unwrap().read()
+    }
+    fn name(&self) -> String {
+        self.lock().unwrap().name()
+    }
+}
+
```
acmorrow, line 58: Should there be a `pub AnalogReaderType`? Something like `pub type AnalogReaderType<W, E = AnalogError> = Arc<Mutex<dyn AnalogReader<W, Error = E>>>;` That'd let you simplify declarations elsewhere, if it works.

micro-rdk/src/common/moisture_sensor.rs
```diff
@@ -37,7 +36,7 @@ impl Readings for MoistureSensor {
 
 impl SensorT<f64> for MoistureSensor {
     fn get_readings(&self) -> anyhow::Result<TypedReadingsResult<f64>> {
-        let reading = self.analog.borrow_mut().read()?;
+        let reading = self.analog.lock().unwrap().read()?;
```
acmorrow, line 39: Also curious why you need the explicit lock/unwrap here.

micro-rdk/src/esp32/analog.rs
```diff
@@ -2,20 +2,19 @@
...
 use crate::esp32::esp_idf_svc::hal::adc::{AdcChannelDriver, AdcDriver};
 use crate::esp32::esp_idf_svc::hal::gpio::ADCPin;
-use core::cell::RefCell;
-use std::rc::Rc;
+use std::sync::{Arc, Mutex};
 
 pub struct Esp32AnalogReader<'a, const A: u32, T: ADCPin> {
     channel: AdcChannelDriver<'a, A, T>,
-    driver: Rc<RefCell<AdcDriver<'a, T::Adc>>>,
+    driver: Arc<Mutex<AdcDriver<'a, T::Adc>>>,
```
acmorrow, line 9: If the `Arc Mutex` wrapping access to `AnalogReader`s is in play, does the embedded `AdcDriver` for the `Esp32AnalogReader` need its own level of `Arc Mutex`'ing?

micro-rdk/src/esp32/board.rs
```diff
@@ -390,7 +388,7 @@ impl Status for EspBoard {
         let mut hm = HashMap::new();
         let mut analogs = HashMap::new();
         self.analogs.iter().for_each(|a| {
-            let mut borrowed = a.borrow_mut();
+            let mut borrowed = a.clone();
```
acmorrow, line 391: Ditto

acmorrow (changes requested):

micro-rdk/src/common/board.rs
gvaradarajan: hmm, I probably don't? I don't remember why I added this

micro-rdk/src/common/analog.rs
gvaradarajan: yes! I meant to do this to match the I2CHandle pattern but I forgot

micro-rdk/src/esp32/analog.rs
gvaradarajan: yes, because the gRPC call grabs an analog reader and then calls read on it, meaning that the analog reader instance needs to implement Send

micro-rdk/src/common/moisture_sensor.rs
gvaradarajan: here I needed it because the receiver is `&self` and not `&mut self`, which forces me to rely on interior mutability (which `MutexGuard` has)

gvaradarajan:
tested on hardware, works (used a temperature sensor)

Makefile
```diff
@@ -94,7 +94,7 @@ build-esp32-with-cred-bin:
 
 flash-esp32-bin:
 ifneq (,$(wildcard ./target/xtensa-esp32-espidf/esp32-server.bin))
-	espflash write-bin 0x0 ./examples/target/esp32-server.bin -b 460800  && sleep 2 && espflash monitor
+	espflash write-bin 0x0 ./target/xtensa-esp32-espidf/esp32-server.bin -b 460800  && sleep 2 && espflash monitor
```
npmenard, line 97: <3

micro-rdk/src/common/board.rs
```diff
@@ -102,20 +103,20 @@ pub trait Board: Status + DoCommand {
 }
 
 /// An alias for a thread-safe handle to a struct that implements the [Board] trait
-pub type BoardType = Arc<Mutex<dyn Board>>;
+pub type BoardType = Arc<Mutex<dyn Board<u16>>>;
```
npmenard, line 105: i don't like the idea to tie a generic to Board especially that for now we only read `u16` Did you try W=u16? otherwise I would leave it to u16 default for now we can revisit it once we need to support another integer type

npmenard (changes requested):
Done with first pass I have strong reservation on having a generic on board otherwise looks good

acmorrow (approved):
LGTM but I share @npmenard's discomfort around the word size becoming a generic parameter of `Board`.

micro-rdk/src/common/board.rs
gvaradarajan: I'm not sure I understand what you mean by W=u16. I did try this
```rust
pub type BoardType = Arc<Mutex<dyn Board<AnalogReaderWord = u16>>>
```
but it did not compile

acmorrow (approved):
LGTM

npmenard (approved):
LGTM

PR #164, opened 2024-03-20
[RSDK-6913] & [RSDK-6914] ServoError+BaseError+GenericError

npmenard:


micro-rdk/src/esp32/mod.rs
```diff
@@ -15,6 +15,7 @@ pub mod exec;
 pub mod hcsr04;
 pub mod i2c;
 pub mod pin;
+#[cfg(feature = "builtin-components")]
```
gvaradarajan, line 18: are we sure about this one? It's really more of a peripheral than it is a component

gvaradarajan (approved):
LGTM

micro-rdk/src/esp32/mod.rs
npmenard: it gives clippy error with "--features esp32 --no-default-features"

gvaradarajan: oh I see, nvm

acmorrow (approved):
LGTM

PR #165, opened 2024-03-20
RSDK-6992 RSDK-6993 - implement DataCollector pt. 1

gvaradarajan:
Part 1 of 2 PRs

The TODO comments are there to provide a sense of the direction of the architecture moving forward and will be removed before merge

For now, to run the tests introduced in the PR, add "data" to the default features in micro-rdk/Cargo.toml before calling `make test`

### DIFF
--- a/micro-rdk/Cargo.toml
+++ b/micro-rdk/Cargo.toml
@@ -22,6 +22,7 @@ builtin-components = []
 camera = []
 esp32 = ["dep:esp-idf-svc", "dep:embedded-svc", "dep:embedded-hal", "esp-idf-svc/std", "esp-idf-svc/alloc"]
 native = ["dep:rustls", "dep:webpki-roots", "dep:rustls-pemfile", "dep:mdns-sd", "dep:local-ip-address", "dep:openssl", "dep:rcgen", "dep:async-std-openssl"]
+data = []
 
 [dev-dependencies]
 test-log.workspace = true
--- a/micro-rdk/src/common/data_collector.rs
+++ b/micro-rdk/src/common/data_collector.rs
@@ -1,0 +1,318 @@
+use std::fmt::Display;
+
+use crate::google::protobuf::Timestamp;
+use crate::proto::app::data_sync::v1::{SensorData, SensorMetadata};
+
+use super::{
+    config::{AttributeError, Kind},
+    movement_sensor::MovementSensor,
+    robot::ResourceType,
+    sensor::{Readings, SensorError},
+};
+
+use chrono::offset::Local;
+use thiserror::Error;
+
+/// A DataCollectorConfig instance is a representation of an element
+/// of the list of "capture_methods" in the "attributes" section of a
+/// component's configuration JSON object as stored in app. Each element
+/// of "capture_methods" is meant to produce an instance of `DataCollector`
+/// as defined below
+#[derive(Debug, Clone)]
+pub struct DataCollectorConfig {
+    pub method: CollectionMethod,
+    pub capture_frequency_hz: f32,
+}
+
+impl TryFrom<&Kind> for DataCollectorConfig {
+    type Error = AttributeError;
+    fn try_from(value: &Kind) -> Result<Self, Self::Error> {
+        let method_str: String = value
+            .get("method")?
+            .ok_or(AttributeError::KeyNotFound("method".to_string()))?
+            .try_into()?;
+        let capture_frequency_hz = value
+            .get("capture_frequency_hz")?
+            .ok_or(AttributeError::KeyNotFound(
+                "capture_frequency_hz".to_string(),
+            ))?
+            .try_into()?;
+        // TODO: RSDK-7127 - Collectors that take arguments (ex. Board Analogs)
+        let method = match method_str.as_str() {
+            "Readings" => CollectionMethod::Readings,
+            "AngularVelocity" => CollectionMethod::AngularVelocity,
+            "LinearAcceleration" => CollectionMethod::LinearAcceleration,
+            "LinearVelocity" => CollectionMethod::LinearVelocity,
+            _ => {
+                return Err(AttributeError::ConversionImpossibleError);
+            }
+        };
+        Ok(DataCollectorConfig {
+            method,
+            capture_frequency_hz,
+        })
+    }
+}
+
+/// A CollectionMethod is an enum whose values are associated with
+/// a method on one or more component traits
+#[derive(Debug, Clone)]
+pub enum CollectionMethod {
+    Readings,
+    // MovementSensor methods
+    AngularVelocity,
+    LinearAcceleration,
+    LinearVelocity,
+    // TODO: RSDK-7127 - Implement collectors for all other applicable components/methods
+}
+
+impl Display for CollectionMethod {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        std::fmt::Display::fmt(
+            match self {
+                Self::Readings => "readings",
+                Self::AngularVelocity => "angularvelocity",
+                Self::LinearAcceleration => "linearacceleration",
+                Self::LinearVelocity => "linearvelocity",
+            },
+            f,
+        )
+    }
+}
+
+#[derive(Error, Debug)]
+pub enum DataCollectionError {
+    #[error("method {0} unsupported for {1}")]
+    UnsupportedMethod(CollectionMethod, String),
+    #[error("no collection methods supported for component")]
+    NoSupportedMethods,
+    #[error(transparent)]
+    SensorCollectionError(#[from] SensorError),
+}
+
+/// A DataCollector represents an association between a data collection method and
+/// a ResourceType (i.e. SensorType & Readings, BoardType & Analogs) and the frequency at
+/// which the results of the method should be stored.
+pub struct DataCollector {
+    name: String,
+    component_type: String,
+    resource: ResourceType,
+    method: CollectionMethod,
+    time_interval_ms: u64,
+}
+
+fn resource_method_pair_is_valid(resource: &ResourceType, method: &CollectionMethod) -> bool {
+    match resource {
+        ResourceType::Sensor(_) => matches!(method, CollectionMethod::Readings),
+        ResourceType::MovementSensor(_) => matches!(
+            method,
+            CollectionMethod::Readings
+                | CollectionMethod::AngularVelocity
+                | CollectionMethod::LinearAcceleration
+                | CollectionMethod::LinearVelocity
+        ),
+        _ => false,
+    }
+}
+
+impl DataCollector {
+    pub fn new(
+        name: String,
+        resource: ResourceType,
+        method: CollectionMethod,
+        capture_frequency_hz: f32,
+    ) -> Result<Self, DataCollectionError> {
+        let time_interval_ms = (1000.0 / capture_frequency_hz) as u64;
+        let component_type = resource.component_type();
+        if !resource_method_pair_is_valid(&resource, &method) {
+            return Err(DataCollectionError::UnsupportedMethod(
+                method,
+                component_type,
+            ));
+        }
+        Ok(DataCollector {
+            name,
+            component_type,
+            resource,
+            method,
+            time_interval_ms,
+        })
+    }
+
+    pub fn from_config(
+        name: String,
+        resource: ResourceType,
+        conf: DataCollectorConfig,
+    ) -> Result<Self, DataCollectionError> {
+        Self::new(name, resource, conf.method, conf.capture_frequency_hz)
+    }
+
+    pub fn name(&self) -> String {
+        self.name.to_string()
+    }
+
+    pub fn component_type(&self) -> String {
+        self.component_type.to_string()
+    }
+
+    pub fn time_interval(&self) -> u64 {
+        self.time_interval_ms
+    }
+
+    pub fn method_str(&self) -> String {
+        self.method.to_string()
+    }
+
+    /// calls the method associated with the collector and returns the resulting data
+    pub(crate) fn call_method(&mut self) -> Result<SensorData, DataCollectionError> {
+        let reading_requested_dt = Local::now().fixed_offset();
+        let data = match &mut self.resource {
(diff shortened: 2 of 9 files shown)


micro-rdk/src/common/data_collector.rs
```diff
@@ -0,0 +1,352 @@
...
+    #[error(transparent)]
+    SensorCollectionError(#[from] SensorError),
+    // TODO: remove when error enums for all components are created (RSDK-6909)
+    #[error(transparent)]
+    MiscCollectionError(#[from] anyhow::Error),
+}
+
+/// A DataCollector represents an association between a data collection method and
+/// a ResourceType (i.e. SensorType & Readings, BoardType & Analogs) and the frequency at
+/// which the results of the method should be stored. It is capable
```
acmorrow, line 127: Looks like the last sentence is incomplete. I was going to say a sentence fragement, but that isn't exactly true.

micro-rdk/src/common/data_collector.rs
```diff
@@ -0,0 +1,352 @@
...
+    type Error = AttributeError;
+    fn try_from(value: &Kind) -> Result<Self, Self::Error> {
+        if !value.contains_key("method")? {
+            return Err(AttributeError::KeyNotFound("method".to_string()));
+        }
+        if !value.contains_key("capture_frequency_hz")? {
+            return Err(AttributeError::KeyNotFound(
+                "capture_frequency_hz".to_string(),
+            ));
+        }
```
acmorrow, line 35: Looks like these cases are already handled via `ok_or` below?

micro-rdk/src/common/data_collector.rs
```diff
@@ -0,0 +1,352 @@
...
+    LinearVelocity,
+    // TODO: PowerSensor methods
+    // Voltage,
+    // Current,
+    // TODO: Board
+    // Analogs(String)
+}
+
+impl CollectionMethod {
+    fn method_str(&self) -> String {
```
acmorrow, line 92: If the only real use of this is for `fmt`, maybe just inline it in there rather than synthesizing a new method name?

micro-rdk/src/common/data_collector.rs
```diff
@@ -0,0 +1,352 @@
...
+            method,
+            capture_frequency_hz,
+        })
+    }
+}
+
+/// A CollectionMethod is an enum whose values are associated with
+/// a method on one or more component traits
+#[derive(Debug, Clone)]
+pub enum CollectionMethod {
```
acmorrow, line 78: This design is very centralized / closed / static, in that if a new value returning method was added to one of these component types, we would need to make a compensating change here before it could be used with data capture. I'm not necessarily saying that's a bad design: it makes it a lot easier to see what is going on. The other way to do this would be to build into each component some ability to enumerate the method names / return Fn objects for each. That'd result in a more decentralized and open design. That said, nothing prevents us from moving to such a design later. So I think on balance I'm fine with starting with the closed design.

micro-rdk/src/common/data_collector.rs
```diff
@@ -0,0 +1,352 @@
...
+            .ok_or(AttributeError::KeyNotFound(
+                "capture_frequency_hz".to_string(),
+            ))?
+            .try_into()?;
+        // TODO: Collectors that take arguments (ex. Board Analogs)
+        // let parameters: &Kind = if let Ok(Some(params)) = value.get("additional_params") {
+        //     params
+        // } else {
+        //     &Kind::NullValue(0)
+        // };
```
acmorrow, line 51: I think maybe remove the commented out code under the `TODO` sections (here and elsewhere) in favor of just a description. The code is unlikely to remain useful and correct because as things change under review these probably won't keep up.

micro-rdk/src/common/data_collector.rs
```diff
@@ -0,0 +1,352 @@
...
+}
+
+impl DataCollector {
+    pub fn new(
+        name: String,
+        resource: ResourceType,
+        method: CollectionMethod,
+        capture_frequency_hz: f32,
+    ) -> Result<Self, DataCollectionError> {
+        let time_interval_ms = ((1.0 / capture_frequency_hz) * 1000.0) as u64;
```
acmorrow, line 159: Or just `1000.0 / capture_frequence_hz as u64`

micro-rdk/src/common/data_collector.rs
```diff
@@ -0,0 +1,352 @@
...
+    pub fn name(&self) -> String {
+        self.name.to_string()
+    }
+
+    pub fn component_type(&self) -> String {
+        self.component_type.to_string()
+    }
+
+    pub fn method_str(&self) -> String {
+        self.method.method_str()
```
acmorrow, line 193: `self.method.to_string()`

micro-rdk/src/common/data_collector.rs
acmorrow: And anyway don't you get `ToString` for free if you implement `Display`?

acmorrow (changes requested):

micro-rdk/src/common/sensor.rs
```diff
@@ -57,6 +65,37 @@ pub trait Readings {
...
+        metadata: Some(SensorMetadata {
+            time_received: Some(Timestamp {
+                seconds: current_date.timestamp(),
+                nanos: current_date.timestamp_subsec_nanos() as i32,
+            }),
+            time_requested: Some(Timestamp {
+                seconds: current_date.timestamp(),
+                nanos: current_date.timestamp_subsec_nanos() as i32,
+            }),
+        }),
```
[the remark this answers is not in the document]
gvaradarajan: this was a mistake, I meant to initialize the `time_requested` before the call to `get_sensor_readings`

micro-rdk/src/common/data_collector.rs
gvaradarajan: sorry I think this was a miscommunication on my part, but it's my plan to delete the TODO comments before merge (although on reflection, I like keeping the comment and removing the commented-out code, thoughts?)

micro-rdk/src/common/data_collector.rs
gvaradarajan: Yeah I had considered that open design and I do agree that I like it better in theory, but I did not want to grapple with the lifetime considerations for function objects, so the design decision was born partially out of laziness

micro-rdk/src/common/data_collector.rs
gvaradarajan: both fair points

micro-rdk/src/common/data_collector.rs
gvaradarajan: yeah, should have backspaced harder

micro-rdk/src/common/data_collector.rs
gvaradarajan: yeah my bad, wrote the way I think and then forgot to clean up

micro-rdk/src/common/robot.rs
```diff
@@ -64,6 +64,24 @@ pub enum ResourceType {
 pub type Resource = ResourceType;
 pub type ResourceMap = HashMap<ResourceName, Resource>;
 
+#[cfg(feature = "data")]
```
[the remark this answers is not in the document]
gvaradarajan: I was concerned I would get a dead code complaint, but we actually ignore those at the top of the file. Let me try removing it

micro-rdk/src/common/sensor.rs
```diff
@@ -57,6 +65,37 @@ pub trait Readings {
     fn get_generic_readings(&mut self) -> Result<GenericReadingsResult, SensorError>;
 }
 
+#[cfg(feature = "data")]
+pub fn get_sensor_readings_data(sensor: &mut dyn Readings) -> Result<SensorData, SensorError> {
```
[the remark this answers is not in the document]
gvaradarajan: Yeah, fair this function isn't quite the answer, I'll implement your second suggestion

micro-rdk/src/common/sensor.rs
```diff
@@ -57,6 +65,37 @@ pub trait Readings {
...
+    let readings = sensor.get_generic_readings()?;
+    let data_struct = Data::Struct(google::protobuf::Struct {
+        fields: HashMap::from([(
+            "readings".to_string(),
```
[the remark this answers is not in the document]
gvaradarajan: After some experimentation and examination of the analogous code in RDK, I discovered that this is the format that the data management service on app understands how to parse for the readings collection method.

micro-rdk/src/common/data_collector.rs
acmorrow: Keep the TODOs, with as much explication as you see fit, but drop the code would be my recommendation.

micro-rdk/src/common/data_collector.rs
acmorrow: Laziness - one of the principle virtues of the programmer! I'm fine with the closed approach, for now.

micro-rdk/src/common/sensor.rs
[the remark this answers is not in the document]
acmorrow: Does it basically expect the method name to be played back as the key or something?

micro-rdk/src/common/sensor.rs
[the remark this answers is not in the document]
acmorrow: I think the same mistake is exists in the other place you do this too.

micro-rdk/src/common/sensor.rs
```diff
@@ -51,10 +57,46 @@ pub(crate) fn register_models(registry: &mut ComponentRegistry) {
...
+    }
+}
+
 pub type TypedReadingsResult<T> = ::std::collections::HashMap<String, T>;
 
 pub trait Readings {
     fn get_generic_readings(&mut self) -> Result<GenericReadingsResult, SensorError>;
+    #[cfg(feature = "data")]
+    fn get_readings_data(&mut self) -> Result<SensorData, SensorError> {
+        let reading_requested_dt = chrono::offset::Local::now().fixed_offset();
```
acmorrow, line 82: Wondering a bit about whether Metadata production belongs in the leaf functions here. If `get_readings_data` returned a `sensor_data::Data`, and `Vector3::to_sensor_data` the same, could timestamp management all be lifted up to `DataCollector`? It'd let you get, for instance, the timestamp fields out of the arguments to `Vector3::to_sensor_data`, which in turn would probably let you use `From` there as well.

The downside I see to that choice would be if the metadata later was expanded in a way that could only be satisfied by the objects themselves.

What do you think?

micro-rdk/src/common/sensor.rs
```diff
@@ -18,6 +18,12 @@ use super::generic::DoCommand;
 use super::i2c::I2CErrors;
 
 use thiserror::Error;
+
+#[cfg(feature = "data")]
+use crate::google::protobuf::Timestamp;
+#[cfg(feature = "data")]
+use crate::proto::app::data_sync::v1::{sensor_data::Data, SensorData, SensorMetadata};
+
```
acmorrow, line 26: I think these can be combined.

micro-rdk/src/common/data_collector.rs
```diff
@@ -0,0 +1,313 @@
+#![allow(dead_code)]
```
acmorrow, line 1: Is there actual dead code in this file? If so can this be scoped more narrowly (genuinely asking, as I don't know).

micro-rdk/src/common/data_collector.rs
```diff
@@ -0,0 +1,313 @@
...
+        self.time_interval_ms
+    }
+
+    pub fn method_str(&self) -> String {
+        self.method.to_string()
+    }
+
+    /// calls the method associated with the collector and returns the resulting data
+    pub(crate) fn call_method(&mut self) -> Result<SensorData, DataCollectionError> {
+        let requested_dt = Local::now().fixed_offset();
```
acmorrow, line 169: This `Local::now` is duplicated work in the `Sensor::get_readings_data` case

acmorrow (commented):
Looking pretty good, with a design question.

micro-rdk/src/common/sensor.rs
gvaradarajan: Originally my plan was to put this function in `DataCollector` rather than in this trait/ Then I could create `From<_> for SensorData` impls for both Vector3 and GenericSensorReadings that initially set `metadata` to None and leave it to the `DataCollector` to determine `metadata`. But the downside you mentioned came to mind and I decided against it. 

However, if you think the other design is better enough to be worth that risk, I'm happy to do that instead

micro-rdk/src/common/data_collector.rs
gvaradarajan: This was a mistake. It was only a problem if one moved "data" into the default features and ran `make clippy-esp32`. I'll remove

micro-rdk/src/common/sensor.rs
acmorrow: I'm leaning towards making the change. After all, the goal here is for the sensor to provide the information that goes into the `.data` `oneof`. It seems fair to conclude that the sensor shouldn't need to be involved in deciding any metadata that the data sync service might need in addition to the sensors own readings.

Also, I wonder if since `sensor_data::Data` is a real type, it wouldn't make more sense to create `From<_> for sensor_data::Data`? Then the metadata isn't even part of the story and there is no need to add any special handling around `None` for the `metadata` field.

acmorrow (approved):
This LGTM if you want to keep the current design, or, we can do another round if you decide to follow through on the alternative design.

micro-rdk/src/common/sensor.rs
gvaradarajan: yeah, it's better to write conversions to `sensor_data::Data`. Ok, I'll make the change

acmorrow (approved):
LGTM

micro-rdk/Cargo.toml
```diff
@@ -15,7 +15,7 @@ links = "micro_rdk"
 crate-type = ["lib"]
 
 [features]
-default = ["builtin-components"]
+default = ["builtin-components", "data"]
```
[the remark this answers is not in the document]
gvaradarajan: yep lol

PR #166, opened 2024-03-21
Remove anyhow from Status

npmenard:


micro-rdk/src/common/status.rs
```diff
@@ -2,15 +2,25 @@ use std::sync::{Arc, Mutex};
...
 use crate::google;
 
+use thiserror::Error;
+
+use super::encoder::EncoderError;
+
+#[derive(Error, Debug)]
+pub enum StatusError {
+    #[error(transparent)]
+    EncoderError(#[from] EncoderError),
```
acmorrow, line 12: Where does `EncoderError` get used with `StatusError`?

acmorrow: Oh, I think I see. From here: https://github.com/viamrobotics/micro-rdk/pull/166/files#diff-53045a0f819a0cfcd264fae4eeea7798314b6efd9572e2fcaf8027ab0593a9acR91

acmorrow (approved):
LGTM

PR #167, opened 2024-03-25
RSDK-7063 - implement data store

gvaradarajan:
~~PR 1 of 2: This PR only implements the data store and its write functionality (reading returns an unimplemented error)~~. Added read functionality as well

For now, to run the tests introduced in the PR, add "data" to the default features in micro-rdk/Cargo.toml before calling make test.

### DIFF
--- a/micro-rdk/Cargo.toml
+++ b/micro-rdk/Cargo.toml
@@ -75,7 +75,7 @@ socket2.workspace = true
 stun_codec.workspace = true
 thiserror.workspace = true
 trackable.workspace = true
-
+ringbuf.workspace = true
 
 [build-dependencies]
 const-gen.workspace = true
--- a/micro-rdk/src/common/data_store.rs
+++ b/micro-rdk/src/common/data_store.rs
@@ -1,0 +1,526 @@
+//! Contains the DataStore trait and a usable StaticMemoryDataStore.
+//! Implementers of the trait are meant to be written to by DataCollectors (RSDK-6992, RSDK-6994)
+//! and read from by a task that uploads the data to app (RSDK-6995)
+
+use crate::proto::app::data_sync::v1::SensorData;
+use bytes::{Buf, BufMut, BytesMut};
+use prost::{encoding::decode_varint, length_delimiter_len, DecodeError, EncodeError, Message};
+use ringbuf::{ring_buffer::RbBase, Consumer, LocalRb, Producer};
+use std::{
+    mem::MaybeUninit,
+    sync::atomic::{AtomicBool, Ordering},
+};
+use thiserror::Error;
+
+use super::data_collector::ResourceMethodKey;
+
+#[derive(Debug, Clone, Copy)]
+pub enum WriteMode {
+    PreserveOrFail,
+    OverwriteOldest,
+}
+
+impl Default for WriteMode {
+    fn default() -> Self {
+        Self::PreserveOrFail
+    }
+}
+
+static mut DATA_STORE: [MaybeUninit<u8>; 1024000] = [MaybeUninit::uninit(); 1024000];
+
+#[derive(Error, Debug)]
+pub enum DataStoreError {
+    #[error(transparent)]
+    EncodingError(#[from] EncodeError),
+    #[error("SensorDataTooLarge")]
+    DataTooLarge,
+    #[error("Store already initialized")]
+    DataStoreInitialized,
+    #[error("Data write failure")]
+    DataWriteFailure,
+    #[error("Buffer full")]
+    DataBufferFull(ResourceMethodKey, SensorData),
+    #[error("Current message is malformed")]
+    DataIntegrityError,
+    #[error("Unknown collector key: {0}")]
+    UnknownCollectorKey(ResourceMethodKey),
+    #[error(transparent)]
+    DecodeError(#[from] DecodeError),
+    #[error("unimplemented")]
+    Unimplemented,
+}
+
+lazy_static::lazy_static! {
+    static ref DATA_STORE_INITIALIZED: AtomicBool = AtomicBool::new(false);
+}
+
+pub trait DataStore {
+    /// Store the sensor data message in a region specified by the ResourceMethodKey. To overwrite
+    /// the oldest messages if necessary, pass true for `overwrite_old_data`
+    fn write_message(
+        &mut self,
+        collector_key: &ResourceMethodKey,
+        message: SensorData,
+        write_mode: WriteMode,
+    ) -> Result<(), DataStoreError>;
+
+    /// Reads the next available message in the store for the given ResourceMethodKey. It should return
+    /// an empty BytesMut with 0 capacity when there are no available messages left.
+    fn read_next_message(
+        &mut self,
+        collector_key: &ResourceMethodKey,
+    ) -> Result<BytesMut, DataStoreError>;
+}
+
+/// StaticMemoryDataStore is an entity that governs the static bytes memory
+/// reserved in DATA_STORE. The memory is segmented based according to the DataCollectors expected
+/// (identified by collector keys) and each segment view is treated as a separate ring buffer of SensorData
+/// messages. Currently, an equal amount of space is alloted to each collector, which will affect
+/// the maximum allowed size of a single message (computed as the length of DATA_STORE divided by
+/// the number of collector keys). It should be treated as a global struct that should only be initialized once
+/// and is not thread-safe (all interactions should be blocking).
+pub struct StaticMemoryDataStore {
+    buffers: Vec<LocalRb<u8, &'static mut [MaybeUninit<u8>]>>,
+    collector_keys: Vec<ResourceMethodKey>,
+}
+
+impl StaticMemoryDataStore {
+    pub fn new(collector_keys: Vec<ResourceMethodKey>) -> Result<Self, DataStoreError> {
+        if !DATA_STORE_INITIALIZED.fetch_or(true, Ordering::SeqCst) {
+            let len_per_buffer = unsafe { DATA_STORE.len() } / collector_keys.len();
+            let mut buffers = Vec::new();
+            for i in 0..(collector_keys.len()) {
+                let start_idx = i * len_per_buffer;
+                let end_idx = (i + 1) * len_per_buffer;
+                unsafe {
+                    buffers.push(LocalRb::from_raw_parts(
+                        &mut DATA_STORE[start_idx..end_idx],
+                        0,
+                        0,
+                    ));
+                }
+            }
+            return Ok(Self {
+                buffers,
+                collector_keys,
+            });
+        }
+        Err(DataStoreError::DataStoreInitialized)
+    }
+
+    fn get_index_for_collector(
+        &self,
+        collector_key: &ResourceMethodKey,
+    ) -> Result<usize, DataStoreError> {
+        self.collector_keys
+            .iter()
+            .position(|key| key == collector_key)
+            .ok_or(DataStoreError::UnknownCollectorKey(collector_key.clone()))
+    }
+}
+
+impl DataStore for StaticMemoryDataStore {
+    fn write_message(
+        &mut self,
+        collector_key: &ResourceMethodKey,
+        message: SensorData,
+        write_mode: WriteMode,
+    ) -> Result<(), DataStoreError> {
+        let buffer_index = self.get_index_for_collector(collector_key)?;
+        let buffer = &self.buffers[buffer_index];
+        let encode_len = message.encoded_len();
+        let total_encode_len = length_delimiter_len(encode_len) + encode_len;
+
+        while total_encode_len > buffer.vacant_len() {
+            if !matches!(write_mode, WriteMode::OverwriteOldest) {
+                return Err(DataStoreError::DataBufferFull(
+                    collector_key.clone(),
+                    message,
+                ));
+            }
+            let mut cons = unsafe { Consumer::new(buffer) };
+            let (left, right) = cons.as_slices();
+            let mut chained = Buf::chain(left, right);
+            let encoded_len = decode_varint(&mut chained)? as usize;
+
+            let advance = length_delimiter_len(encoded_len);
+            unsafe { cons.advance(advance) };
+            cons.skip(encoded_len);
+        }
+        unsafe {
+            let mut prod = Producer::new(buffer);
+            let (left, right) = prod.free_space_as_slices();
+            let mut chained = BufMut::chain_mut(left, right);
+            message.encode_length_delimited(&mut chained)?;
+            prod.advance(total_encode_len);
+        }
+
+        Ok(())
+    }
+    fn read_next_message(
+        &mut self,
+        collector_key: &ResourceMethodKey,
+    ) -> Result<BytesMut, DataStoreError> {
+        let buffer_index = self.get_index_for_collector(collector_key)?;
+        let buffer = &self.buffers[buffer_index];
+        if buffer.is_empty() {
+            return Ok(BytesMut::with_capacity(0));
+        }
+
+        let mut cons = unsafe { Consumer::new(buffer) };
+        let (left, right) = cons.as_slices();
+        let mut chained = Buf::chain(left, right);
+        let encoded_len = decode_varint(&mut chained)? as usize;
+        let advance = length_delimiter_len(encoded_len);
+        unsafe { cons.advance(advance) };
+
+        let mut msg_bytes = BytesMut::with_capacity(encoded_len);
+        unsafe {
+            msg_bytes.set_len(encoded_len);
+        }
+        cons.pop_slice(&mut msg_bytes);
+        Ok(msg_bytes)
+    }
+}
+
+impl Drop for StaticMemoryDataStore {
+    fn drop(&mut self) {
+        DATA_STORE_INITIALIZED.store(false, Ordering::SeqCst);
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::collections::HashMap;
+
+    use crate::common::data_collector::{CollectionMethod, ResourceMethodKey};
+    use crate::common::data_store::DataStore;
+    use crate::common::data_store::DataStoreError;
+    use crate::common::data_store::WriteMode;
+    use crate::common::data_store::DATA_STORE;
+    use crate::google::protobuf::Timestamp;
+    use crate::google::protobuf::{value::Kind, Struct, Value};
+    use crate::proto::app::data_sync::v1::sensor_data::Data;
+    use crate::proto::app::data_sync::v1::{SensorData, SensorMetadata};
+    use prost::{length_delimiter_len, Message};
+
+    #[test_log::test]
+    fn test_data_store() {
+        let reading_requested_dt = chrono::offset::Local::now().fixed_offset();
+
+        let empty_message = SensorData {
+            metadata: None,
+            data: None,
+        };
+        let thing_key = ResourceMethodKey {
+            r_name: "thing".to_string(),
+            component_type: "rdk::component::sensor".to_string(),
+            method: CollectionMethod::Readings,
+        };
+        let empty_message_2 = SensorData {
+            metadata: None,
+            data: Some(Data::Struct(Struct {
+                fields: HashMap::new(),
+            })),
+        };
+        let thing_2_key = ResourceMethodKey {
+            r_name: "thing".to_string(),
+            component_type: "rdk::component::movement_sensor".to_string(),
+            method: CollectionMethod::Readings,
+        };
+        let data_message_no_metadata = SensorData {
+            metadata: None,
+            data: Some(Data::Struct(Struct {
+                fields: HashMap::from([
+                    (
+                        "thing_1".to_string(),
+                        Value {
+                            kind: Some(Kind::NumberValue(245.01)),
+                        },
+                    ),
+                    (
+                        "thing_2".to_string(),
+                        Value {
+                            kind: Some(Kind::BoolValue(true)),
+                        },
+                    ),
+                ]),
+            })),
+        };
+        let reading_received_dt = chrono::offset::Local::now().fixed_offset();
+        let data_message = SensorData {
+            metadata: Some(SensorMetadata {
+                time_requested: Some(Timestamp {
+                    seconds: reading_requested_dt.timestamp(),
+                    nanos: reading_requested_dt.timestamp_subsec_nanos() as i32,
+                }),
+                time_received: Some(Timestamp {
+                    seconds: reading_received_dt.timestamp(),
+                    nanos: reading_received_dt.timestamp_subsec_nanos() as i32,
+                }),
+            }),
+            data: Some(Data::Struct(Struct {
+                fields: HashMap::from([
+                    (
+                        "thing_1".to_string(),
+                        Value {
+                            kind: Some(Kind::NumberValue(245.01)),
+                        },
+                    ),
+                    (
+                        "thing_2".to_string(),
+                        Value {
+                            kind: Some(Kind::BoolValue(true)),
+                        },
+                    ),
+                ]),
+            })),
+        };
+        let collector_keys = vec![thing_key.clone(), thing_2_key.clone()];
+        let store = super::StaticMemoryDataStore::new(collector_keys);
+        assert!(store.is_ok());
+        let mut store = store.unwrap();
+
+        let res = store.write_message(&thing_key, empty_message, Default::default());
+        assert!(res.is_ok());
+        let res = store.write_message(&thing_key, data_message, Default::default());
+        assert!(res.is_ok());
+
+        let res = store.write_message(&thing_2_key, empty_message_2, Default::default());
+        assert!(res.is_ok());
+        let res = store.write_message(&thing_2_key, data_message_no_metadata, Default::default());
+        assert!(res.is_ok());
+
+        let read_message = store.read_next_message(&thing_key);
+        assert!(read_message.is_ok());
+        let mut read_message = read_message.unwrap();
+        let read_message = SensorData::decode(&mut read_message);
+        assert!(read_message.is_ok());
+        let read_message = read_message.unwrap();
+        let expected_msg = SensorData {
+            metadata: None,
+            data: None,
+        };
+        assert_eq!(read_message, expected_msg);
+
+        let read_message = store.read_next_message(&thing_key);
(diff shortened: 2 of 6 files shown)


micro-rdk/Cargo.toml
```diff
@@ -15,13 +15,15 @@ links = "micro_rdk"
 crate-type = ["lib"]
 
 [features]
-default = ["builtin-components"]
+default = ["builtin-components", "data"]
```
gvaradarajan, line 18: rm before merge

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,448 @@
...
+        let mut res = Vec::new();
+        let mut return_remaining = false;
+        for req in messages {
+            if return_remaining {
+                res.push(req);
+                continue;
+            }
+            let encode_len = req.encoded_len();
+            if encode_len > max_msg_size {
+                return Err(DataStoreError::DataTooLarge);
```
acmorrow, line 138: This is sort of the scenario I was mentioning above: if you return this error, you may have already written some of the SensorData from `messages`, but you don't get the benefit of the `Vec` return.

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,448 @@
...
+            buffer
+                .push(0)
+                .map_err(|_| DataStoreError::DataWriteFailure)?;
+            let len_bytes = (encode_len as u32).to_be_bytes();
+            buffer.push_slice(&len_bytes);
+
+            let mut buf = BytesMut::with_capacity(req.encoded_len());
+            req.encode(&mut buf)?;
+            let mut buf_iter = buf.into_iter();
+            buffer.push_iter(&mut buf_iter);
```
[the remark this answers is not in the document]
gvaradarajan: no the normal push methods already do that, the overwrite methods overwrite the _latest_ item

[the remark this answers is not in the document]
acmorrow: Ah, I read the docs too quickly... sorry.

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,448 @@
...
+            }
+            if encode_len + 5 > buffer.vacant_len() {
+                return_remaining = true;
+                res.push(req);
+                continue;
+            }
+            buffer
+                .push(0)
+                .map_err(|_| DataStoreError::DataWriteFailure)?;
+            let len_bytes = (encode_len as u32).to_be_bytes();
```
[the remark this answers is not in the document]
gvaradarajan: I did this (the zero-byte and the length in big-endian) out of consistency with the logic `GrpcServer::encode_message`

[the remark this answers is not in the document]
acmorrow: OK I see. On the other hand, it doesn't look like that encoding commonality is actually leveraged, or at least isn't today, since in `read_messages` you actually pop the zero and the encoded len, then return only the encoded bytes. But if you would like to keep it because you anticipate making use of it in the future, I don't mind. Though I think a comment would be good.

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,448 @@
...
+
+/// StaticMemoryDataStore is an entity that governs the static bytes memory
+/// reserved in DATA_STORE. The memory is segmented based according to the DataCollectors expected
+/// (identified by collector keys) and each segment view is treated as a separate ring buffer of SensorData
+/// messages. Currently, an equal amount of space is alloted to each collector, which will affect
+/// the maximum allowed size of a single message (see get_maximum_message_size below)
+/// It should be treated as a global struct that should only be initialized once and is not
+/// thread-safe (all interactions should be blocking).
+pub struct StaticMemoryDataStore {
+    buffers: Vec<LocalRb<u8, &'static mut [MaybeUninit<u8>]>>,
```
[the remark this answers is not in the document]
gvaradarajan: I'm expecting single-threaded access, which is why I've gone with LocalRb

npmenard (commented):
Looks pretty good!! Left a couple of comments

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,512 @@
...
+    Unimplemented,
+}
+
+pub enum StoreMessage {
+    Message(BytesMut),
+    EndOfBuffer,
+}
+
+lazy_static::lazy_static! {
+    static ref DATA_STORE_INITIALIZED: Arc<AtomicBool> = Arc::new(AtomicBool::new(false));
```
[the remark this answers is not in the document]
gvaradarajan: Nope, did it out of habit

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,512 @@
...
+        let buffer = &mut self.buffers[buffer_index];
+        let encode_len = message.encoded_len();
+        if encode_len > max_msg_size {
+            return Err(DataStoreError::DataTooLarge);
+        }
+        if encode_len + 5 > buffer.vacant_len() {
+            return Err(DataStoreError::DataBufferFull(collector_key, message));
+        }
+        buffer
+            .push(0)
```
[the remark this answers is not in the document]
gvaradarajan: yeah that was my plan for determining the start of a new message. The logic is in `read_next_message`

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,512 @@
...
+            .map_err(|_| DataStoreError::DataWriteFailure)?;
+        let len_bytes = (encode_len as u32).to_be_bytes();
```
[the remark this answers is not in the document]
gvaradarajan: because that's the way we did it in `grpc.rs`. @acmorrow mentioned this as well. I understand casting is more efficient with little-endian though, so if you both feel that performance increase is worth it I'm happy to change

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,512 @@
...
+    fn read_next_message(
+        &mut self,
+        collector_key: ResourceMethodKey,
+    ) -> Result<StoreMessage, DataStoreError> {
+        let buffer_index = self.get_index_for_collector(&collector_key)?;
+        let buffer = &mut self.buffers[buffer_index];
+
+        if let Some(&&zero_byte) = buffer.iter().peekable().peek() {
+            if zero_byte != 0 {
+                return Err(DataStoreError::DataIntegrityError);
```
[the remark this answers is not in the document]
gvaradarajan: I believe the only remaining recourse it to clear the store and start from scratch (losing the existing data as a side effect), but that doesn't feel like the responsibility of this function

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,512 @@
...
+        buffer.pop_slice(msg_vec.as_mut_slice());
+        let mut msg_bytes = BytesMut::with_capacity(encoded_len);
+        msg_bytes.put(Cursor::new(msg_vec));
+        Ok(StoreMessage::Message(msg_bytes))
+    }
+    fn clear(&mut self) {
+        for buffer in self.buffers.iter_mut() {
+            buffer.clear();
+            unsafe {
+                buffer.set_head(0);
```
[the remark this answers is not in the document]
gvaradarajan: I guess there's no reasonable scenario now. Originally I thought I could make the data survive restart, so I figured it would be necessary to have a way to clear the store if, for example, there was a breaking proto change that would make uploading the existing data impossible.

micro-rdk/src/common/data_store.rs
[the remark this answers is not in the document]
gvaradarajan: This is a good point actually, I should have looked more carefully. I'll change to little-endian actually

micro-rdk/src/common/data_store.rs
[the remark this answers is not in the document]
gvaradarajan: changing to little-endian

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,448 @@
...
+            if encode_len > max_msg_size {
+                return Err(DataStoreError::DataTooLarge);
+            }
+            if encode_len + 5 > buffer.vacant_len() {
+                return_remaining = true;
+                res.push(req);
+                continue;
+            }
+            buffer
+                .push(0)
```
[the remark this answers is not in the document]
gvaradarajan: This signifies the start of a message in the ringbuffer

micro-rdk/src/common/data_store.rs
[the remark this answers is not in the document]
npmenard: Then use 0xFF or some other "magic" bytes, 0 is too common since you are storing size immediately afterwards

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,475 @@
...
+        unsafe {
+            let mut prod = Producer::new(buffer as &LocalRb<u8, &'static mut [MaybeUninit<u8>]>);
+            let (left, right) = prod.free_space_as_slices();
+            let mut chained = BufMut::chain_mut(left, right);
+            message.encode(&mut chained)?;
+            prod.advance(encode_len);
+        }
+        Ok(())
+    }
+    fn read_next_message(
```
acmorrow, line 141: Similar general thoughts about whether you can use `Consume` traits for `[u8]`, and keep an array of `Consume` objects for each section.

micro-rdk/src/common/data_store.rs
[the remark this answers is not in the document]
gvaradarajan: I agree that a checksum safeguards against a wider variety of causes rather than a simple overwrite bug in store_message, but is this not technically more space efficient?

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,475 @@
...
+            return Err(DataStoreError::DataBufferFull(collector_key, message));
+        }
+        buffer
+            .push(0)
+            .map_err(|_| DataStoreError::DataWriteFailure)?;
+        let len_bytes = (encode_len as u32).to_le_bytes();
+        buffer.push_slice(&len_bytes);
+
+        unsafe {
+            let mut prod = Producer::new(buffer as &LocalRb<u8, &'static mut [MaybeUninit<u8>]>);
```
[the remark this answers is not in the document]
gvaradarajan: - I've perhaps misunderstood, but how would I use write_all to prevent a separate allocation that I would have to copy the bytes into? This was why I didn't use `buffer.push_slice()`
- I find it maybe equally awkward to store something on the struct that contains a reference to something else on the struct and the pattern of creating a Producer on the fly is one that the ringbuf library uses internally in implementing `Rb`

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,475 @@
...
+                buffer.set_head(0);
+                buffer.set_tail(0);
+            }
+        }
+    }
+}
+
+impl Drop for StaticMemoryDataStore {
+    fn drop(&mut self) {
+        self.clear();
```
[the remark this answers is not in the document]
gvaradarajan: I can't remember why I did this, so I guess not. Will remove.

micro-rdk/src/common/data_store.rs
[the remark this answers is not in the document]
acmorrow: You are right, I got myself tangled up in the traits. It's unfortunate, because Prost's `encode` only operates on `BytesMut`, and `Producer` can write to an `io::Writer` and `BytesMut` offers `.writer` to produce a `Writer`, but as far as I can see you can't get a `Producer` to appear as a `BytesMut` such that you can give it to `Prost::encode` nor can you get `Prost::encode` to write to an `io::Writer` which would be another way, one of which was what I was hoping for and convinced myself worked.

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,480 @@
...
+            .push(0)
+            .map_err(|_| DataStoreError::DataWriteFailure)?;
+        let len_bytes = (encode_len as u32).to_le_bytes();
+        buffer.push_slice(&len_bytes);
+
+        unsafe {
+            let mut prod = Producer::new(buffer as &LocalRb<u8, &'static mut [MaybeUninit<u8>]>);
+            let (left, right) = prod.free_space_as_slices();
+            let mut chained = BufMut::chain_mut(left, right);
+            message.encode(&mut chained)?;
```
acmorrow, line 134: While looking at the Prost Message regarding the above, I noticed that it has an `encode_length_delimited` and `decode_length_delimited` pair. Maybe that could be used instead of manually encoding the length above?

micro-rdk/src/common/data_store.rs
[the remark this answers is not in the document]
acmorrow: Mostly I'm suggesting that the delimiter isn't worth it. If we have properly length encoded, when will we encounter a missing / broken one outside of memory corruption?

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,480 @@
...
+                return Err(DataStoreError::DataIntegrityError);
+            }
+            let _ = buffer.pop();
+        } else {
+            return Ok(BytesMut::with_capacity(0));
+        }
+        let mut encoded_len: [u8; 4] = [0; 4];
+        buffer.pop_slice(&mut encoded_len);
+        let encoded_len = u32::from_le_bytes(encoded_len) as usize;
+        if encoded_len > unsafe { DATA_STORE.len() } {
```
acmorrow, line 158: I think the bound can be tighter: it can't be longer than any one of the individual ring buffers, right? Though, if you take my suggestion above to not manually encode the length and use the `*_length_delimited` methods this may no longer be possible.

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,480 @@
...
+        &mut self,
+        collector_key: ResourceMethodKey,
+        message: SensorData,
+    ) -> Result<(), DataStoreError> {
+        let buffer_index = self.get_index_for_collector(&collector_key)?;
+        let buffer = &mut self.buffers[buffer_index];
+        let encode_len = message.encoded_len();
+        if encode_len + 5 > buffer.vacant_len() {
+            return Err(DataStoreError::DataBufferFull(collector_key, message));
+        }
```
acmorrow, line 123: Doesn't this effectively prevent overwriting old data, which I thought was the behavior we wanted? Otherwise, if we are offline and not pulling data out with `read_next_message`, we will only retain the oldest data and discard newer. Note that this will require some care, since you can't just overwrite, you need to ensure that once overwritten, that the new head points to the beginning of an encoded message.

acmorrow (changes requested):

micro-rdk/src/common/data_store.rs
gvaradarajan: So the reason that I didn't use those methods is because they encode the length as a `u64` which felt unnecessarily space inefficient. If you feel the trade-off for the simplicity is worth it though, I can change

micro-rdk/src/common/data_store.rs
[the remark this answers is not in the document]
gvaradarajan: ok, I'll remove

micro-rdk/src/common/data_store.rs
acmorrow: I believe it encodes as varint which could actually do better than u32 for small messages:
- https://docs.rs/prost/latest/src/prost/message.rs.html#89
- https://protobuf.dev/programming-guides/encoding/#varints
You would need to use `encoded_len_varint` too, I think, to know how many total bytes were required.

micro-rdk/src/common/data_store.rs
acmorrow: Wouldn't that result in dropping old data when offline even in the case where we had plenty of space to buffer instead?

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,482 @@
...
+impl DataStore for StaticMemoryDataStore {
+    fn store_message(
+        &mut self,
+        collector_key: ResourceMethodKey,
+        message: SensorData,
+    ) -> Result<(), DataStoreError> {
+        let buffer_index = self.get_index_for_collector(&collector_key)?;
+        let buffer = &mut self.buffers[buffer_index];
+        let encode_len = message.encoded_len();
+        if encode_len + 4 > buffer.vacant_len() {
```
acmorrow, line 121: Re-posting, since I think it got hidden away as outdated:

> Wouldn't that result in dropping old data when offline even in the case where we had plenty of space to buffer instead?

gvaradarajan: I can't seem to find the conversation that this was a part of, why does this happen again?

micro-rdk/src/common/data_store.rs
gvaradarajan: that's fair, but can't we then simply write it back to the store?

acmorrow: It feels to me like there is a separation of concerns issue with that approach, because data retention / eviction logic and policy ends up mixed between the `DataStore` and the higher layer. Whereas if the `DataStore` just continually overwrites old with new when necessary, the higher layer doesn't need to know at all, I think.

gvaradarajan: ok, so I think what I'll do in this case then is simply consume messages until the current message will fit

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,514 @@
...
+        let encode_len = message.encoded_len();
+        let total_encode_len = length_delimiter_len(encode_len) + encode_len;
+        if total_encode_len > buffer.vacant_len() {
+            if overwrite_old_data {
+                let mut cons = unsafe {
+                    Consumer::new(buffer as &LocalRb<u8, &'static mut [MaybeUninit<u8>]>)
+                };
+                // we'll skip messages until we can fit the current one
+                while total_encode_len > buffer.vacant_len() {
+                    let encoded_len = leb128::read::unsigned(&mut cons)? as usize;
```
acmorrow: I'd really like to somehow use https://docs.rs/prost/0.11.0/prost/fn.decode_length_delimiter.html here instead, so we don't need to pull in `leb128`, but I haven't quite managed to figure out how to do it. Can you add a TODO? I'd prefer to use prost on both sides of encoding and decoding.

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,514 @@
...
+            let mut chained = BufMut::chain_mut(left, right);
+            message.encode_length_delimited(&mut chained)?;
+            prod.advance(total_encode_len);
+        }
+
+        Ok(())
+    }
+    fn read_next_message(
+        &mut self,
+        collector_key: ResourceMethodKey,
```
acmorrow, line 158: Ditto question re `&` for `collector_key`

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,514 @@
...
+            .position(|key| key == collector_key)
+            .ok_or(DataStoreError::UnknownCollectorKey(collector_key.clone()))
+    }
+}
+
+impl DataStore for StaticMemoryDataStore {
+    fn store_message(
+        &mut self,
+        collector_key: ResourceMethodKey,
+        message: SensorData,
```
acmorrow, line 120: Same question re `message`.

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,514 @@
...
+
+        Ok(())
+    }
+    fn read_next_message(
+        &mut self,
+        collector_key: ResourceMethodKey,
+    ) -> Result<BytesMut, DataStoreError> {
+        let buffer_index = self.get_index_for_collector(&collector_key)?;
+        let buffer = &mut self.buffers[buffer_index];
+        if buffer.iter().peekable().peek().is_none() {
```
acmorrow, line 162: Could this just be `len`?

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,514 @@
...
+        overwrite_old_data: bool,
+    ) -> Result<(), DataStoreError> {
+        let buffer_index = self.get_index_for_collector(&collector_key)?;
+        let buffer = &mut self.buffers[buffer_index];
+        let encode_len = message.encoded_len();
+        let total_encode_len = length_delimiter_len(encode_len) + encode_len;
+        if total_encode_len > buffer.vacant_len() {
+            if overwrite_old_data {
+                let mut cons = unsafe {
+                    Consumer::new(buffer as &LocalRb<u8, &'static mut [MaybeUninit<u8>]>)
```
acmorrow, line 130: Why is the cast needed? Asking mostly for myself.

micro-rdk/src/common/data_store.rs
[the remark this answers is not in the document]
acmorrow: I think we should remove the method unless/until we need it.

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,514 @@
...
+    fn store_message(
+        &mut self,
+        collector_key: ResourceMethodKey,
+        message: SensorData,
+        overwrite_old_data: bool,
+    ) -> Result<(), DataStoreError>;
+
+    /// Reads the next available message in the store for the given ResourceMethodKey. It should return
+    /// an empty BytesMut with 0 capacity when there are no available messages left.
+    fn read_next_message(
```
acmorrow, line 60: Can we make the verb part of this method name pair better with `store_message` or its renaming? `store/retrieve`, `push/pull`, `read/write`.

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,514 @@
...
+        collector_key: ResourceMethodKey,
+    ) -> Result<BytesMut, DataStoreError> {
+        let buffer_index = self.get_index_for_collector(&collector_key)?;
+        let buffer = &mut self.buffers[buffer_index];
+        if buffer.iter().peekable().peek().is_none() {
+            return Ok(BytesMut::with_capacity(0));
+        }
+        let mut cons =
+            unsafe { Consumer::new(buffer as &LocalRb<u8, &'static mut [MaybeUninit<u8>]>) };
+        let encoded_len = leb128::read::unsigned(&mut cons)? as usize;
```
acmorrow: Ditto TODO re eliminating `leb128` in favor of `prost`

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,514 @@
...
+        let mut msg_bytes = BytesMut::with_capacity(encoded_len);
+        unsafe {
+            msg_bytes.set_len(encoded_len);
+        }
+        cons.pop_slice(&mut msg_bytes);
```
acmorrow, line 178: Is there a way to get some sort of Writer on BytesMut such that the `unsafe set_len` isn't needed?

acmorrow (commented):
Overall this looks good. Comments here are more or less all polish type things.

micro-rdk/src/common/data_store.rs
gvaradarajan: yeah it could

micro-rdk/src/common/data_store.rs
gvaradarajan: I did intend to pass ownership of the message completely to the function but I guess it's more flexible to pass by reference

micro-rdk/src/common/data_store.rs
```diff
@@ -0,0 +1,514 @@
...
+}
+
+pub trait DataStore {
+    /// Store the sensor data message in a region specified by the ResourceMethodKey. To overwrite
+    /// the oldest messages if necessary, pass true for `overwrite_old_data`
+    fn store_message(
+        &mut self,
+        collector_key: ResourceMethodKey,
+        message: SensorData,
+        overwrite_old_data: bool,
```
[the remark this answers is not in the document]
gvaradarajan: I'm not sure I understand, with the boolean you can still use `Default::default()`, what does the enum do here?

micro-rdk/src/common/data_store.rs
gvaradarajan: it's not now, I previously needed it because I needed `buffer` to be `&mut`, but now that I'm using Producers and Consumers I don't

micro-rdk/src/common/data_store.rs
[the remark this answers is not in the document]
acmorrow: Using an enum makes it (hopefully) clearer at the call site what is happening without needing to consult the implementation:

```
thing.write_message(key, data, false);  // What does the `false` mean???
```

vs

```
thing.write_message(key, data, RetentionPolicy::PreserveOrFail);   // Will error if no space
thing.write_message(key, data, RetentionPolicy::Overwrite);  // Will overwrite
thing.write_message(key, data, Default::default());  // 
```

Alternatively:

```
thing.write_message(key, data );  // Default behavior is to preserve
thing.write_message_overwrite(key ,data);  // Named method to opt-in to overwrite behavior
```

micro-rdk/src/common/data_store.rs
```diff
@@ -114,32 +114,36 @@ impl StaticMemoryDataStore {
...
+
+        while total_encode_len > buffer.vacant_len() {
+            if !overwrite_old_data {
+                return Err(DataStoreError::DataBufferFull(
+                    collector_key.clone(),
+                    message,
+                ));
             } else {
-                return Err(DataStoreError::DataBufferFull(collector_key, message));
+                let mut cons = unsafe { Consumer::new(buffer) };
```
acmorrow, line 129: Do you even need the `else`?

micro-rdk/src/common/data_store.rs
acmorrow: Fine whichever way you decide it, no need for an update.

acmorrow (approved):
LGTM

micro-rdk/src/common/data_store.rs
acmorrow: Still interested in this, btw, if you have ideas.

gvaradarajan: Couldn't find a way to do this that didn't involve allocating an extra copy

acmorrow: OK, thanks for checking.

acmorrow (approved):

PR #168, opened 2024-03-31
Add link to JIRA in README.md

zaporter:
Wanted to submit a bug / help request this weekend but there was no link to JIRA on the README.

This is a copy-over from the RDK.

I'd **much rather** submit a github issue so enabling issues would also work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,8 @@ the ESP32.
 - Discord: <https://discord.gg/viam>
 - Support: <https://support.viam.com>
 
+If you have a bug or an idea, please open an issue in our [JIRA project](https://viam.atlassian.net/).
+
 ## (In)stability Notice
 
 > **Warning**


npmenard (commented):
Can you create issues as a non Viam user? I tried with my personal address but couldn't.
Shouldn't user use https://viam.atlassian.net/servicedesk/customer/portals now?

zaporter:
> I tried with my personal address but couldn't.

Same -- for some reason I remember someone saying that people could

> Shouldn't user use https://viam.atlassian.net/servicedesk/customer/portals now?

I have no idea -- that seems like a non-public issue submission. If I wasn't a viam-employee, I don't think I would ever submit issues to that. I would assume that is for gui issues like app.viam, not issues with the micro-rdk internals and throw my hands up in frustration and walk away.

npmenard (approved):
Well let's merge this anyway so that we are on parity with RDK, I'll try to chat with Steve about what we can do to make issue reporting more streamlined

PR #169, opened 2024-03-31
Add canon.yaml to project template

zaporter:
Docs give instructions to use canon but without a canon.yaml it uses debian:latest and fails.

https://docs.viam.com/get-started/installation/prepare/microcontrollers/development-setup/


npmenard (approved):
LGTM

PR #175, opened 2024-04-01
RSDK-6916

npmenard:
This is the final PR removing anyhow from micro-rdk, examples and templates


micro-rdk/src/native/entry.rs
```diff
@@ -69,13 +69,14 @@ pub async fn serve_web_inner(
...
-                            let logs = vec![config_log_entry(datetime, Some(&err))];
+                            let logs = vec![config_log_entry(datetime, Some(err))];
                             client
                                 .push_logs(logs)
                                 .await
                                 .expect("could not push logs to app");
                         }
-                        panic!("{}", err)
+                        //TODO shouldn't panic here
+                        panic!("couldn't build robot");
```
gvaradarajan, line 79: do we not want to print the error in the panic message?

gvaradarajan (approved):
LGTM

micro-rdk/src/native/entry.rs
npmenard: No it's more a a placeholder for when we will implement offline mode or try building again

PR #178, opened 2024-04-08
Replace H2 with Hyper

npmenard:
Remove H2 for Hyper. This changes the way we handle executors, for now this approach is an intermediary step (eg : I don't want executor to be thread local like this)  We will need to rewrite this later.


micro-rdk/src/esp32/entry.rs
```diff
@@ -28,117 +28,106 @@ use super::{
...
-                        Ok(robot) => {
-                            if let Some(datetime) = cfg_received_datetime {
-                                let logs = vec![config_log_entry(datetime, None)];
-                                client
-                                    .push_logs(logs)
-                                    .await
-                                    .expect("could not push logs to app");
-                            }
-                            robot
+
```
acmorrow, line 46: GH diff didn't do a great job here, but from what I can tell there are no substantive changes here, right?

micro-rdk/src/esp32/entry.rs
```diff
@@ -164,7 +153,16 @@ pub fn serve_web(
     let exec = Esp32Executor::new();
     let cloned_exec = exec.clone();
 
-    let fut = cloned_exec.run(Box::pin(serve_web_inner(
+    cloned_exec
```
acmorrow, line 156: Neat! A few questions though:
- Is this the general shape of things for periodic tasks? For instance, data sync?
- Why do you need to `clone` the executor above after `new`ing it?

gvaradarajan (approved):
LGTM, one question about executors

micro-rdk/src/esp32/entry.rs
```diff
@@ -164,7 +153,16 @@ pub fn serve_web(
...
+        .spawn(async {
+            loop {
+                Timer::after(Duration::from_secs(150)).await;
+                unsafe { crate::esp32::esp_idf_svc::sys::esp_task_wdt_reset() };
+            }
+        })
+        .detach();
+
+    cloned_exec.run_forever(Box::pin(serve_web_inner(
```
acmorrow, line 165: Why is the `Box::pin` required?

micro-rdk/src/common/app_client.rs
```diff
@@ -169,18 +177,20 @@ pub(crate) struct AppSignaling(
...
 impl<'a> AppClient<'a> {
     pub(crate) async fn connect_signaling(&mut self) -> Result<AppSignaling, AppClientError> {
+        let (sender, receiver) = async_channel::bounded::<Bytes>(1);
         let r = self
             .grpc_client
             .build_request(
                 "/proto.rpc.webrtc.v1.SignalingService/Answer",
                 Some(&self.jwt),
                 &self.config.rpc_host,
+                BodyExt::boxed(StreamBody::new(receiver.map(|b| Ok(Frame::data(b))))),
```
acmorrow, line 187: I'm missing something here. Maybe two things:
- I can't find a `map` on the type that I think `receiver` is (edit: Gautham found it for me).
- I'm not clear on what the addition of this channel is doing or why

micro-rdk/src/esp32/exec.rs
```diff
@@ -2,55 +2,47 @@
...
     Future,
 };
-use std::rc::Rc;
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 /// This executor is local and bounded to the CPU that created it usually you would create it after spwaning a thread on a specific core
-pub struct Esp32Executor<'a> {
-    /// A local executor
-    executor: Rc<LocalExecutor<'a>>,
+pub struct Esp32Executor {}
```
acmorrow, line 10: Is the above comment accurate? I don't see anything in the docs for `LocalExecutor` that says it is bound to a specific core. It is bound to the creating thread.

micro-rdk/src/native/entry.rs
```diff
@@ -30,8 +28,9 @@ pub async fn serve_web_inner(
     tls_server_config: NativeTlsServerConfig,
     repr: RobotRepresentation,
     ip: Ipv4Addr,
-    exec: NativeExecutor<'_>,
+    exec: NativeExecutor,
 ) {
+    log::info!("web start");
```
acmorrow, line 33: Leftover debugging?, and on line 45 too.

micro-rdk/src/esp32/exec.rs
```diff
@@ -2,55 +2,47 @@
...
     /// Return a new executor bounded to the current core.
     pub fn new() -> Self {
-        Esp32Executor {
-            executor: Rc::new(LocalExecutor::new()),
-        }
-    }
-    /// Spawn a future onto the local executor
-    pub fn spawn<T: 'a>(&self, future: impl Future<Output = T> + 'a) -> Task<T> {
-        self.executor.spawn(future)
+        Self {}
```
acmorrow, line 19: All `Esp32Executor`s are now equivalent. Should this be `get` or similar now? Or `instance`? Are we planning to move to a world where there is a single executor?

micro-rdk/src/esp32/exec.rs
```diff
@@ -2,55 +2,47 @@
...
+    // Spawn a future onto the local executor
+    pub fn spawn<T: 'static>(&self, future: impl Future<Output = T> + 'static) -> Task<T> {
+        EX.with(|e| e.spawn(future))
     }
-}
 
-impl<'a> Default for Esp32Executor<'a> {
-    fn default() -> Self {
-        Self::new()
+    pub fn run_forever<T>(&self, future: impl Future<Output = T>) -> T {
```
acmorrow, line 26: This might be clearer as `block_on`

micro-rdk/src/native/exec.rs
```diff
@@ -1,58 +1,49 @@
...
-    future::{self},
+    future::{self, block_on},
     Future,
 };
-use std::rc::Rc;
 
 use crate::common::webrtc::exec::WebRtcExecutor;
 
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
```
acmorrow, line 10: All the same questions in this executor as in the esp32 one.

acmorrow (approved):
LGTM, lots of small questions and small nits /  suggestions, but nothing that I think needs to be acted on unless you are interested. Answers to some of the questions would be appreciated to help me advance my understanding of async rust!

micro-rdk/src/common/conn/server.rs
```diff
@@ -525,22 +524,18 @@ where
 }
 
 pin_project_lite::pin_project! {
-    struct WebRTCSignalingAnswerer<'a,'b, C,D,F> {
+    struct WebRTCSignalingAnswerer<'b, C,D,F> {
```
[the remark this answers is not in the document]
npmenard: nope

micro-rdk/src/common/conn/utils.rs
```diff
@@ -20,6 +20,38 @@ pub struct WebRtcNoOp {
     fp: Fingerprint,
 }
 
+impl rt::Read for WebRtcNoOp {
```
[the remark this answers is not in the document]
npmenard: yes they are intended to return Poll::Pending so the future can never resolve

micro-rdk/src/common/grpc_client.rs
```diff
@@ -112,7 +106,7 @@ where
         cx: &mut std::task::Context<'_>,
     ) -> std::task::Poll<Option<Self::Item>> {
         if self.buffer.is_empty() {
-            let chunk = match self.receiver_half.poll_data(cx) {
+            let chunk = match std::pin::Pin::new(&mut self.receiver_half).poll_frame(cx) {
```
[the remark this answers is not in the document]
npmenard: poll_frame requires self to be Pin.

micro-rdk/src/common/grpc_client.rs
```diff
@@ -149,45 +139,39 @@ where
     }
 }
 #[cfg(feature = "native")]
-type Executor<'a> = NativeExecutor<'a>;
+type Executor = NativeExecutor;
 #[cfg(feature = "esp32")]
-type Executor<'a> = Esp32Executor<'a>;
+type Executor = Esp32Executor;
```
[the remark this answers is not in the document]
npmenard: I'd like to keep them separated if we decide to sub the executor for esp32

micro-rdk/src/common/grpc_client.rs
```diff
@@ -220,33 +205,27 @@ impl<'a> GrpcClient<'a> {
...
+        r.body(body).map_err(GrpcClientError::HyperHttpError)
     }
 
     pub(crate) async fn send_request_bidi<R, P>(
         &mut self,
-        r: Request<()>,
-        message: Option<R>, // we shouldn't need this to get server headers when initiating a
-                            // bidi stream
+        r: Request<BoxBody<Bytes, hyper::Error>>,
+        sender: Sender<Bytes>, // we shouldn't need this to get server headers when initiating a
```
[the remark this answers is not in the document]
npmenard: nope

micro-rdk/src/common/grpc_client.rs
```diff
@@ -258,34 +237,27 @@ impl<'a> GrpcClient<'a> {
...
         &mut self,
-        r: Request<()>,
-        body: Bytes,
+        r: Request<BoxBody<Bytes, hyper::Error>>,
     ) -> Result<(Bytes, HeaderMap), GrpcClientError> {
-        let http2_connection = self.http2_connection.clone();
+        let mut http2_connection = self.http2_connection.clone();
         // verify if the server can accept a new HTTP2 stream
-        let mut http2_connection = http2_connection.ready().await?;
+        // TODO Error handling
```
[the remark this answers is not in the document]
npmenard: fixed here I forgot

micro-rdk/src/esp32/entry.rs
npmenard: no not really :)

micro-rdk/src/esp32/entry.rs
npmenard: we don't really need to that anymore has we can access it from thread_local_storage. I haven't found a definitive answer as too whether executors should be passed around of if they should be sourced from the current thread context
We don't really need to clone it's more to emphasize that it is the same underlying object

micro-rdk/src/esp32/entry.rs
npmenard: It will force the state machine to be allocated on the ram vs stack potentially

micro-rdk/src/esp32/exec.rs
npmenard: I removed it

micro-rdk/src/native/entry.rs
npmenard: :P

micro-rdk/src/esp32/exec.rs
npmenard: getting it will take us in lifetime nightmare :P

micro-rdk/src/esp32/exec.rs
npmenard: done

PR #181, opened 2024-04-12
start implementation of DataManager

gvaradarajan:
Creates a `DataManager` for use in data management. This implements data collection logic as well as the loop for running collections and data synchronization, but does not yet implement the actual synchronization logic (or the actual task for running data management), so the structure is likely to change. Moving forward it should receive its `DataCollectors` from the robot, and its `sync_interval_ms` from an instance of `RobotConfig`.

### DIFF
--- a/micro-rdk/src/common/data_collector.rs
+++ b/micro-rdk/src/common/data_collector.rs
@@ -1,4 +1,5 @@
 use std::fmt::Display;
+use std::time::Duration;
 
 use crate::google::protobuf::Timestamp;
 use crate::proto::app::data_sync::v1::{SensorData, SensorMetadata};
@@ -115,7 +116,7 @@ pub struct DataCollector {
     component_type: String,
     resource: ResourceType,
     method: CollectionMethod,
-    time_interval_ms: u64,
+    time_interval: Duration,
 }
 
 fn resource_method_pair_is_valid(resource: &ResourceType, method: &CollectionMethod) -> bool {
@@ -140,6 +141,7 @@ impl DataCollector {
         capture_frequency_hz: f32,
     ) -> Result<Self, DataCollectionError> {
         let time_interval_ms = (1000.0 / capture_frequency_hz) as u64;
+        let time_interval = Duration::from_millis(time_interval_ms);
         let component_type = resource.component_type();
         if !resource_method_pair_is_valid(&resource, &method) {
             return Err(DataCollectionError::UnsupportedMethod(
@@ -152,7 +154,7 @@ impl DataCollector {
             component_type,
             resource,
             method,
-            time_interval_ms,
+            time_interval,
         })
     }
 
@@ -177,8 +179,8 @@ impl DataCollector {
         self.component_type.to_string()
     }
 
-    pub fn time_interval(&self) -> u64 {
-        self.time_interval_ms
+    pub fn time_interval(&self) -> Duration {
+        self.time_interval
     }
 
     pub fn method_str(&self) -> String {
@@ -235,12 +237,21 @@ impl DataCollector {
             data: Some(data),
         })
     }
+
+    pub fn resource_method_key(&self) -> ResourceMethodKey {
+        ResourceMethodKey {
+            r_name: self.name(),
+            component_type: self.component_type(),
+            method: self.method.clone(),
+        }
+    }
 }
 
 #[cfg(test)]
 mod tests {
     use std::collections::HashMap;
     use std::sync::{Arc, Mutex};
+    use std::time::Duration;
 
     use super::{CollectionMethod, DataCollectionError, DataCollector, DataCollectorConfig};
     use crate::common::config::{AttributeError, Kind};
@@ -306,6 +317,7 @@ mod tests {
         let conf =
             DataCollectorConfig::try_from(&conf_kind).expect("data collector config parse failed");
         let mut coll = DataCollector::from_config("fake".to_string(), resource, &conf)?;
+        assert_eq!(coll.time_interval(), Duration::from_millis(10));
         let data = coll.call_method()?.data;
         assert!(data.is_some());
         let data = data.unwrap();
--- a/micro-rdk/src/common/data_manager.rs
+++ b/micro-rdk/src/common/data_manager.rs
@@ -1,0 +1,611 @@
+use std::time::Duration;
+
+use crate::common::data_collector::{DataCollectionError, DataCollector};
+use crate::common::data_store::DataStore;
+use crate::proto::app::data_sync::v1::SensorData;
+
+use super::data_collector::ResourceMethodKey;
+use super::data_store::{DataStoreError, WriteMode};
+use async_io::Timer;
+use bytes::BytesMut;
+use thiserror::Error;
+
+#[derive(Debug, Error)]
+pub enum DataManagerError {
+    #[error("no data collectors in manager")]
+    NoCollectors,
+    #[error(transparent)]
+    CollectionError(#[from] DataCollectionError),
+    #[error(transparent)]
+    StoreError(#[from] DataStoreError),
+    #[error("queried time interval {0} is not a multiple of minimum time_interval{1}")]
+    ImproperTimeInterval(u64, u64),
+}
+
+pub struct DataManager<StoreType> {
+    collectors: Vec<DataCollector>,
+    store: StoreType,
+    sync_interval: Duration,
+    min_interval: Duration,
+    part_id: String,
+}
+
+impl<StoreType> DataManager<StoreType>
+where
+    StoreType: DataStore,
+{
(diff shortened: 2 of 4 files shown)


micro-rdk/src/common/data_manager.rs
```diff
@@ -0,0 +1,402 @@
...
+    #[error(transparent)]
+    CollectionError(#[from] DataCollectionError),
+    #[error(transparent)]
+    StoreError(#[from] DataStoreError),
+}
+
+pub struct DataManager<StoreType> {
+    collectors: Vec<DataCollector>,
+    store: StoreType,
+    sync_interval_ms: u64,
```
acmorrow, line 26: Do we potentially gain anything by representing this as a `Duration`? Ditto for the various other places we use `_ms` / `u64`?

micro-rdk/src/common/data_manager.rs
```diff
@@ -0,0 +1,402 @@
...
+        intervals
+    }
+
+    pub async fn run(&mut self) -> Result<(), DataManagerError> {
+        let min_interval = self
+            .collection_intervals()
+            .into_iter()
+            .min()
+            .ok_or(DataManagerError::NoCollectors)?;
+        let min_interval = std::cmp::min(min_interval, self.sync_interval_ms);
```
acmorrow, line 69: Should there be some sort of floor for `min_interval`? Some sort of quantum that all intervals are snapped to to hopefully reduce the number of unique intervals if there are values that are close to the same?

micro-rdk/src/common/data_manager.rs
```diff
@@ -0,0 +1,402 @@
...
+        let mut loop_counter: u64 = 0;
+        loop {
+            Timer::after(Duration::from_millis(min_interval)).await;
+            if (self.sync_interval_ms / min_interval) % loop_counter == 0 {
```
acmorrow, line 73: Should there be some sort of quantization done? Should it round up or down, etc?

micro-rdk/src/common/data_manager.rs
```diff
@@ -0,0 +1,402 @@
...
+                    Err(err) => return Err(err.into()),
+                };
+            }
+            // TODO: implement actual upload logic here, will likely have to change struct
+            // and make this function async
+        }
+        Ok(())
+    }
+
+    pub fn collect_and_store_readings(
```
acmorrow, line 107: Does this need to be `pub`?

micro-rdk/src/common/data_collector.rs
```diff
@@ -230,6 +230,14 @@ impl DataCollector {
             data: Some(data),
         })
     }
+
+    pub fn resource_method_key(&self) -> ResourceMethodKey {
```
acmorrow, line 234: Should this just be `From<DataCollector>` for `ResourceMethodKey`?

micro-rdk/src/common/data_manager.rs
```diff
@@ -0,0 +1,402 @@
...
+    }
+
+    pub(crate) fn collection_intervals(&self) -> Vec<u64> {
+        let mut intervals: Vec<u64> = self.collectors.iter().map(|x| x.time_interval()).collect();
+        intervals.sort();
+        intervals.dedup();
+        intervals
+    }
+
+    pub async fn run(&mut self) -> Result<(), DataManagerError> {
```
acmorrow, line 63: Just thinking out loud, but if this function ever does return a `DataManagerError`, what should be done with that error? Should this function get called again? Are we down forever?

micro-rdk/src/common/data_manager.rs
```diff
@@ -0,0 +1,402 @@
...
+    fn sync(&mut self) -> Result<(), DataManagerError> {
+        for collector_key in self.collectors.iter().map(|c| c.resource_method_key()) {
+            // TODO: check for internet access before attempting to read from store
+            let mut readings_to_upload: Vec<BytesMut> = vec![];
+            loop {
+                match self.store.read_next_message(&collector_key) {
+                    Ok(msg) => {
+                        if msg.is_empty() {
+                            break;
+                        } else {
```
acmorrow, line 94: I don't think you need the`else`

micro-rdk/src/common/data_manager.rs
```diff
@@ -0,0 +1,402 @@
...
+        time_interval_ms: u64,
+    ) -> Result<(), DataManagerError> {
+        for (collector_key, reading) in self.collect_readings_for_interval(time_interval_ms)? {
+            self.store
+                .write_message(&collector_key, reading, WriteMode::OverwriteOldest)?
+        }
+        Ok(())
+    }
+
+    fn collect_readings_for_interval(
```
acmorrow, line 118: So we will end up doing #unique-collection-intervals scans of the collectors array? So, potentially N^2 in number of collectors if all the intervals are different? To be clear, I'm not convinced that is really a problem since I expect N to be very small. Just noting it.

acmorrow (changes requested):
This looks fairly ready to go to me actually, mostly just questions and small things.

micro-rdk/src/common/data_manager.rs
gvaradarajan: Yes, you're right. The logic in `collection_intervals` does not account for this. Will update

micro-rdk/src/common/data_manager.rs
gvaradarajan: That will happen automatically I think. Dividing two integers does floor division I believe

micro-rdk/src/common/data_manager.rs
gvaradarajan: Nope, oops

micro-rdk/src/common/data_collector.rs
gvaradarajan: I considered that, but this would consume the data collector. I forgot that you could do `From<&DataCollector>` though. Would this be preferred?

micro-rdk/src/common/data_manager.rs
gvaradarajan: I believed that to be the responsibility of the task that initialized the manager and calls `run`? I could be convinced otherwise, maybe

micro-rdk/src/common/data_manager.rs
gvaradarajan: Yes, I've taken to programming less computationally efficient solutions for micro-rdk when I believe there's memory to be saved because the number of elements we deal with is usually small. But I may be overlooking something here, I'll double-check the logic

micro-rdk/src/common/data_collector.rs
acmorrow: No preference, I'm still learning what is idiomatic for rust and was curious about this. Reading the third bullet of https://doc.rust-lang.org/std/convert/trait.From.html#when-to-implement-from makes me think it should not be done.

micro-rdk/src/common/data_manager.rs
acmorrow: I'm fine with deferring an answer until later / higher up the stack.

micro-rdk/src/common/data_manager.rs
gvaradarajan: Couldn't think of a way to reduce time complexity without some sort of caching (which wouldn't seem worth it)

acmorrow (approved):
This LGTM though I'm not sure why we wouldn't just use u64 for millis.

micro-rdk/src/common/data_manager.rs
```diff
@@ -0,0 +1,447 @@
...
+            .collect();
+        intervals.sort();
+        intervals.dedup();
+        intervals
+    }
+
+    pub async fn run(&mut self) -> Result<(), DataManagerError> {
+        let mut loop_counter: u64 = 0;
+        let min_interval_ms = self.min_interval_ms();
+        loop {
```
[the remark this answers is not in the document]
gvaradarajan: you're right I have the modulo reversed by accident

micro-rdk/src/common/data_manager.rs
```diff
@@ -79,22 +79,29 @@ where
...
+            if loop_counter % (interval / min_interval_ms) == 0 {
+                self.collect_and_store_readings(interval)?;
             }
-            loop_counter += 1;
         }
+        Ok(())
     }
 
     fn sync(&mut self) -> Result<(), DataManagerError> {
+        println!("nsync");
```
acmorrow: Remove

acmorrow (approved):

micro-rdk/src/common/data_manager.rs
acmorrow: Sorry, this was stale from a previous look at this PR. I know you disbanded nsync already.

PR #182, opened 2024-04-15
[RSDK-7254] generate webrtc cert on esp32

npmenard:
Allow generating a webrtc certificate at runtime

### DIFF
--- a/micro-rdk/src/esp32/certificate.rs
+++ b/micro-rdk/src/esp32/certificate.rs
@@ -4,6 +4,7 @@ use std::{
     mem::MaybeUninit,
 };
 
+use chrono::{NaiveDate, NaiveDateTime};
 use esp_idf_svc::sys::{
     mbedtls_ctr_drbg_context, mbedtls_ctr_drbg_free, mbedtls_ctr_drbg_init,
     mbedtls_ctr_drbg_random, mbedtls_ctr_drbg_seed, mbedtls_ecp_gen_key,
@@ -16,7 +17,8 @@ use esp_idf_svc::sys::{
     mbedtls_x509write_crt_set_issuer_key, mbedtls_x509write_crt_set_issuer_name,
     mbedtls_x509write_crt_set_md_alg, mbedtls_x509write_crt_set_subject_key,
     mbedtls_x509write_crt_set_subject_name, mbedtls_x509write_crt_set_validity,
-    mbedtls_x509write_crt_set_version, SHA_TYPE_SHA2_256,
+    mbedtls_x509write_crt_set_version, MBEDTLS_ERR_PK_BAD_INPUT_DATA, MBEDTLS_X509_CRT_VERSION_3,
+    SHA_TYPE_SHA2_256,
 };
 
 use crate::common::webrtc::certificate::{Certificate, Fingerprint};
@@ -24,7 +26,7 @@ use crate::common::webrtc::certificate::{Certificate, Fingerprint};
 #[derive(Clone)]
 pub struct WebRtcCertificate {
     serialized_der: Vec<u8>,
-    key_pair: Vec<u8>,
+    priv_key: Vec<u8>,
     fingerprint: Fingerprint,
 }
 
@@ -32,7 +34,7 @@ impl<'a> WebRtcCertificate {
     pub fn new(serialized_der: Vec<u8>, key_pair: Vec<u8>, fingerprint: &'a str) -> Self {
         Self {
             serialized_der,
-            key_pair,
+            priv_key: key_pair,
             fingerprint: Fingerprint::try_from(fingerprint).unwrap(),
         }
     }
@@ -43,7 +45,7 @@ impl Certificate for WebRtcCertificate {
         &self.serialized_der
     }
     fn get_der_keypair(&self) -> &'_ [u8] {
-        &self.key_pair
+        &self.priv_key
     }
     fn get_fingerprint(&self) -> &'_ Fingerprint {
         &self.fingerprint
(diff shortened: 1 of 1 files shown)


micro-rdk/src/esp32/certificate.rs
```diff
@@ -28,3 +49,277 @@ impl Certificate for WebRtcCertificate {
...
+
+        unsafe {
+            MbedTLSError::to_unit_result(mbedtls_x509write_crt_set_validity(
+                &mut self.crt_context as *mut mbedtls_x509write_cert,
+                not_before.as_ptr(),
+                not_after.as_ptr(),
+            ))
+        }?;
+
+        let mut work_buffer = vec![0_u8; 2048];
```
acmorrow, line 229: - What guarantees that 2048 is sufficient?
- If there is a known upper bound, can we use a constant, or the `work_buffer`'s capacity in the call to `write_crt_der` below?

micro-rdk/src/esp32/certificate.rs
```diff
@@ -28,3 +49,277 @@ impl Certificate for WebRtcCertificate {
...
+
+        let ret = unsafe {
+            MbedTLSError::to_result(mbedtls_pk_write_key_der(
+                &mut self.kp_context as *mut _,
+                work_buffer.as_mut_ptr(),
+                2048,
+            ))
+        }?;
+
+        let der_kp = work_buffer[2048 - ret as usize..].to_vec();
```
acmorrow, line 267: Is this private key or key pair? `mbedtls_pk_write_key_der` seems to be talking about private keys (`pk`) per the docs, but `der_kp` seems like "key pair" and is in fact passed as the `key_pair` to the `WebRtcCertificate` below. So I'm not quite sure how to read it. If the code is correct and it is just the names that are tricky, maybe make the names more elaborated.

micro-rdk/src/esp32/certificate.rs
```diff
@@ -28,3 +49,277 @@ impl Certificate for WebRtcCertificate {
...
+        assert_eq!(der_kp.len(), ret as usize);
+
+        let fingerprint = Fingerprint::new("sha-256".to_owned(), der_fp);
```
acmorrow, line 270: Maybe move this up to line 258 right after you constructed `der_fp`? It seems out of sequence here since what is being built here is `der_kp`.

micro-rdk/src/esp32/certificate.rs
```diff
@@ -28,3 +49,277 @@ impl Certificate for WebRtcCertificate {
...
+            kp_context.assume_init()
+        };
+        let crt_context = unsafe {
+            let mut crt_context = MaybeUninit::uninit();
+            mbedtls_x509write_crt_init(crt_context.as_mut_ptr());
+            crt_context.assume_init()
+        };
+
+        Self {
+            oid_cn: String::from("VIAM Webrtc"),
```
acmorrow, line 304: Maybe `WebRTC`?

micro-rdk/src/esp32/certificate.rs
```diff
@@ -28,3 +49,277 @@ impl Certificate for WebRtcCertificate {
...
+            oid_c: String::from("US"),
+            oid_o: String::from("VIAM"),
```
acmorrow, line 306: Is there some existing art that indicates these specific values? Do they need to conform with anything specific?

micro-rdk/src/esp32/certificate.rs
```diff
@@ -28,3 +49,277 @@ impl Certificate for WebRtcCertificate {
...
+        self
+    }
+    pub fn build(mut self) -> Result<WebRtcCertificate, MbedTLSError> {
+        unsafe {
+            MbedTLSError::to_unit_result(mbedtls_ctr_drbg_seed(
+                &mut self.drbg_context as *mut mbedtls_ctr_drbg_context,
+                Some(mbedtls_entropy_func),
+                &mut self.entropy as *mut _ as *mut c_void,
+                std::ptr::null(),
+                0,
```
acmorrow, line 156: - Is `mbedtls_entropy_func` wired up to hardware rng on the esp32 automatically for us in some way? Or should we be looking to connect to it with `entropy_add_source()` and `esp[_fill]_random`?
- Do we want a static personalization string here just for good measure?

acmorrow (changes requested):

micro-rdk/src/esp32/certificate.rs
npmenard: This is the pattern they use on esp-idf so I am tempted to not stray too much. I am trying to understand whether this is enough for certificate generation.

micro-rdk/src/esp32/certificate.rs
```diff
@@ -28,3 +49,277 @@ impl Certificate for WebRtcCertificate {
...
+            MbedTLSError::to_unit_result(mbedtls_x509write_crt_set_issuer_name(
+                &mut self.crt_context as *mut mbedtls_x509write_cert,
+                subject_name.as_ptr(),
+            ))
+        }?;
+
+        unsafe {
+            MbedTLSError::to_unit_result(mbedtls_x509write_crt_set_subject_name(
+                &mut self.crt_context as *mut mbedtls_x509write_cert,
+                subject_name.as_ptr(),
```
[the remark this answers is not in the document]
npmenard: yes it's a self signed cert so issuer == subject (at least that is my understanding)

micro-rdk/src/esp32/certificate.rs
npmenard: unless we start adding a lot of extension and extra information we should be safe with 2048

micro-rdk/src/esp32/certificate.rs
npmenard: No in most cases they recommend to set them so some sane values but its doesn't look like it's super critical since it is self signed.

micro-rdk/src/esp32/certificate.rs
npmenard: you're correct it's just the private key

micro-rdk/src/esp32/certificate.rs
```diff
@@ -1,17 +1,40 @@
...
+    priv_key: Vec<u8>,
     fingerprint: Fingerprint,
 }
 
 impl<'a> WebRtcCertificate {
     pub fn new(serialized_der: Vec<u8>, key_pair: Vec<u8>, fingerprint: &'a str) -> Self {
         Self {
             serialized_der,
-            key_pair,
+            priv_key: key_pair,
```
acmorrow, line 37: Still a few places where it says `pair` around here.

acmorrow (approved):
LGTM mod a naming nit.

PR #183, opened 2024-04-15
[RSDK-7248] add provisioning api

npmenard:
Add provisioning server and Storage trait

### DIFF
--- a/micro-rdk/src/common/provisioning/server.rs
+++ b/micro-rdk/src/common/provisioning/server.rs
@@ -1,14 +1,20 @@
 #![allow(dead_code)]
-use std::{pin::Pin, rc::Rc, sync::Mutex};
+use std::{marker::PhantomData, pin::Pin, rc::Rc, task::Poll};
 
 use bytes::{BufMut, Bytes, BytesMut};
-use futures_lite::Future;
+use futures_lite::{Future, FutureExt};
 use http_body_util::BodyExt;
-use hyper::{body::Incoming, header::CONTENT_TYPE, http, service::Service, Request, Response};
+use hyper::{
+    body::Incoming, header::CONTENT_TYPE, http, rt, server::conn::http2, service::Service, Request,
+    Response,
+};
 use prost::Message;
 
 use crate::{
-    common::grpc::{GrpcBody, GrpcError, GrpcResponse},
+    common::{
+        grpc::{GrpcBody, GrpcError, GrpcResponse},
+        webrtc::api::AtomicSync,
+    },
     proto::provisioning::{
         self,
         v1::{
@@ -51,7 +57,8 @@ impl ProvisioningServiceBuilder {
             provisioning_info: Rc::new(self.provisioning_info),
             last_connection_attempt: Rc::new(self.last_connection_attempt),
             reason: Rc::new(self.reason),
-            storage: Rc::new(Mutex::new(storage)),
+            storage,
+            credential_ready: AtomicSync::default(),
         }
     }
 }
@@ -84,7 +91,8 @@ struct ProvisioningService<S> {
     provisioning_info: Rc<Option<ProvisioningInfo>>,
     last_connection_attempt: Rc<Option<NetworkInfo>>,
     reason: Rc<ProvisioningReason>,
-    storage: Rc<Mutex<S>>,
+    storage: S,
+    credential_ready: AtomicSync,
 }
 
 impl<S> ProvisioningService<S>
@@ -121,7 +129,7 @@ where
             resp.errors
                 .push("stored credentials are invalid".to_owned())
         }
-        resp.has_smart_machine_credentials = self.storage.lock().unwrap().has_stored_credentials();
+        resp.has_smart_machine_credentials = self.storage.has_stored_credentials();
         let len = resp.encoded_len();
         let mut buffer = BytesMut::with_capacity(5 + len);
         buffer.put_u8(0);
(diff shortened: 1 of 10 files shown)


micro-rdk/src/common/provisioning/server.rs
```diff
@@ -0,0 +1,528 @@
...
+        },
+    };
+
+    use super::ProvisioningService;
+
+    async fn run_provisioning_server(
+        ex: NativeExecutor,
+        srv: ProvisioningService<MemoryCredentialStorage>,
+    ) {
+        let listen = TcpListener::bind("127.0.0.1:56432");
```
gvaradarajan, line 296: maybe make this url & port a constant since it has repeated use?

gvaradarajan (approved):
LGTM, only a couple minor comments (and a question)

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -0,0 +1,528 @@
...
+{
+    type Output = Result<(), hyper::Error>;
+    fn poll(
+        self: Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Self::Output> {
+        let this = self.project();
+
+        match this.connection.poll(cx) {
+            Poll::Pending => futures_lite::ready!(this.credential_ready.poll(cx)),
```
npmenard, line 228: At this time it's unclear whether we should shutdown provisioning when credential have been stored or if we let the provisioning client close the connection.
Since we do not envision running the provisioning server while the main server is running we will have to pick a poison :)

gvaradarajan: I shudder to think that this is possible, but could this not potentially be used in the future to reconfigure the robot (in which case we would not want to shut it down)?

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -0,0 +1,528 @@
...
+            .await
+            .map_err(|_| GrpcError::RpcFailedPrecondition)?
+            .to_bytes();
+        match parts.uri.path() {
+            "/viam.provisioning.v1.ProvisioningService/GetSmartMachineStatus" => {
+                self.get_smart_machine_status()
+            }
+            "/viam.provisioning.v1.ProvisioningService/SetSmartMachineCredentials" => {
+                self.set_smart_machine_credentials(body.split_off(5))
+            }
```
acmorrow, line 116: During what part of the lifetime of a micro-rdk run will these endpoints be available?

acmorrow (commented):
I'm still trying to make sure I understand the flow of control here, but here are my initial comments

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -0,0 +1,528 @@
...
+        self
+    }
+    fn with_reason(mut self, reason: ProvisioningReason) -> Self {
+        self.reason = reason;
+        self
+    }
+    fn with_network_info(mut self, info: NetworkInfo) -> Self {
+        let _ = self.last_connection_attempt.insert(info);
+        self
+    }
```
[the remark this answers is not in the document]
npmenard: doing this will do a "move out of mutable reference"  when chaining calls. We could wrap it in an inner type like for fs::OpenOptions but seems a bit contrive for this purpose

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -0,0 +1,528 @@
...
+            reason: Rc::new(self.reason),
+            storage,
+            credential_ready: AtomicSync::default(),
+        }
+    }
+}
+
+#[derive(PartialEq)]
+enum ProvisioningReason {
+    Erased,
```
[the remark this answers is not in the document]
npmenard: ok for `Unprovisioned`

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -0,0 +1,528 @@
...
+    fn set_manufacturer(&mut self, manufacturer: String) {
+        self.0.manufacturer = manufacturer;
+    }
+}
+
+#[derive(Clone)]
+struct ProvisioningService<S> {
+    provisioning_info: Rc<Option<ProvisioningInfo>>,
+    last_connection_attempt: Rc<Option<NetworkInfo>>,
+    reason: Rc<ProvisioningReason>,
```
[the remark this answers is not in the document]
npmenard: yep we need clone to be cheap since the Service will be cloned for each request

micro-rdk/src/common/provisioning/server.rs
npmenard: just during provisioning

micro-rdk/src/common/provisioning/server.rs
npmenard: +1

micro-rdk/src/common/provisioning/storage.rs
```diff
@@ -0,0 +1,45 @@
...
+    fn has_stored_credentials(&self) -> bool {
+        self.robot_id.lock().unwrap().is_some() && self.robot_secret.lock().unwrap().is_some()
+    }
+    fn store_robot_credentials(&self, cfg: CloudConfig) -> Result<(), Self::Error> {
+        // TODO: make ticket : ignore app_address for now but need to add it later
+        let _ = self.robot_id.lock().unwrap().insert(cfg.id);
+        let _ = self.robot_secret.lock().unwrap().insert(cfg.secret);
+        Ok(())
+    }
+    fn get_robot_credentials(&self) -> Result<(String, String), Self::Error> {
```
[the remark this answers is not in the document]
npmenard: Yep it works in single_thread but may break in multithreaded. I wrapped it all in a struct for good measure

PR #184, opened 2024-04-17
implement configuration logic for DataManager

gvaradarajan:


micro-rdk/src/common/data_manager.rs
```diff
@@ -20,6 +24,37 @@ pub enum DataManagerError {
...
+    InitializationRobotError(#[from] RobotError),
+}
+
+fn get_data_sync_interval(cfg: &ConfigResponse) -> Result<Option<Duration>, DataManagerError> {
+    let robot_config = cfg.config.clone().ok_or(DataManagerError::ConfigError)?;
+    Ok(
+        if let Some(data_cfg) = robot_config
+            .services
+            .iter()
+            .find(|svc_cfg| svc_cfg.r#type == *"data_manager")
```
acmorrow, line 39: What if there is more than one in the config?

micro-rdk/src/common/data_manager.rs
```diff
@@ -51,6 +86,24 @@ where
         })
     }
 
+    pub fn from_robot_and_config(
```
acmorrow, line 89: I'll admit I have some worries about the fact that the service isn't being created via the same configuration parsing and instantiation logic that works with the regular robot config. If the `DataManagerService` were to become a more active component, in the sense of having inbound-to-robot RPCs that we needed to honor, how would those get acted on?

I'm willing to handwave it for now I think, and let the data manager be a one-off, but if we ever wanted to implement another service, I think at that point I'd want to see the micro-rdk grow a real service concept and then fold the data service in under that mechanism.

Also worth noting that currently, there isn't any real difference between a component and a service, nor does micro-rdk have a concept of resource/component/service that I'm aware of.

In any event, I'm not proposing we make changes now - just pointing it out as future work.

micro-rdk/src/native/entry.rs
```diff
@@ -84,6 +88,14 @@ pub async fn serve_web_inner(
         (cfg_response, robot)
     };
 
+    #[cfg(feature = "data")]
+    // TODO: Spawn data task here. May have to move the initialization below to the task itself
+    let _data_manager_svc = DataManager::<StaticMemoryDataStore>::from_robot_and_config(
```
acmorrow, line 93: I think at some point we will want DataManager to work in terms of `dyn DataStore`, but given that we currently have only one implementation, I don't mind making the binding static.

Presumably, though, at some point the choice of `DataStore` would be 1) defaulted and 2) configurable from the config.

micro-rdk/src/common/data_store.rs
```diff
@@ -70,6 +70,13 @@ pub trait DataStore {
         &mut self,
         collector_key: &ResourceMethodKey,
     ) -> Result<BytesMut, DataStoreError>;
+
+    /// Initializes from resource-method keys with a separate memory region dedicated to each key.
```
acmorrow, line 74: - The part of the doc comment about memory regions seems too implementation focused for trait-level documentation.
- I'm a little unclear on why the trait should be mandating the form of construction for implementations.

acmorrow (changes requested):
Looks basically good, with a few small comments here and there but nothing that I think will hold up approving soon.

micro-rdk/src/common/data_manager.rs
gvaradarajan: I don't think this is an expected occurrence, though possible. Now that you mention, I'm just going to throw a special error if there are multiple

micro-rdk/src/native/entry.rs
gvaradarajan: agreed, I'll put in a todo (plus JIRA ticket) to this effect

micro-rdk/src/common/data_store.rs
gvaradarajan: I think the concept of a data store only makes sense in the context of data management, which will always require that the store be configurable based on the collectors, so this doesn't feel like a very strict mandate. I do agree about taking out the "memory regions" part of the doc comment though

micro-rdk/src/common/data_manager.rs
acmorrow: Just because it lacks a robust outward facing api today, for our purposes, doesn't mean there won't be one in the future, even if our means of transporting data is push vs pull, there could be other functionality added later. I'm hopeful that when we need to add the next service, whatever it is, that we can find a way to pull the data stuff into that framework. But, it isn't something to worry about now, anyway.

acmorrow (approved):

PR #185, opened 2024-04-18
implement data task

gvaradarajan:


micro-rdk/src/common/conn/server.rs
```diff
@@ -324,12 +329,13 @@ where
...
-                let _ = self.app_client.insert(app_client);
+                let _ = self.app_client.lock().await.insert(app_client);
             }
-
+            let mut app_client_guard = self.app_client.lock().await;
+            let app_client = app_client_guard.as_mut().unwrap();
             let sig = if let Some(webrtc_config) = self.webrtc_config.as_ref() {
                 let ip = self.app_config.get_ip();
-                let signaling = self.app_client.as_mut().unwrap().connect_signaling();
+                let signaling = app_client.connect_signaling();
```
gvaradarajan, line 338: @acmorrow  `signaling` represents a borrow of `app_client` and by extension `app_client_guard`, which means that as long as signaling is in scope the lock never releases. Subsequently we never get it back in L. 199 of `data_manager.rs`. And that's why I think this PR is garbage

acmorrow: My suggestion for now is to see if you can sidestep the issue by not sharing an app client and letting data sync have its own. If that works, then you can get the rest of things working and we can come back to sharing the app client as a separate piece of work.

gvaradarajan:
Approach does not work

PR #186, opened 2024-04-19
RSDK-7301 Add missing error checks for number parsing in board

acmorrow:
I was able to fairly easily reproduce the error in https://viam.atlassian.net/browse/RSDK-7301 by passing a non-numeric argument to `pin.Set` with the golang sdk:

```
	myBoard, err := board.FromRobot(machine, "board")
	if err != nil {
		logger.Fatal(err)
	}

	// Get the GPIOPin with pin number 15.
	pin, err := myBoard.GPIOPinByName("q")
	if err != nil {
		logger.Fatal(err)
	}

	// Set the pin to high.
	err = pin.Set(context.Background(), true, nil)
	if err != nil {
		logger.Fatal(err)
	}
```

### DIFF
--- a/micro-rdk/src/common/grpc.rs
+++ b/micro-rdk/src/common/grpc.rs
@@ -643,7 +643,10 @@ where
             None => return Err(ServerError::from(GrpcError::RpcUnavailable)),
         };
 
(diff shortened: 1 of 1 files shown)


gvaradarajan (approved):
LGTM

PR #188, opened 2024-04-20
[RSDK-7253] discovery mdns

npmenard:
Adding mDNS service & examples. For now it uses MemroyCredential Storage so we can test the flow with the mobile app. We will do something more permanent for the esp32 at a later stage

### DIFF
--- a/examples/esp32/esp32-server.rs
+++ b/examples/esp32/esp32-server.rs
@@ -10,15 +10,12 @@ mod esp32 {
     include!(concat!(env!("OUT_DIR"), "/robot_secret.rs"));
 
     use log::*;
+    use micro_rdk::common::entry::RobotRepresentation;
     #[cfg(feature = "qemu")]
     use micro_rdk::esp32::esp_idf_svc::eth::EspEth;
     use micro_rdk::esp32::esp_idf_svc::eventloop::EspSystemEventLoop;
     use micro_rdk::esp32::esp_idf_svc::sys::{
         g_wifi_feature_caps, CONFIG_FEATURE_CACHE_TX_BUF_BIT,
-    };
-    use micro_rdk::{
-        common::{app_client::AppClientConfig, entry::RobotRepresentation},
-        esp32::{certificate::WebRtcCertificate, entry::serve_web, tls::Esp32TLSServerConfig},
     };
     #[cfg(feature = "qemu")]
     use std::net::Ipv4Addr;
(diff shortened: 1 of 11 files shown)


micro-rdk/src/native/entry.rs
```diff
@@ -120,6 +135,117 @@ pub async fn serve_web_inner(
...
+    storage: S,
+    last_error: Option<String>,
+) -> Result<(AppClientConfig, NativeTlsServerConfig), Box<dyn std::error::Error>>
+where
+    S: CredentialStorage + Clone + 'static,
+    S::Error: Debug,
+    GrpcError: From<S::Error>,
+{
+    let hostname = format!(
+        "provisioning-{}-{}",
```
acmorrow, line 152: Should this `provisioning` string include some more detail? This is the viam v1 provisioning API that is served here.

micro-rdk/src/native/entry.rs
```diff
@@ -120,6 +135,117 @@ pub async fn serve_web_inner(
...
+
+    let srv = ProvisioningServiceBuilder::new().with_provisioning_info(info);
+
+    let srv = if let Some(error) = last_error {
+        srv.with_last_error(error)
+    } else {
+        srv
+    };
+    let srv = srv.build(storage.clone());
+    let listen = TcpListener::bind(ip.to_string() + ":0")?;
```
acmorrow, line 166: For native, does it make sense to only bind to one v4 ip if the device is multi-homed?

micro-rdk/src/native/entry.rs
```diff
@@ -120,6 +135,117 @@ pub async fn serve_web_inner(
...
+    let listen: Async<TcpListener> = listen.try_into()?;
+
+    let port = listen.get_ref().local_addr()?.port();
+
+    log::info!("provisioning server run on {}:{}", ip, port);
+
+    mdns.add_service(
+        "provisioning",
```
acmorrow, line 174: Same questions re the `provisioning` string.

micro-rdk/src/esp32/entry.rs
```diff
@@ -142,6 +156,136 @@ pub async fn serve_web_inner(
...
+        srv
+    };
+    let srv = srv.build(storage.clone());
+    let listen = TcpListener::bind(ip.to_string() + ":0")?;
+    let listen: Async<TcpListener> = listen.try_into()?;
+
+    let port = listen.get_ref().local_addr()?.port();
+
+    mdns.add_service(
+        "provisioning",
```
acmorrow, line 196: If this string is meant to be the same between native and esp32, maybe it should be a constant somewhere. Ditto other uses of the `provisioning` string. Also, see my comment in the native version of this about the exact contentx of the string.

micro-rdk/src/esp32/entry.rs
```diff
@@ -142,6 +156,136 @@ pub async fn serve_web_inner(
...
+        "_rpc",
+        "_tcp",
+        port,
+        &[("provisioning", "")],
+    )?;
+    loop {
+        let incoming = listen.accept().await;
+        let (stream, _) = incoming?;
+
+        let stream = Esp32Stream::LocalPlain(stream);
```
acmorrow, line 206: I think a comment is warranted anytime we create something unencrypted.

micro-rdk/src/native/entry.rs
```diff
@@ -120,6 +135,117 @@ pub async fn serve_web_inner(
...
+        "_tcp",
+        port,
+        &[("provisioning", "1")],
+    )?;
+    loop {
+        let incoming = listen.accept().await;
+        let (stream, _) = incoming?;
+        log::info!("will attempt to provision");
+
+        let stream = NativeStream::LocalPlain(stream);
```
acmorrow, line 185: Ditto comment for anything unencrypted

acmorrow (approved):
LGTM. I left a few small suggestions and comment requests in a few places, but I don't think anything that'd require another round of review.

examples/esp32/esp32-server.rs
```diff
@@ -93,25 +90,56 @@ mod esp32 {
...
+        }
+        #[cfg(feature = "provisioning")]
+        {
+            use micro_rdk::common::provisioning::{
+                server::ProvisioningInfo, storage::MemoryCredentialStorage,
+            };
+            let mut info = ProvisioningInfo::default();
+            info.set_fragment_id("d385b480-3d19-4fad-a928-b5c18a58d0ed".to_string());
+            info.set_manufacturer("viam".to_owned());
+            info.set_model("test-esp32".to_owned());
```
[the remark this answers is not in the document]
npmenard: Yes we should but I would like to keep it to a separate ticket, we don't know for now how we are going to let user do provisioning (we wait for the mobile app for that) Since its gated behind a feature we don't have to integrate perfectly. The fragment is public and contains just a fake board

micro-rdk/src/native/conn/mdns.rs
```diff
@@ -31,19 +40,21 @@ impl Mdns for NativeMdns {
...
     ) -> Result<(), MdnsError> {
         let ty_domain = format!("{}.{}.local.", service_type.as_ref(), proto.as_ref());
+        let srv_hostname = format!("{}.{}", self.hostname, &ty_domain);
 
         let props: HashMap<String, String> = txt
             .iter()
             .map(|(k, v)| ((*k).into(), (*v).into()))
             .collect();
 
+        //let properties = [("PATH", "one"), ("Path", "two"), ("PaTh", "three")];
```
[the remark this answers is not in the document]
npmenard: +1

micro-rdk/src/native/entry.rs
npmenard: Maybe but not sure :)

micro-rdk/src/native/entry.rs
npmenard: My assumption is that the service will be discoverable via mDNS so if the server ip can be reached you should see it. Note that the assumption could be wrong since on your computer it didn't seem to works.

PR #189, opened 2024-04-23
implement data task

gvaradarajan:


gvaradarajan:
NOTE: works for native, stack overflow on ESP32

gvaradarajan:
now works on esp32, just needed slightly more stack for the main task. Waiting for approval on the general approach

micro-rdk/src/common/data_manager.rs
```diff
@@ -165,23 +198,30 @@ where
...
+                        ..Default::default()
+                    }),
+                    sensor_contents: vec![],
                 };
+                let data: Result<Vec<SensorData>, DecodeError> = readings_to_upload
+                    .into_iter()
+                    .map(SensorData::decode)
+                    .collect();
+                upload_request.sensor_contents = data?;
+                app_client.upload_data(upload_request).await?;
```
acmorrow, line 223: Is `readings_to_upload` held across the await? If so, could that be avoided?

acmorrow (commented):
I know this is going to close, but here are my thoughts.

gvaradarajan:
closing, this approach is now outdated

PR #191, opened 2024-04-29
[RSDK-7439] chore: better wifi errors

mattjperez:
Previously, micro-rdk would just throw a panic with a line number when failing to start wifi (ssid/pass misconfiguration, etc).

This change gives slightly better error messages and includes the `EspError` value, ex. `failed to start wifi: EspError(263)`.

Adding these changes highlighted a couple code smells. The changes were repeated in three separate areas in almost the same way. Would be great if we could move the wifi setup code into the micrordk for users to pull and initialize with a couple lines instead of the boilerplate. It also enables us to provide better error messaging; the current error outputs are esp specific and need to be looked up [here](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/error-codes.html)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,4 +5,3 @@ Cargo.lock
 /canary/.venv/*
 .vscode
 components_esp32.lock
-.envrc
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ Cargo.lock
 /canary/.venv/*
 .vscode
 components_esp32.lock
+.envrc
--- a/examples/esp32/esp32-server.rs
+++ b/examples/esp32/esp32-server.rs
@@ -89,8 +89,15 @@ mod esp32 {
         #[allow(clippy::redundant_clone)]
         #[cfg(not(feature = "qemu"))]
         let (ip, _wifi) = {
-            let wifi = start_wifi(periph.modem, sys_loop_stack).unwrap();
-            (wifi.wifi().sta_netif().get_ip_info().unwrap().ip, wifi)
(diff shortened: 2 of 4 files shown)


acmorrow (approved):
LGTM. There will still be panics but at least now they will give some insight into the problem.

.gitignore
```diff
@@ -5,3 +5,4 @@ Cargo.lock
 /canary/.venv/*
 .vscode
 components_esp32.lock
+.envrc
```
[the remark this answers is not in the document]
mattjperez: sounds good

examples/esp32/esp32-server.rs
```diff
@@ -89,8 +89,15 @@ mod esp32 {
         #[allow(clippy::redundant_clone)]
         #[cfg(not(feature = "qemu"))]
         let (ip, _wifi) = {
-            let wifi = start_wifi(periph.modem, sys_loop_stack).unwrap();
-            (wifi.wifi().sta_netif().get_ip_info().unwrap().ip, wifi)
+            let wifi = start_wifi(periph.modem, sys_loop_stack).expect("failed to start wifi");
```
gvaradarajan, line 92: since `start_wifi` returns an EspError instance can we print the error in the panic statement. This means calling `panic` instead of `expect` but I think it's worth the visibility (same comment applies to the corresponding change in the template)

gvaradarajan (changes requested):
LGTM, just one adjustment

examples/esp32/esp32-server.rs
gvaradarajan: resolved offline, change unnecessary

gvaradarajan (approved):

PR #195, opened 2024-05-01
RSDK-7387 Handle incoming http2 connections in a coroutine

acmorrow:
Per discussions, `ViamServer::serve` is going to need to be decomposed into co-routines which can run concurrently.

The current implementation only does that for incoming WebRTC connections. Incoming HTTP2 connections simply took over service until complete. This PR changes handling for HTTP2 connections to work similarly to how WebRTC connections work.

HTTP2 connections are afforded the highest priority, since this seemed closest in spirit to the existing behavior, where established inbound HTTP2 connection would simply block new connection handling.

Creating this as draft for now while I do some more testing, but please feel free to provide any feedback you may have now.

### DIFF
--- a/examples/native/native-server.rs
+++ b/examples/native/native-server.rs
@@ -9,7 +9,9 @@ mod native {
     };
 
     pub(crate) fn main_native() {
-        env_logger::init();
+        env_logger::builder()
+            .format_timestamp(Some(env_logger::TimestampPrecision::Millis))
+            .init();
 
         let repr = RobotRepresentation::WithRegistry(Box::default());
 
--- a/micro-rdk/src/common/conn/server.rs
+++ b/micro-rdk/src/common/conn/server.rs
@@ -400,9 +400,13 @@ where
                     Ok(c) => {
                         let robot = robot.clone();
                         let exec = self.exec.clone();
-                        let t = self.exec.spawn(async move { Self::serve_http2(c, exec, robot).await });
+                        let t = self
+                            .exec
+                            .spawn(async move { Self::serve_http2(c, exec, robot).await });
                         // Incoming direct HTTP2 connections take top priority.
-                        self.incoming_connection_manager.insert_new_conn(t, u32::MAX).await;
+                        self.incoming_connection_manager
+                            .insert_new_conn(t, u32::MAX)
+                            .await;
                         Ok(())
                     }
                 },
@@ -411,7 +415,9 @@ where
                     Ok(_) => {
                         let prio = c.prio;
                         let t = self.exec.spawn(async move { c.run().await });
-                        self.incoming_connection_manager.insert_new_conn(t, prio).await;
+                        self.incoming_connection_manager
+                            .insert_new_conn(t, prio)
+                            .await;
                         Ok(())
                     }
                 },
@@ -424,8 +430,7 @@ where
         connection: T,
         exec: Executor,
         robot: Arc<Mutex<LocalRobot>>,
-    ) -> Result<(), ServerError>
-    {
+    ) -> Result<(), ServerError> {
         let srv = GrpcServer::new(robot.clone(), GrpcBody::new());
         Box::new(
             http2::Builder::new(exec)
--- a/micro-rdk/src/common/conn/server.rs
+++ b/micro-rdk/src/common/conn/server.rs
@@ -26,7 +26,7 @@ use crate::{
 
 use async_io::Timer;
 use futures_lite::prelude::*;
-use futures_lite::{future::Boxed, ready, Future};
+use futures_lite::{future::Boxed, ready};
 use hyper::{rt, server::conn::http2};
 
 use async_executor::Task;
@@ -277,7 +277,7 @@ pub struct ViamServer<'a, C, T, CC, D, L> {
     app_connector: C,
     app_config: AppClientConfig,
     app_client: Option<AppClient<'a>>,
-    webrtc_manager: WebRTCConnectionManager,
+    incoming_connection_manager: IncomingConnectionManager,
 }
 impl<'a, C, T, CC, D, L> ViamServer<'a, C, T, CC, D, L>
 where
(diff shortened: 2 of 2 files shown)


micro-rdk/src/common/conn/server.rs
```diff
@@ -26,7 +26,7 @@ use crate::{
 
 use async_io::Timer;
 use futures_lite::prelude::*;
-use futures_lite::{future::Boxed, ready, Future};
+use futures_lite::{future::Boxed, ready};
```
acmorrow, line 29: `Future` is already in via `prelude::*` so there was a warning I cleaned up.

micro-rdk/src/common/conn/server.rs
```diff
@@ -277,7 +277,7 @@ pub struct ViamServer<'a, C, T, CC, D, L> {
     app_connector: C,
     app_config: AppClientConfig,
     app_client: Option<AppClient<'a>>,
-    webrtc_manager: WebRTCConnectionManager,
+    incoming_connection_manager: IncomingConnectionManager,
```
acmorrow, line 280: Potentially just `ConnectionManager`.

micro-rdk/src/common/conn/server.rs
```diff
@@ -411,19 +426,14 @@ where
             }
         }
     }
-    async fn serve_http2<U>(
-        &self,
-        mut c: U,
+    async fn serve_http2(
```
acmorrow, line 429: This function is now simple enough it could probably just be inlined above, but I think I'd like to keep it rather than making `serve` more complex.

gvaradarajan (approved):
I like this unification a lot

micro-rdk/src/common/conn/server.rs
```diff
@@ -411,19 +426,14 @@ where
...
-    where
-        U: Http2Connector<Stream = T>,
-    {
+    ) -> Result<(), ServerError> {
         let srv = GrpcServer::new(robot.clone(), GrpcBody::new());
-        let connection = c.accept().await.map_err(|e| ServerError::Other(e.into()))?;
-
         Box::new(
-            http2::Builder::new(self.exec.clone())
+            http2::Builder::new(exec)
```
acmorrow, line 436: ⬇️ I think we may need to raise `max_concurrent_streams` to allow inbound gRPC/HTTP2 to function. At least with a python client, I see it attempt to open additional streams after some time connected, even when making a request every second, and opening that additional stream fails if this is set to one.

@npmenard - How costly would it be to set this to `2`, which seems sufficient to let a connected python client not fail after a short time.

micro-rdk/src/common/conn/server.rs
mattjperez: +1 for `connection_manager` and `ConnectionManager`

micro-rdk/src/common/conn/server.rs
```diff
@@ -411,19 +426,14 @@ where
...
                 .initial_connection_window_size(2048)
                 .initial_stream_window_size(2048)
                 .max_send_buf_size(4096)
```
mattjperez, line 439: Would it be better to move this functionality to [here](https://github.com/acmorrow/micro-rdk/blob/45c9de1f4efc39cac7691a14d1887788ddc6c082/micro-rdk/src/common/conn/server.rs#L505)?

Run can take an enum with the appropriate arguments. something like `c.run(Serve::Http{exec, robot})`

mattjperez (approved):
Looks great! Definitely is better to keep general behavior consistent across connection types. As a user, I wouldn't assume these behave differently.

micro-rdk/src/common/conn/server.rs
acmorrow: Maybe, but I often try to avoid moving code while also changing it. It makes it much harder to review. I think I'd like to keep it here for now until I've finished iterating on `ViamServer::serve`. I'll add a TODO.

micro-rdk/src/common/conn/server.rs
acmorrow: I think I changed my mind. It really is only handling inbound connections - ones we accepted.

micro-rdk/src/common/conn/server.rs
mattjperez: sounds good. maybe tag it with `good-first-issue`. we'll need more of those kinds of tasks soon

examples/native/native-server.rs
```diff
@@ -9,7 +9,9 @@ mod native {
     };
 
     pub(crate) fn main_native() {
-        env_logger::init();
+        env_logger::builder()
```
acmorrow, line 12: A drive-by: while debugging on native I found the logging timestamps were missing millisecond detail.

micro-rdk/src/common/conn/server.rs
acmorrow: I've made the change and went to two streams permitted. The Python SDK, for one, has a periodic task which calls `RobotService/ResourceNames` to validate connection. Without the option for a second stream, this will fail and cause the python client to disconnect.

acmorrow:
I think this is ready for a real look. There is still something surprising: a python client connected to the native server via gRPC/HTTP2 eventually disconnects after repeated timeouts in its "connection validation" logic which calls `ResourceNames` periodically. I'm going to test tomorrow to see if I see the same behavior on `main`. If so then it won't be a regression and we can probably move forward with this, else we will need to figure out what I've broken.

micro-rdk/src/common/conn/server.rs
npmenard: While building or returned by clippy?

micro-rdk/src/common/conn/server.rs
npmenard: Yes that tracks, we limited it to one for the ESP32-WROOM which we dropped

micro-rdk/src/common/conn/server.rs
npmenard: agreed

npmenard (approved):
LGTM

mattjperez (approved):

micro-rdk/src/common/conn/server.rs
acmorrow: While building native on my mac.

PR #196, opened 2024-05-02
Fix esp32's HTTP2 server

npmenard:
Essentially remove support for setting a CA cert in  the TLS config and use a TcpListener instead of a raw socket


mattjperez (commented):
Changes look good.

What was the motivation behind this though? Didn't see an associated ticket.

acmorrow (approved):
LGTM

PR #198, opened 2024-05-04
[RSDK-7313] Don't panic when receiving trailers early

npmenard:


micro-rdk/src/common/conn/server.rs
```diff
@@ -382,18 +382,19 @@ where
 
             let connection = match connection {
                 Ok(c) => c,
+
```
acmorrow, line 385: This file looks like whitespace changes only so it can probably be reverted.

acmorrow (approved):
LGTM. Am I correct that the goal here isn't so much fixing the "elusive issue" mentioned in the ticket, but rather 1) avoiding the unwrap panic and instead 2) kicking a more informative error further up the stack in order to let us understand the issue more completely?

micro-rdk/src/common/grpc_client.rs
```diff
@@ -116,7 +118,33 @@ where
...
+                Ok(data) => data,
+                Err(e) => {
+                    if let Some(trailers) = e.trailers_ref() {
+                        if trailers.contains_key("grpc-message")
+                            && trailers.contains_key("grpc-status")
+                        {
+                            return Poll::Ready(Some(Err(GrpcClientError::GrpcError {
+                                code: trailers
+                                    .get("grpc-status")
+                                    .unwrap()
```
mattjperez, line 131: Can we map these unwraps (and the ones immediately below) with one of the GrpcClientErrors?

npmenard: Good point but I don't want to clutter grpc error enum. Added defaults instead

npmenard:
> LGTM. Am I correct that the goal here isn't so much fixing the "elusive issue" mentioned in the ticket, but rather 1) avoiding the unwrap panic and instead 2) kicking a more informative error further up the stack in order to let us understand the issue more completely?

Yes the error is not surfaced in  logs but could be if we wanted

PR #199, opened 2024-05-08
[RSDK-7221] relax pinned versions of rustls

npmenard:
This PR updates rustls and viam-rust-utils allowing to relax the pinned version of futures-rustls

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -57,7 +57,7 @@ esp-idf-svc = { version = "=0.48.1", default-features = false }
 espflash = { git = "https://github.com/viamrobotics/espflash.git", branch = "monitor-output" }
 futures = "0.3.28"
 futures-lite = "1"
-futures-rustls = "=0.22.0"
+futures-rustls = "0.24"
 futures-util = "0.3.30"
 gethostname = "0.4.3"
 http-body-util = "0.1.1"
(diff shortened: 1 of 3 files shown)


mattjperez (approved):

Cargo.toml
```diff
@@ -57,7 +57,7 @@ esp-idf-svc = { version = "=0.48.1", default-features = false }
 espflash = { git = "https://github.com/viamrobotics/espflash.git", branch = "monitor-output" }
 futures = "0.3.28"
 futures-lite = "1"
-futures-rustls = "=0.22.0"
+futures-rustls = "0.24"
```
acmorrow, line 60: Just curious why 0.24 and not the latest which is 0.26.

Cargo.toml
```diff
@@ -86,7 +86,7 @@ reqwest = "0.11.20"
 reqwless = "0.5.0"
 ring = "0.16.20"
 ringbuf = "0.3.3"
-rustls = { version = "0.20.7", features = ["logging", "tls12"] }
+rustls = { version = "0.21", features = ["logging", "tls12"] }
```
acmorrow, line 89: Ditto why not 0.23 which is latest.

acmorrow (approved):
LGTM, though I'd like to know why we are upgrading to not-quite-latest. A fine answer to those questions is "because it doesn't work to do so".

Cargo.toml
npmenard: it clashes with viam-rust-utils which is stuck in the past because of the bug with DTLS padding :P

Cargo.toml
npmenard: same as above

PR #201, opened 2024-05-08
[RSDK-7387] Drop lifetime from GrpcClient, AppClient, and ViamServer

acmorrow:
A small simplification: drops a lifetime out of three important structs, so it seems worthwhile for the small memory cost of holding the URI in an owned string.


npmenard (approved):
LGTM

PR #203, opened 2024-05-09
stabilize server behavior on network interruptions

gvaradarajan:
Introduces the following behavioral changes:
- at the beginning of `main_esp32` in `esp32-server.rs` and `esp32-server-with-cred`, there is now a continuous loop for connecting to Wi-Fi rather than simply panicking when unsuccessful
- in `ViamServer::serve`, if the base station remains connected but there is not internet access, the loop now continues without attempting to connect signaling
- in `ViamServer::serve`, if the base station is disconnected, the `app_client` is dropped until network connection has been re-established
- we now attempt reconnection upon receipt of a station disconnected event (while the station is disconnected, the event is repeatedly fired on what seems to be a safe interval)

**NOTE: The templates have not been adjusted because they point to the last released version. There should be a ticket to adjust them accordingly upon the next major version release.** It will be linked here when created

### DIFF
--- a/examples/esp32-with-cred/esp32-server-with-cred.rs
+++ b/examples/esp32-with-cred/esp32-server-with-cred.rs
@@ -177,14 +177,20 @@ mod esp32 {
         #[allow(clippy::redundant_clone)]
         #[cfg(not(feature = "qemu"))]
         let network = {
-            let mut wifi =
-                Esp32WifiNetwork::new(sys_loop_stack.clone(), SSID.to_string(), PASS.to_string())
-                    .unwrap();
+            let mut wifi = Esp32WifiNetwork::new(
+                sys_loop_stack.clone(),
+                nvs_vars.wifi_ssid.clone(),
+                nvs_vars.wifi_pwd.clone(),
+            )
+            .unwrap();
             loop {
                 match wifi.connect() {
                     Ok(()) => break,
                     Err(_) => {
-                        log::info!("wifi could not connect to SSID: {:?}, retrying...", SSID);
+                        log::info!(
+                            "wifi could not connect to SSID: {:?}, retrying...",
+                            nvs_vars.wifi_ssid
+                        );
                         std::thread::sleep(std::time::Duration::from_millis(300));
                     }
                 }
--- a/examples/esp32-with-cred/esp32-server-with-cred.rs
+++ b/examples/esp32-with-cred/esp32-server-with-cred.rs
@@ -182,7 +182,7 @@ mod esp32 {
                 nvs_vars.wifi_ssid.clone(),
                 nvs_vars.wifi_pwd.clone(),
             )
-            .unwrap();
+            .expect("could not configure wifi");
             loop {
                 match wifi.connect() {
                     Ok(()) => break,
--- a/examples/esp32-with-cred/esp32-server-with-cred.rs
+++ b/examples/esp32-with-cred/esp32-server-with-cred.rs
@@ -25,6 +25,7 @@ mod esp32 {
                 board::FakeBoard,
                 robot::{LocalRobot, ResourceMap, ResourceType},
             },
+            esp32::conn::network::eth_configure,
             proto::common::v1::ResourceName,
         },
         std::{
@@ -36,14 +37,8 @@ mod esp32 {
 
     #[cfg(not(feature = "qemu"))]
     use {
-        embedded_svc::wifi::{
-            AuthMethod, ClientConfiguration as WifiClientConfiguration,
-            Configuration as WifiConfiguration,
-        },
         micro_rdk::common::registry::ComponentRegistry,
-        micro_rdk::esp32::esp_idf_svc::hal::{peripheral::Peripheral, prelude::Peripherals},
-        micro_rdk::esp32::esp_idf_svc::sys::esp_wifi_set_ps,
-        micro_rdk::esp32::esp_idf_svc::wifi::{BlockingWifi, EspWifi},
+        micro_rdk::esp32::conn::network::Esp32WifiNetwork,
     };
 
     #[derive(Debug, Error)]
@@ -52,8 +47,6 @@ mod esp32 {
         NVSKeyError(String),
         #[error("{0}")]
         EspError(EspError),
-        #[error("Error obtaining peripherals")]
-        PeripheralsError,
     }
 
     impl From<EspError> for ServerError {
@@ -123,11 +116,6 @@ mod esp32 {
         micro_rdk::esp32::esp_idf_svc::log::EspLogger::initialize_default();
         let sys_loop_stack = EspSystemEventLoop::take().unwrap();
 
-        #[cfg(not(feature = "qemu"))]
-        let periph = Peripherals::take()
-            .map_err(|_| ServerError::PeripheralsError)
-            .unwrap();
-
         #[cfg(feature = "qemu")]
         let repr = {
             let board = Arc::new(Mutex::new(FakeBoard::new(vec![])));
@@ -157,7 +145,7 @@ mod esp32 {
         let nvs_vars = NvsStaticVars::new().unwrap();
 
         #[cfg(feature = "qemu")]
-        let (ip, _block_eth) = {
+        let network = {
             use micro_rdk::esp32::esp_idf_svc::hal::prelude::Peripherals;
             info!("creating eth object");
             let mut eth = Box::new(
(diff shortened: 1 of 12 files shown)


examples/esp32-with-cred/esp32-server-with-cred.rs
```diff
@@ -190,22 +177,21 @@ mod esp32 {
...
+        let network = {
+            let wifi = loop {
+                match Esp32WifiNetwork::new(
+                    sys_loop_stack.clone(),
+                    &nvs_vars.wifi_ssid,
+                    &nvs_vars.wifi_pwd,
+                ) {
+                    Ok(wifi) => break wifi,
+                    Err(_) => {
+                        log::info!("wifi could not connect, retrying...");
```
acmorrow, line 189: This is nice in that now we won't crash if WiFI isn't there. But it is still the case that we can't come up if wifi is down. Should we be able to? We could still, for instance, be collecting sensor data in that state. We'd need a cached config though, which I don't think we have yet. Do we have plans/tickets on how to get there?

templates/project/src/main.rs
```diff
@@ -83,7 +83,7 @@ fn main() {
     let cfg = AppClientConfig::new(
         ROBOT_SECRET.to_owned(),
         ROBOT_ID.to_owned(),
-        ip,
+        wifi,
```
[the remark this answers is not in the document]
acmorrow: Nevermind, I missed your comment in the description about the templates not being updated yet.

examples/esp32-with-cred/esp32-server-with-cred.rs
npmenard, line 189: nit: add the SSID in the formatting

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -0,0 +1,134 @@
...
+#[cfg(feature = "qemu")]
+fn eth_configure<'d, T>(
+    sl_stack: &EspSystemEventLoop,
+    eth: EspEth<'d, T>,
+) -> Result<Box<BlockingEth<EspEth<'d, T>>>, EspError> {
+    let mut eth = BlockingEth::wrap(eth, sl_stack.clone())?;
+    eth.start()?;
+    eth.wait_netif_up()?;
+
+    log::info!("ETH IP {:?}", eth.eth().netif().get_ip_info().unwrap().ip);
```
npmenard, line 122: can remove this log line

npmenard (changes requested):
Done with first pass. Looks great I tried it quickly and it works like a charm :P awesome work

micro-rdk/src/common/conn/server.rs
```diff
@@ -305,30 +317,57 @@ where
...
+                        if !self.network.is_connected().unwrap_or(false) {
+                            // the IP may have changed (such as in the case where we are disconnected
+                            // from a Wi-Fi base station), so we want to prompt the creation of a new
+                            // app client
+                            let _ = self.app_client.take();
+                        }
+                        continue;
+                    }
+                };
+            }
```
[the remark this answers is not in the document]
gvaradarajan: I'll try testing it again, but I found it took much longer for that function to fail than simply pinging app directly (maybe because of the async part?)

micro-rdk/src/esp32/entry.rs
```diff
@@ -30,16 +32,22 @@ use super::{
     tls::{Esp32TLS, Esp32TLSServerConfig},
 };
 
+use crate::common::conn::network::Network;
 use async_io::Timer;
 
+#[cfg(not(feature = "qemu"))]
+type Esp32NetworkType = crate::esp32::conn::network::Esp32WifiNetwork;
+#[cfg(feature = "qemu")]
+type Esp32NetworkType = Box<BlockingEth<EspEth<'static, OpenEth>>>;
```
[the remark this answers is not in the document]
gvaradarajan: I can move these to that file for sure, but it wouldn't end up reducing the number of cfg blocks. Is that still ok?

examples/esp32-with-cred/esp32-server-with-cred.rs
gvaradarajan: I was planning on making that adjustment when introducing the data task

micro-rdk/src/common/conn/server.rs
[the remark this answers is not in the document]
gvaradarajan: ok yeah, I've reworked it a little. Let me know what you think.

micro-rdk/src/esp32/entry.rs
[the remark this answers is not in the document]
gvaradarajan: actually, I ended up removing these types completely

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -0,0 +1,134 @@
...
+/// A wrapper around the wifi structure available in esp-idf-svc with and adjustment to support
+/// reconnection
+#[cfg(not(feature = "qemu"))]
+pub struct Esp32WifiNetwork {
+    inner: Arc<Mutex<Box<BlockingWifi<EspWifi<'static>>>>>,
+    sl_stack: EspSystemEventLoop,
+    _subscription: Option<EspSubscription<'static, System>>,
+}
+
+#[cfg(not(feature = "qemu"))]
```
npmenard, line 35: don't forget to remove qemu feature inside micro-rdk

micro-rdk/src/esp32/entry.rs
```diff
@@ -32,14 +35,19 @@ use super::{
 
 use async_io::Timer;
 
+// #[cfg(not(feature = "qemu"))]
+// type Esp32NetworkType = crate::esp32::conn::network::Esp32WifiNetwork;
+// #[cfg(feature = "qemu")]
```
npmenard, line 40: remove :)

micro-rdk/src/esp32/entry.rs
```diff
@@ -32,14 +35,19 @@ use super::{
...
+
 pub async fn serve_web_inner(
     app_config: AppClientConfig,
     _tls_server_config: Esp32TLSServerConfig,
     repr: RobotRepresentation,
-    _ip: Ipv4Addr,
     webrtc_certificate: WebRtcCertificate,
     exec: Esp32Executor,
     max_webrtc_connection: usize,
+    network: impl Network,
```
npmenard, line 50: cool

npmenard (approved):
I left another handful of comments
LGTM otherwise

acmorrow (approved):
LGTM

micro-rdk/src/esp32/conn/network.rs
gvaradarajan: oh I misinterpreted what you had asked. Yeah will do

micro-rdk/src/native/entry.rs
```diff
@@ -233,7 +240,7 @@ pub fn serve_with_provisioning<S>(
     let mut last_error = None;
     let (app_config, tls_server_config) = loop {
         match cloned_exec.block_on(serve_provisioning_async(
-            ip,
+            network.get_ip(),
```
acmorrow, line 243: Similar thoughts here as above in this file. In particular, this comment from `ExternallyManagedNetwork`:
```
    // TODO: provide a way for an external managed network to communicate a change in IP
    // address
```

When that happens, will we remember to revisit the code here?

micro-rdk/src/esp32/entry.rs
```diff
@@ -263,7 +263,7 @@ pub fn serve_with_provisioning<S>(
     let mut last_error = None;
     let (app_config, tls_server_config) = loop {
         match cloned_exec.block_on(Box::pin(serve_provisioning_async(
-            ip,
+            network.get_ip(),
```
acmorrow, line 266: What if we don't have one yet / it changes?

acmorrow (approved):
LGTM. I have some questions and observations, but I'm also fine with the answers to those questions and observations being to defer them into tickets or that they are already on the roadmap.

gvaradarajan:
In agreement with @acmorrow, two follow-ups are required:
1. A way for code using micro-RDK as a library to [communicate a change in IP address](https://viam.atlassian.net/browse/RSDK-7629) if it is taking responsibility for managing internet connection (represented a TODO comment in the PR)
2. Instances of MDNS (`NativeMdns` and `EspMdns`) [should take](https://viam.atlassian.net/browse/RSDK-7628) an implementer of the `Network` trait introduced in this PR rather than a static IP address

mattjperez (commented):
Looks great!

Just nits about unwraps we talked about offline.

PR #204, opened 2024-05-13
[RSDK-6253] Update a device's image while retaining credentials data

mattjperez:
tested with local app image on running device:
`cargo r -p micro-rdk-installer -- update-app-image --binary-path target/xtensa-esp32-espidf/esp32-server.bin`

The `update-app-image` command does the following in sequence:
- read the partition table of a currently running device
- download/read the new/selected image into memory
- compare the partition tables between the old app image and new one
- if they do not match, prompt user to re-build and flash device from scratch
- overwrite the running device's app image with the new one

Modifications are restricted to just the factory partition, leaving pre-existing NVS data intact.

### DIFF
--- a/micro-rdk-installer/Cargo.toml
+++ b/micro-rdk-installer/Cargo.toml
@@ -34,3 +34,4 @@ thiserror.workspace = true
 tokio.workspace = true
 viam-rust-utils.workspace = true
 viam.workspace = true
+esp-idf-part = "0.5.0"
--- a/micro-rdk-installer/Cargo.toml
+++ b/micro-rdk-installer/Cargo.toml
@@ -21,7 +21,6 @@ espflash.workspace = true
 gethostname.workspace = true
 local-ip-address.workspace = true
 log.workspace = true
-md5.workspace = true
 packed_struct.workspace = true
 rcgen.workspace = true
 reqwest.workspace = true
--- a/micro-rdk-installer/Cargo.toml
+++ b/micro-rdk-installer/Cargo.toml
@@ -33,4 +33,4 @@ thiserror.workspace = true
 tokio.workspace = true
 viam-rust-utils.workspace = true
 viam.workspace = true
-esp-idf-part = "0.5.0"
+esp-idf-part.workspace = true
--- a/micro-rdk-installer/src/main.rs
+++ b/micro-rdk-installer/src/main.rs
@@ -321,15 +321,15 @@ fn main() -> Result<(), Error> {
         }
         Some(Commands::WriteFlash(args)) => {
             let config = Config::load().map_err(|err| Error::SerialConfigError(err.to_string()))?;
-            let tmp_dir = tempfile::Builder::new()
-                .prefix("micro-rdk-bin")
-                .tempdir()
-                .map_err(Error::FileError)?;
+            let tmp_path = tempfile::NamedTempFile::new()
+                .map_err(Error::FileError)?
+                .path()
+                .to_path_buf();
             let app_path = match &args.flash_args.bootloader {
                 Some(path) => PathBuf::from(path),
                 None => {
                     let rt = Runtime::new().map_err(Error::AsyncError)?;
-                    rt.block_on(download_micro_rdk_release(&tmp_dir, args.version.clone()))?
+                    rt.block_on(download_micro_rdk_release(&tmp_path, args.version.clone()))?
                 }
             };
             let nvs_metadata = read_nvs_metadata(app_path.clone())?;
@@ -374,8 +374,7 @@ fn main() -> Result<(), Error> {
 fn update_app_image(args: &AppImageArgs) -> Result<(), Error> {
     let config = Config::load().map_err(|err| Error::SerialConfigError(err.to_string()))?;
 
-    let dir = tempfile::tempdir().map_err(Error::FileError)?;
-    let tmp_old = dir.path().join("running-ptable.img");
+    let tmp_old = tempfile::NamedTempFile::new().map_err(Error::FileError)?;
 
     log::info!("Retrieving running partition table");
     let mut flasher =
@@ -386,7 +385,7 @@ fn update_app_image(args: &AppImageArgs) -> Result<(), Error> {
             PARTITION_TABLE_SIZE,
             DEFAULT_BLOCK_SIZE,
             DEFAULT_MAX_IN_FLIGHT,
-            tmp_old.clone(),
+            tmp_old.path().to_path_buf(),
         )
         .map_err(|_| Error::FlashConnect)?;
 
@@ -395,10 +394,10 @@ fn update_app_image(args: &AppImageArgs) -> Result<(), Error> {
         .map_err(|e| Error::PartitionTableError(e.to_string()))?;
 
     log::info!("Retrieving new image");
-    let tmp_new = tempfile::Builder::new()
-        .prefix("micro-rdk-bin")
-        .tempdir()
-        .map_err(Error::FileError)?;
+    let tmp_new = tempfile::NamedTempFile::new()
+        .map_err(Error::FileError)?
+        .path()
+        .to_path_buf();
     let app_path_new = match &args.flash_args.bootloader {
         Some(path) => PathBuf::from(path),
         None => {
--- a/micro-rdk-installer/src/main.rs
+++ b/micro-rdk-installer/src/main.rs
@@ -401,18 +401,29 @@ fn update_app_image(args: &AppImageArgs) -> Result<(), Error> {
         }
     };
 
-    let mut new_ptable_buf = Vec::with_capacity(PARTITION_TABLE_SIZE.into());
+    let mut new_ptable_buf = vec![0; PARTITION_TABLE_SIZE.into()];
     log::info!("Extracting new partition table");
-    let app_file_new = OpenOptions::new()
+    let mut app_file_new = OpenOptions::new()
         .read(true)
         .write(true)
         .open(app_path_new.clone())
         .map_err(Error::FileError)?;
 
-    let _num_read = app_file_new
-        .read_at(&mut new_ptable_buf, PARTITION_TABLE_ADDR.into())
+    let file_len = app_file_new.metadata().map_err(Error::FileError)?.len();
+    log::info!("new app length: {}", file_len);
+    if file_len < PARTITION_TABLE_SIZE.into() {
+        log::error!("file length is less than partition size")
+    }
+
+    app_file_new
+        .seek(SeekFrom::Start(PARTITION_TABLE_ADDR.into()))
+        .unwrap();
+    app_file_new
+        .read_exact(&mut new_ptable_buf)
         .map_err(Error::FileError)?;
-    let new_ptable = PartitionTable::try_from_bytes(new_ptable_buf.clone()).unwrap();
+    println!("{:?}", new_ptable_buf);
+    let new_ptable = PartitionTable::try_from_bytes(new_ptable_buf.clone())
+        .map_err(|_| Error::PartitionTableError)?;
 
     // Compare partition tables
     if old_ptable != new_ptable {
--- a/micro-rdk-installer/src/main.rs
+++ b/micro-rdk-installer/src/main.rs
@@ -269,7 +269,6 @@ fn init_logger() {
 fn main() -> Result<(), Error> {
     init_logger();
     let cli = Cli::parse();
-    let config = Config::load().map_err(|err| Error::SerialConfigError(err.to_string()))?;
     match &cli.command {
         Some(Commands::WriteCredentials(args)) => {
             let app_path = PathBuf::from(args.binary_path.clone());
--- a/micro-rdk-installer/src/main.rs
+++ b/micro-rdk-installer/src/main.rs
@@ -28,6 +28,8 @@ use tokio::runtime::Runtime;
 
 const PARTITION_TABLE_ADDR: u16 = 0x8000;
 const PARTITION_TABLE_SIZE: u16 = 0xc00;
+const EMPTY_BYTE: u8 = 0xFF;
+const APP_IMAGE_PARTITION_NAME: &str = "factory";
 
 #[derive(Deserialize, Debug)]
 struct AppCloudConfig {
@@ -410,9 +412,9 @@ fn update_app_image(args: &AppImageArgs) -> Result<(), Error> {
         .map_err(Error::FileError)?;
 
     let file_len = app_file_new.metadata().map_err(Error::FileError)?.len();
-    log::info!("new app length: {}", file_len);
     if file_len < PARTITION_TABLE_SIZE.into() {
-        log::error!("file length is less than partition size")
+        log::error!("file length is less than partition size");
+        return Err(Error::PartitionTableError);
     }
 
     app_file_new
@@ -421,7 +423,6 @@ fn update_app_image(args: &AppImageArgs) -> Result<(), Error> {
     app_file_new
         .read_exact(&mut new_ptable_buf)
         .map_err(Error::FileError)?;
-    println!("{:?}", new_ptable_buf);
     let new_ptable = PartitionTable::try_from_bytes(new_ptable_buf.clone())
         .map_err(|_| Error::PartitionTableError)?;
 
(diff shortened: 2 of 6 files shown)


micro-rdk-installer/src/main.rs
```diff
@@ -269,6 +298,121 @@ fn main() -> Result<(), Error> {
...
+            }
+
+            // get nvs data from the old binary
+            let dir = tempfile::tempdir().unwrap();
+            let nvs_path = dir.path().join("nvs.data");
+            log::info!("extracting nvs data from device");
+            // let mut flasher = connect(&args.connect_args, &config, false, false).unwrap();
+            flasher
+                .read_flash(nvs_offset, nvs_size, 0x1000, 64, nvs_path.clone())
+                .map_err(Error::EspFlashError)?;
```
mattjperez, line 390: stuck here for some reason. keep getting a `(CorruptData(4096, 4039)`, where `4096` is the read block size, and `4039` is the size of the chunk that was read.

This step is for retrieving the actual NVS data and not just partition info.
Trying to read the full image at the start (before parsing the running partition table) fails with this error as well.

mattjperez: Note: related? 
https://github.com/espressif/esptool/issues/823#issue-1531203064

npmenard: is the stub properly loaded? https://docs.espressif.com/projects/esptool/en/latest/esp32/esptool/flasher-stub.html

micro-rdk-installer/src/main.rs
```diff
@@ -269,6 +298,121 @@ fn main() -> Result<(), Error> {
...
+
+            log::info!("Writing nvs data to new image");
+            let mut nvs_file = OpenOptions::new()
```
npmenard, line 393: if you are just replacing the app image then you might not need to extract the NVS partition and just erase the section of flash containing the app.

mattjperez: you're right.  `espflash::Flasher::write_bin_to_flash` worked.

micro-rdk-installer/src/main.rs
```diff
@@ -214,13 +239,18 @@ fn write_credentials_to_app_binary(
...
+fn flash(
+    flash_args: FlashArgs,
+    monitor_args: Option<MonitorArgs>,
+    config: &Config,
+    app_path: PathBuf,
+) -> Result<(), Error> {
     log::info!("Connecting...");
     let mut flasher = connect(
-        &args.monitor_args.connect_args,
+        &monitor_args.clone().unwrap().connect_args,
```
gvaradarajan, line 250: I see we're assuming `monitor_args` is always some `flash_args.monitor` is true. Instead of assuming, could we check that we're in a valid state regarding these two settings (and return an error if we're not)?

gvaradarajan (changes requested):
LGTM, waiting for resolution of the two comments in this review and for me to give it a quick test

micro-rdk-installer/src/main.rs
mattjperez: actually, it might not be necessary anymore for `MonitorArgs` to be `Option`. checking now

micro-rdk-installer/src/main.rs
```diff
@@ -331,3 +367,101 @@ fn main() -> Result<(), Error> {
...
+    let mut new_ptable_buf = vec![0; PARTITION_TABLE_SIZE.into()];
+    log::info!("Extracting new partition table");
+    let mut app_file_new = OpenOptions::new()
+        .read(true)
+        .write(true)
+        .open(app_path_new.clone())
+        .map_err(Error::FileError)?;
+
+    let file_len = app_file_new.metadata().map_err(Error::FileError)?.len();
+    if file_len < PARTITION_TABLE_SIZE.into() {
```
npmenard, line 416: looks weird to me shouldn't it be PARTITION_TABLE_SIZE+PARTITION_TABLE_ADDR?

micro-rdk-installer/src/main.rs
```diff
@@ -331,3 +367,101 @@ fn main() -> Result<(), Error> {
...
+        app_size
+    );
+    let mut app_segment = vec![EMPTY_BYTE; app_size as usize];
+    // write just this data
+    app_file_new
+        .read_at(&mut app_segment, app_offset.into())
+        .map_err(Error::FileError)?;
+    log::info!("Writing new app segment to flash");
+    flasher
+        .write_bin_to_flash(app_offset, &app_segment, None)
```
npmenard, line 463: add progress

micro-rdk-installer/src/main.rs
mattjperez: you're right, changed.

gvaradarajan (approved):
LGTM

micro-rdk-installer/src/main.rs
```diff
@@ -214,13 +239,18 @@ fn write_credentials_to_app_binary(
     Ok(())
 }
 
-fn flash(args: WriteFlashArgs, config: &Config, app_path: PathBuf) -> Result<(), Error> {
+fn flash(
+    flash_args: FlashArgs,
+    monitor_args: MonitorArgs,
```
npmenard, line 244: Why do we pass MonitorArgs now? what's the upside? how does the new invocation looks like?

mattjperez: many of `espflash`'s method's, such as `Flasher::connect`, take these structs as parameters. Sadly, because the structs are all annotated as [`non-exhaustive`](https://github.com/esp-rs/espflash/blob/890f17ac46ea8d579118ae5a726c3a940ede9065/espflash/src/cli/mod.rs#L53), they can't just be initialized directly within our functions (ex `let m_args = MonitorArgs{..}`). 

This means that the only way we can create these input arguments via clap.
`cargo-espflash` highlights this in that it also passes all there structs directly or with a thin wrapper.
In addition, we have to be mindful about what `espflash` Commands we wrap in our own because the flattened fields of all inner commands must be unique, which sadly isn't always the case. `espflash` as a library isn't very ergonomic.

The new invocation is `cargo r -p micro-rdk-installer -- update-app-image --bootloader target/xtensa-esp32-espidf/esp32-server.bin`, with the breaking change being the `--bootloader` flag to specify the binary. Bumping `espflash` to 3.0 was a breaking change so it seemed an opportune moment to also have breaking change for the installer.

micro-rdk-installer/Cargo.toml
```diff
@@ -34,3 +33,4 @@ thiserror.workspace = true
 tokio.workspace = true
 viam-rust-utils.workspace = true
 viam.workspace = true
+esp-idf-part = "0.5.0"
```
acmorrow, line 36: This should be pulled up to the workspace level `Cargo.toml` and then pulled in here via `esp-idf-part.workspace = true`.

micro-rdk-installer/src/main.rs
```diff
@@ -331,3 +367,105 @@ fn main() -> Result<(), Error> {
...
+
+    log::info!("Retrieving running partition table");
+    let mut flasher =
+        connect(&args.connect_args, &config, false, false).map_err(|_| Error::FlashConnect)?;
+    flasher
+        .read_flash(
+            PARTITION_TABLE_ADDR.into(),
+            PARTITION_TABLE_SIZE.into(),
+            0x1000,
+            64,
```
acmorrow, line 385: What is the origin of these values? Why are they correct? When, if ever, might they need to be updated? Let's add some comments or declare some named constants.

micro-rdk-installer/src/main.rs
```diff
@@ -331,3 +367,105 @@ fn main() -> Result<(), Error> {
     };
     Ok(())
 }
+
+fn update_app_image(args: &AppImageArgs) -> Result<(), Error> {
+    let config = Config::load().map_err(|err| Error::SerialConfigError(err.to_string()))?;
+
+    let dir = tempfile::tempdir().map_err(Error::FileError)?;
+    let tmp_old = dir.path().join("running-ptable.img");
```
acmorrow, line 375: Maybe better to not name the file and just use https://docs.rs/tempfile/latest/tempfile/fn.tempfile.html instead.

micro-rdk-installer/src/main.rs
```diff
@@ -331,3 +367,105 @@ fn main() -> Result<(), Error> {
...
+            tmp_old.clone(),
+        )
+        .map_err(|_| Error::FlashConnect)?;
+
+    let old_ptable_buf = fs::read(tmp_old).map_err(Error::FileError)?;
+    let old_ptable = PartitionTable::try_from_bytes(old_ptable_buf.clone())
+        .map_err(|e| Error::PartitionTableError(e.to_string()))?;
+
+    log::info!("Retrieving new image");
+    let tmp_new = tempfile::Builder::new()
```
acmorrow, line 395: Similar question re why not just `tempfile::tempfile`

micro-rdk-installer/src/main.rs
```diff
@@ -331,3 +367,105 @@ fn main() -> Result<(), Error> {
...
+        .prefix("micro-rdk-bin")
+        .tempdir()
+        .map_err(Error::FileError)?;
+    let app_path_new = match &args.flash_args.bootloader {
```
acmorrow, line 399: Why `bootloader` here? Not saying that it is wrong, just a little surprising.

micro-rdk-installer/src/main.rs
```diff
@@ -331,3 +367,105 @@ fn main() -> Result<(), Error> {
...
+            let rt = Runtime::new().map_err(Error::AsyncError)?;
+            rt.block_on(download_micro_rdk_release(&tmp_new, args.version.clone()))?
+        }
+    };
+
+    let mut new_ptable_buf = vec![0; PARTITION_TABLE_SIZE.into()];
+    log::info!("Extracting new partition table");
+    let mut app_file_new = OpenOptions::new()
+        .read(true)
+        .write(true)
```
acmorrow, line 411: Why is it `write` as well?

micro-rdk-installer/src/main.rs
```diff
@@ -331,3 +367,105 @@ fn main() -> Result<(), Error> {
...
+        .map_err(Error::FileError)?;
+    app_file_new
+        .read_exact(&mut new_ptable_buf)
+        .map_err(Error::FileError)?;
+    let new_ptable = PartitionTable::try_from_bytes(new_ptable_buf.clone())
+        .map_err(|e| Error::PartitionTableError(e.to_string()))?;
+
+    // Compare partition tables
+    if old_ptable != new_ptable {
+        log::error!("rebuild and flash micro-rdk from scratch");
```
acmorrow, line 433: Do you need to log this if you are returning the below `Err`? If you still think logging is useful, I'd suggest including the "why" here as well, like you do for the returned Err

micro-rdk-installer/src/main.rs
```diff
@@ -331,3 +367,105 @@ fn main() -> Result<(), Error> {
...
+        return Err(Error::PartitionTableError(
+            "old and new app image partition tables do not match".to_string(),
+        ));
+    }
+
+    log::info!("Partition tables match!");
```
acmorrow, line 439: If you are logging/erroring if they don't match, you probably don't need to log if they do

micro-rdk-installer/src/main.rs
```diff
@@ -331,3 +367,105 @@ fn main() -> Result<(), Error> {
...
+
+    let nvs_partition_info = new_ptable.find(APP_IMAGE_PARTITION_NAME).ok_or_else(|| {
```
acmorrow, line 441: The naming of this variable is confusing to me. Why is `nvs` making an appearance here if this is the APP_IMAGE_PARTITION_NAME?

micro-rdk-installer/src/main.rs
```diff
@@ -331,3 +367,105 @@ fn main() -> Result<(), Error> {
...
+        Error::PartitionTableError(format!(
+            "failed to find `{}` partition",
+            APP_IMAGE_PARTITION_NAME
+        ))
+    })?;
+    let app_offset = nvs_partition_info.offset();
+    let app_size = nvs_partition_info.size();
```
acmorrow, line 448: Similarly confused re app vs nvs here, maybe I'm just misunderstanding. If so, maybe add comments.

acmorrow (changes requested):

micro-rdk-installer/src/main.rs
mattjperez: It's just how espflash's clap structs have things named. Not a huge fan of it

micro-rdk-installer/src/main.rs
mattjperez: they're default values used in `espflash`. will add comments about it

micro-rdk-installer/src/main.rs
mattjperez: ok, so it seems `read_flash` needs a `Path` as input and the `File` created from tempfile doesn't have a path. TIL not all `File`s have paths, making them more akin to file references or descriptors.

micro-rdk-installer/src/main.rs
mattjperez: similar to above, except it's how our `download_micro_rdk_release` uses paths and not just something that takes `Read`.

micro-rdk-installer/src/main.rs
mattjperez: does not, was residual from previous logic. removing

micro-rdk-installer/src/main.rs
mattjperez: no, you're right. it's residual from when I was trying to extract the nvs and embed it into the new app image's nvs partition. will fix

micro-rdk-installer/src/main.rs
mattjperez: same as above

acmorrow (commented):
Code is looking pretty good, though I think there is a related temp file API that would let you avoid needing to come up with names of your own but still gives you paths.

I did note that it appears that the partition table offset can be adjusted via sdkconfig. I'm not sure what we ought to do about that. I'd be fine with the answer being "nothing" or "ticket it".

micro-rdk-installer/src/main.rs
```diff
@@ -24,6 +26,14 @@ use secrecy::Secret;
 use serde::Deserialize;
 use tokio::runtime::Runtime;
 
+const PARTITION_TABLE_ADDR: u32 = 0x8000;
```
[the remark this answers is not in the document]
mattjperez: An option would keep a list of possible values we use and iterate through it when trying to extract the partition table. 

Currently, even [`espflash` hardcodes it](https://github.com/viamrobotics/espflash/blob/15991b5c1184b06d2f243dad7479b3de661fe4d2/espflash/src/targets/mod.rs#L194).

A perfect solution would involve fetching/initializing the sdkconfig, parsing it, using that info when extracting; which may not match if the sdkconfig changed this default address between the old and new app images.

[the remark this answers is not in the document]
mattjperez: Either way, it's an optimization at best and new feature at worst. Better left to another ticket if we want to look closer at this.

micro-rdk-installer/src/main.rs
npmenard: feels counterintuitive to use `bootloader` to pass an app image via a bootloader flag we should add a `bin` or `app-bin` arg.

micro-rdk-installer/src/main.rs
mattjperez: renamed it to `binary_path`

micro-rdk-installer/src/main.rs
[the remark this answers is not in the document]
acmorrow: I'm fine with going forward with the hard-coded value give that espflash does the same.

acmorrow (approved):
This LGTM though I share @npmenard's concerns about the use of the term `bootloader`.

npmenard:
I can't invoke `write-flash` anymore seems like it's expecting `--elf` and `--log-format` or is another argment mandatory now? Seems like it's the case since c317551b7680f755224821a3adbbeb7d768e14c3 
also passing `--monitor` with `update-app-image` doesn't open a serial monitor after flashing is done

PR #205, opened 2024-05-14
RSDK-7387 Make AppClient shareable so other async tasks can use it

acmorrow:
Demonstrates sharing the app client between the main serve loop and a periodic task - in this case, checking against app for whether the machine should restart.

I think this approach works fairly well but I'd like to get thoughts from the team before I go further (comments, etc.). I'll leave some inline comments as well.

I recommend reading the two `entry.rs` files first to get a sense of what it looks like at the top level. Then read `restart_monitor.rs` to see how the `RestartMonitor` works when called, then see `AppClient::check_for_restart` for how that is backed. Then have a look at the changes in `server.rs` and the change from `connect_signaling` to `initiate_signaling` in `app_client.rs`.


micro-rdk/src/common/app_client.rs
```diff
@@ -285,10 +315,58 @@ impl AppClient {
...
+    }
 }
 
 impl Drop for AppClient {
     fn drop(&mut self) {
         log::debug!("dropping AppClient")
     }
 }
+
+pub trait PeriodicAppClientTask {
```
acmorrow, line 365: If this approach survives I will doc comment this trait. This is what @gvaradarajan will implement against to provide something that the `DataManager` can use to periodically sync.

micro-rdk/src/common/conn/server.rs
```diff
@@ -276,8 +304,9 @@ pub struct ViamServer<C, T, CC, D, L> {
     exec: Executor,
     app_connector: C,
     app_config: AppClientConfig,
-    app_client: Option<AppClient>,
+    app_client: Rc<futures_util::lock::Mutex<Option<AppClient>>>,
```
acmorrow, line 307: The key change: access to the AppClient is now Rc'd and async Mutex'd. Many async tasks can hold a handle to it, each can await having private access to it, and they all hold a consistent view of whether the AppClient managed by the main `serve` loop is currently populated or not.

micro-rdk/src/common/conn/server.rs
```diff
@@ -296,40 +325,81 @@ where
...
             app_config,
-            app_client: None,
+            app_client: Default::default(),
             incoming_connection_manager: IncomingConnectionManager::new(max_concurent_connections),
+            app_client_tasks,
         }
     }
     pub async fn serve(&mut self, robot: Arc<Mutex<LocalRobot>>) {
         let cloned_robot = robot.clone();
+
```
acmorrow, line 343: Transforms each element of the vector of boxed PeriodicAppClientTask implementors into an actual async task, and collects them into a vector so we don't need to detach.

micro-rdk/src/common/conn/server.rs
```diff
@@ -296,40 +325,81 @@ where
...
+            .drain(..)
+            .map(|mut task| {
+                let app_client = Rc::clone(&self.app_client);
+                self.exec.spawn(async move {
+                    let mut duration = task.get_default_period();
+                    loop {
+                        let _ = async_io::Timer::after(duration).await;
+                        let mut app_client_guard = app_client.lock().await;
+                        if app_client_guard.is_none() {
+                        } else {
```
acmorrow, line 355: This was lazy I'l clean it up

micro-rdk/src/common/conn/server.rs
```diff
@@ -296,40 +325,81 @@ where
...
+                            match task.invoke(app_client_guard.as_mut().unwrap()).await {
+                                Ok(None) => continue,
+                                Ok(Some(next_duration)) => {
+                                    duration = next_duration;
+                                }
+                                Err(e) => {
+                                    log::warn!(
+                                        "Periodic task {} failed with error {}",
+                                        task.name(),
+                                        e
```
acmorrow, line 365: Open to suggestions on what better to do than just log this.

micro-rdk/src/common/conn/server.rs
```diff
@@ -296,40 +325,81 @@ where
...
+                self.exec.spawn(async move {
+                    let mut duration = task.get_default_period();
+                    loop {
+                        let _ = async_io::Timer::after(duration).await;
+                        let mut app_client_guard = app_client.lock().await;
+                        if app_client_guard.is_none() {
+                        } else {
+                            match task.invoke(app_client_guard.as_mut().unwrap()).await {
+                                Ok(None) => continue,
+                                Ok(Some(next_duration)) => {
```
acmorrow, line 358: This machinery around duration tracking was mostly to serve what appears to be the intention of the restart endpoint to be able to reply with a different timeout for the next check. It certainly adds some complexity so I'd be OK finding another way.

micro-rdk/src/common/conn/server.rs
```diff
@@ -296,40 +325,81 @@ where
...
+
         loop {
             let _ = async_io::Timer::after(std::time::Duration::from_millis(300)).await;
 
-            if self.app_client.is_none() {
-                let conn = self.app_connector.connect().await.unwrap();
-                let cloned_exec = self.exec.clone();
-                let grpc_client = Box::new(
-                    GrpcClient::new(conn, cloned_exec, "https://app.viam.com:443")
+            async {
```
acmorrow, line 378: This doesn't actually need to be async, just a scope for the lock. I'll remove the async / await.

micro-rdk/src/common/restart_monitor.rs
```diff
@@ -0,0 +1,47 @@
...
+use futures_lite::Future;
+use std::pin::Pin;
+use std::time::Duration;
+
+pub struct RestartMonitor {
+    restart_hook: Option<Box<dyn FnOnce()>>,
+}
+
+impl RestartMonitor {
+    pub fn new(restart_hook: impl FnOnce() + 'static) -> Self {
```
acmorrow, line 11: I had to add this `'static` but I'm not sure why and I'd prefer not to have it.

micro-rdk/src/common/restart_monitor.rs
```diff
@@ -0,0 +1,47 @@
...
+        &'a mut self,
+        app_client: &'b mut AppClient,
+    ) -> Pin<Box<dyn Future<Output = Result<Option<Duration>, AppClientError>> + 'b>> {
+        Box::pin(async move {
+            match app_client.check_for_restart().await {
+                Ok(None) => {
+                    self.restart();
+                }
+                Ok(Some(duration)) => Ok(Some(duration)),
+                Err(e) => Err(e),
```
acmorrow, line 43: Is there a nicer way to write this?

micro-rdk/src/common/app_client.rs
```diff
@@ -165,10 +173,11 @@ impl AppClientBuilder {
     }
 }
 
+#[derive(Clone)]
```
acmorrow, line 176: This is here from an older approach where we were instead going to `Clone` the app client. It could probably be backed out.

micro-rdk/src/common/app_client.rs
```diff
@@ -218,6 +227,27 @@ impl AppClient {
         Ok(CertificateResponse::decode(r)?)
     }
 
+    pub(crate) fn initiate_signaling(
```
acmorrow, line 230: This no longer awaits completion of a call because of the long polling nature of `/Answer` as discussed in https://viam.atlassian.net/issues/RSDK-7575

micro-rdk/src/common/app_client.rs
```diff
@@ -285,10 +315,58 @@ impl AppClient {
...
+            const MIN_RESTART_DURATION: Duration = Duration::from_secs(1);
+            const DEFAULT_RESTART_DURATION: Duration = Duration::from_secs(5);
+            Ok(response
+                .restart_check_interval
+                .map_or_else(
+                    || DEFAULT_RESTART_DURATION,
+                    |d| {
+                        Duration::try_from(d).map_or_else(
+                            |e| DEFAULT_RESTART_DURATION,
+                            |d| d.max(MIN_RESTART_DURATION),
```
acmorrow, line 350: This can probably be simplified a bit so that each constant only needs to be used once.

micro-rdk/src/common/conn/server.rs
```diff
@@ -169,6 +175,27 @@ where
             app_connector: self.app_connector,
             app_config: self.app_config,
             max_connections: self.max_connections,
+            app_client_tasks: self.app_client_tasks,
+        }
+    }
+    pub fn with_periodic_app_client_task(
```
acmorrow, line 181: Continuing the builder pattern here is a little forced because there is no type change that needs to be accommodated, but it seemed best to just do as the other methods do.

micro-rdk/src/esp32/entry.rs
```diff
@@ -149,6 +150,9 @@ pub async fn serve_web_inner(
             max_webrtc_connection,
         )
         .with_webrtc(webrtc)
+        .with_periodic_app_client_task(Box::new(RestartMonitor::new(|| unsafe {
```
acmorrow, line 153: The payoff for this approach is that we can just dependency inject the esp2 specific restart in the esp32 `entry.rs`, and a different one in the native `entry.rs`.

micro-rdk/src/native/entry.rs
```diff
@@ -129,6 +130,7 @@ pub async fn serve_web_inner(
     let mut srv = ViamServerBuilder::new(mdns, cloned_exec, client_connector, app_config, 3)
         .with_http2(tls_listener, 12346)
         .with_webrtc(webrtc)
+        .with_periodic_app_client_task(Box::new(RestartMonitor::new(|| std::process::exit(0))))
```
acmorrow, line 133: Calling `exit` here isn't fully faithful to the intent of restart. We could try to get fancy and `exec` `/proc/self/exe` or something, but short of that I think just `exit`ing is best.

micro-rdk/src/common/restart_monitor.rs
acmorrow, line 11: I also would have liked to make `restart_hook` be `FnOnce -> !` but that doesn't seem to be supported yet.

micro-rdk/src/esp32/entry.rs
```diff
@@ -149,6 +150,9 @@ pub async fn serve_web_inner(
...
+            crate::esp32::esp_idf_svc::sys::esp_restart()
```
npmenard, line 154: if there is an easy way I would rather do a unwind of the server before restart so we avoid leaving connections open against app.viam.com

npmenard (approved):
LGTM couple of comments

micro-rdk/src/common/app_client.rs
acmorrow: My idea is that the data manager will produce an object supporting this trait and attach it to the `serve` loop. On each invocation it will push data to app.

micro-rdk/src/common/conn/server.rs
acmorrow: Good point about rw. There isn't an async rwlock in `futures_utils` but I could try https://docs.rs/async-rwlock/latest/async_rwlock/ if we are OK with picking up that dependency and it works out. I'll do that now and see how it goes.

micro-rdk/src/esp32/entry.rs
acmorrow: Let me think about how that might be achieved. I sense some trickiness about it. Would you be OK with moving forward with this approach for now and tackling clean unwind in a follow-up PR?

micro-rdk/src/common/conn/server.rs
acmorrow: Seems like the newer version is https://docs.rs/async-lock/3.3.0/async_lock/

micro-rdk/src/esp32/entry.rs
acmorrow: @npmenard - I'd like to hold off on this unless you feel strongly that we need this piece (which I agree with) before Gautham can integrate.

micro-rdk/src/common/conn/server.rs
```diff
@@ -296,40 +328,91 @@ where
...
+                        .await
+                        .insert(app_client);
+                }
             }
 
             let sig = if let Some(webrtc_config) = self.webrtc_config.as_ref() {
                 let ip = self.app_config.get_ip();
-                let signaling = self.app_client.as_mut().unwrap().connect_signaling();
+                let rguard = self.app_client.read().await;
+                let signaling = rguard.as_ref().unwrap().initiate_signaling();
```
gvaradarajan, line 415: let me know if this is too much for this commit, but I wonder if we can make the actual signaling server logic just another instance of `PeriodicAppClientTask`? It feels like it would be cleaner with the one glaring problem maybe being the complexity of also accomplishing the `app_client` invalidation in L. 475. Maybe `invoke` can also return whether the `app_client` needs to be invalidated?

micro-rdk/src/common/conn/server.rs
acmorrow: The thing I like less about that is that it opens up the possibility of using the API wrong, because then you could accidentally call `ViamServer::add_periodic_app_client_task` after `serve` had started to no effect. With it on the builder, there's no confusion, because `build` consumes the builder so you can't call `with_periodic_app_client_task` after `build`, and once you have built you have a `ViamServer` and you can't then do anything but call `serve` on it.

micro-rdk/src/common/conn/server.rs
acmorrow: Done

micro-rdk/src/common/conn/server.rs
acmorrow: I think it is an interesting idea and it goes in the direction I think this is headed of decomposing `serve` into services, but I don't think I want to undertake it before getting you unblocked. Would a ticket be OK?

acmorrow (commented):
@gvaradarajan - Replies to your comments...

micro-rdk/src/common/conn/server.rs
gvaradarajan: this is a great point, I agree with leaving it as is

micro-rdk/src/common/conn/server.rs
gvaradarajan: yeah, I think a ticket would be perfectly fine

micro-rdk/src/esp32/entry.rs
gvaradarajan: wanted to agree with looking into this (and also that it should be a follow-up for later)

gvaradarajan (approved):
LGTM, excited to make use of it!

PR #207, opened 2024-05-16
RSDK-7149 - implement resolution for analogs and remove get_board_status

gvaradarajan:


micro-rdk/src/esp32/analog.rs
```diff
@@ -42,4 +42,17 @@ impl<'a, const A: u32, T: ADCPin> AnalogReader<u16> for Esp32AnalogReader<'a, A,
...
+    fn resolution(&self) -> crate::common::analog::AnalogResolution {
+        // TODO: In esp32/board.rs we have selected the ADC_ATTEN_DB_11 setting
+        // for our ADC drivers, resulting in the values below (see docs on ESP-IDF for more info).
+        // If and when we make this configurable, this function's implementation will have to change
+        // as well
+
+        // all units are in mV
+        AnalogResolution {
+            min_range: 150.0,
+            max_range: 2450.0,
```
npmenard, line 54: true for the linear range for the ADC can actually measure higher than 2.45mv might make more sense to use `esp_adc_cal_raw_to_voltage` to feed min max and resolution (0-4095) into this struct.

npmenard: it's far from perfect since adc curve are not linear for error will accumulate especially outside the 150-2450 range

npmenard (changes requested):
Seems like we need to update the board status function too since they have changed the layout of Board.Status proto

micro-rdk/src/lib.rs
```diff
@@ -71,6 +71,11 @@ pub mod proto {
                 include!("gen/viam.app.datasync.v1.rs");
             }
         }
+        pub mod mltraining {
+            pub mod v1 {
+                include!("gen/viam.app.mltraining.v1.rs");
```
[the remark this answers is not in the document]
gvaradarajan: ^ yep

micro-rdk/src/esp32/analog.rs
gvaradarajan: sorry, I'm a little unclear here. That function seems to directly compute voltage, so do you mean I should replace the raw value with the value in mV? If so does it actually make sense to even return non-trivial values for the resolution?

micro-rdk/src/lib.rs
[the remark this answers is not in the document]
npmenard: annoying

micro-rdk/src/esp32/analog.rs
npmenard: The range in this struct is the suggested voltage range for esp32 for accurate voltage measurement where raw to volatge is mostly linear. That doesn't mean ADC is not able to measure voltage outside of this range. I don't know if the purpose of this api is to offload the conversion to SDKs but if it is the case then we should calcultage mon/max voltage by calling ` esp_adc_cal_raw_to_voltage(0,&chars)` for min and ` esp_adc_cal_raw_to_voltage(4095,&chars)` for min/max (can use the rust bindings for that too) and assume it's linear for the range (which is not) Would need to confirm the values are not too insane

Ideally we should return mV instead of raw

gvaradarajan: I agree with the last sentence, I wonder if we can just provide the voltage and abide by some standard for the existing resolution values?

npmenard: :ok_hand:

micro-rdk/src/common/analog.rs
```diff
@@ -35,12 +35,26 @@ impl AnalogReader<u16> for FakeAnalogReader {
     fn read(&mut self) -> Result<u16, Self::Error> {
         self.internal_read()
     }
+    fn resolution(&self) -> AnalogResolution {
+        Default::default()
+    }
+}
+
+#[derive(Debug, Default)]
```
mattjperez, line 43: do we want to be returning derived Default values here? We'd always be returning a `min_range`, `max_range`, and `step_size` of `0.0`. 

Would it not be better to return some "realistic" defaults that we can test the FakeAnalogReader against later?

mattjperez (commented):
Looks great! Just a question, but otherwise LGTM. Will let the others approve since there were pending modification requests.

micro-rdk/src/esp32/analog.rs
npmenard: if the returned value goes above 2450 what's happening on the SDK side?

npmenard (approved):
LGTM I am glad board status gets removed :P I see in the proto file there is still 
``` proto
message Status {
  map<string, int32> analogs = 1;
  map<string, int64> digital_interrupts = 2;
}
```
Do we know what's the purpose? Do we need to implement it 

Lastly how do we validate behavior? IRC Board Status was just returning Analog readers and DigitalInterrupt status (which we didn't implement?)

micro-rdk/src/common/analog.rs
gvaradarajan: yeah, sure I could throw in some numbers I guess

gvaradarajan:
> LGTM I am glad board status gets removed :P I see in the proto file there is still
> 
> ```proto
> message Status {
>   map<string, int32> analogs = 1;
>   map<string, int64> digital_interrupts = 2;
> }
> ```
> 
> Do we know what's the purpose? Do we need to implement it
> 
> Lastly how do we validate behavior? IRC Board Status was just returning Analog readers and DigitalInterrupt status (which we didn't implement?)

I checked RDK and that message was being used only internally and gets turned into a generic struct field for `RobotStatus`. I figured we didn't need the extra step.

I figured asking for `RobotStatus` through an SDK and seeing that the analogs are returned without error and in mV would be sufficient

gvaradarajan:
Tested on hardware

PR #209, opened 2024-05-20
[RSDK-7497] change entrypoints

npmenard:
With Wifi provisioning we can finally remove our reliance on pre fetching robot config and certificates during build or having to set the wifi configuration with the installer. 

The starting flow has be updated accordingly.
When building if a `viam.json` is present with credentials then the robot will attempt to use these credentials to instantiate the robot. When absent micro-rdk will start the provisioning server first for esp32 build wifi provisioning will be enabled for native build it won't be.

Using the NVS partition will keep things backward compatible with the installer so it can stiull be use to set the credentials if needed.


Makefile
```diff
@@ -60,10 +60,10 @@ endif
 	$(QEMU_ESP32_XTENSA)/qemu-system-xtensa -nographic -machine esp32 -gdb tcp::3334 -nic user,model=open_eth,hostfwd=udp::-:61205 -drive file=target/xtensa-esp32-espidf/debug/debug.bin,if=mtd,format=raw -S
 
 upload: cargo-ver
-	cargo +esp espflash flash --package examples --monitor --partition-table examples/esp32/partitions.csv --baud 460800 -f 80mhz --bin esp32-server --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort
+	cargo +esp espflash flash --package examples --monitor --partition-table examples/esp32/partitions.csv --baud 926100 -f 80mhz --bin esp32-server --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort
```
npmenard, line 63: note to me: change back to 460800

examples/build.rs
```diff
@@ -19,222 +10,36 @@ pub struct Config {
...
-        }
-        None => false,
-    };
+    //    let mut has_robot_credentials = false;
     if env::var("TARGET").unwrap() == "xtensa-esp32-espidf" {
         if std::env::var_os("IDF_PATH").is_none() {
-            return Err(anyhow::anyhow!(
-                "You need to run IDF's export.sh before building"
-            ));
+            panic!("You need to run IDF's export.sh before building");
```
npmenard, line 21: panic is fine here, errors where never that useful

examples/esp32/esp32-server.rs
```diff
@@ -65,69 +53,44 @@ mod esp32 {
...
-        #[cfg(feature = "provisioning")]
+        #[cfg(not(has_robot_config))]
         {
-            use micro_rdk::common::provisioning::{server::ProvisioningInfo, storage::RAMStorage};
+	    // Pass NVS
             let mut info = ProvisioningInfo::default();
             info.set_fragment_id("d385b480-3d19-4fad-a928-b5c18a58d0ed".to_string());
             info.set_manufacturer("viam".to_owned());
             info.set_model("test-esp32".to_owned());
             let storage = RAMStorage::default();
```
npmenard, line 78: once testing is done use NVS storage by default

examples/esp32/esp32-server.rs
```diff
@@ -65,69 +53,44 @@ mod esp32 {
...
-            micro_rdk::esp32::entry::serve_with_provisioning(
-                storage,
-                info,
-                repr,
-                network,
-                max_connection,
-            );
+            log::info!("Will provision");
```
npmenard, line 79: remove stray comments

examples/native/native-server.rs
```diff
@@ -16,32 +22,24 @@ mod native {
...
-
-            let app_config =
-                AppClientConfig::new(ROBOT_SECRET.to_owned(), ROBOT_ID.to_owned(), "".to_owned());
-
-            micro_rdk::native::entry::serve_web(app_config, cfg, repr, network);
+            if ROBOT_ID.is_some() && ROBOT_SECRET.is_some() {
+                let ram_storage = RAMStorage::new("", "", ROBOT_ID.unwrap(), ROBOT_SECRET.unwrap());
+                serve_web_with_external_network(None, repr, 3, ram_storage, network);
+            }
+            //TODO what?
```
npmenard, line 31: remove stray comment

examples/native/native-server.rs
```diff
@@ -16,32 +22,24 @@ mod native {
...
+            use micro_rdk::common::provisioning::server::ProvisioningInfo;
             let mut info = ProvisioningInfo::default();
             info.set_fragment_id("d385b480-3d19-4fad-a928-b5c18a58d0ed".to_string());
             info.set_manufacturer("viam".to_owned());
-            info.set_model("test".to_owned());
-            let storage = MemoryCredentialStorage::default();
-            micro_rdk::native::entry::serve_with_provisioning(storage, info, repr, ip);
+            info.set_model("test-esp32".to_owned());
+            let storage = RAMStorage::default();
+            log::info!("Will provision");
```
npmenard, line 41: remove log statement

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -266,9 +331,19 @@ where
...
-            self.credential_ready.done();
+
+        match self.wifi_manager.as_ref() {
+            Some(_) => {
+                if self.storage.has_wifi_credentials() {
+                    self.credential_ready.done()
+                }
+            }
+            None => {
+                log::error!("READY");
```
npmenard, line 342: remove log statement

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -391,6 +469,112 @@ impl WifiManager for () {
     }
 }
 
+#[cfg(feature = "native")]
```
npmenard, line 472: not too happy with this but I wanted to avoid duplicating code for each variant and using generics became quickly too contrived and unreadable.

micro-rdk/src/esp32/entry.rs
```diff
@@ -159,146 +177,204 @@ pub async fn serve_web_inner(
...
+                if let Err(error) = ret {
+                    log::info!(
+                        "Couldn't connect to {} cause : {:?}",
+                        storage.get_wifi_credentials().unwrap().ssid,
+                        error
+                    );
+                    continue;
+                }
+                // Assume connected to internet so any error should be forwarded to provisioning
+                if let Err(e) = validate_robot_credentials(
```
npmenard, line 247: nit : probably better expressed with map_or or equivalent

micro-rdk/src/esp32/entry.rs
```diff
@@ -114,7 +132,7 @@ pub async fn serve_web_inner(
             }
         };
 
-        (cfg_response, robot)
+        (cfg_response, robot, tls_server_config)
```
gvaradarajan, line 135: I've always wondered this but forgot to ask; why do we need this TLS config to still be in scope for the rest of this function?

micro-rdk/src/esp32/entry.rs
```diff
@@ -159,146 +177,204 @@ pub async fn serve_web_inner(
...
-                Esp32TLSServerConfig::new(tls_certs, serv_key, serv_key_len),
-            ));
+
+    use super::conn::network::Esp32WifiNetwork;
+
+    let mut last_error: Option<Box<dyn std::error::Error>> = None;
+
+    let info = info.unwrap_or_default();
+
+    let network = loop {
```
gvaradarajan, line 230: if we made the network and executor types generic parameters, could we actually put this function in common? Edit: In fact, I have this question about all the functions in this file

micro-rdk/src/esp32/entry.rs
```diff
@@ -159,146 +177,204 @@ pub async fn serve_web_inner(
...
+// Four cases:
+// 1) No Robot Credentials + WiFi without external network
+// 2) No Robot Credentials with external network\
+// 3) Robot Credentials with external network
+// 4) Robot Credentials + WiFi without external network
+
 #[cfg(feature = "provisioning")]
-async fn serve_provisioning_async<S>(
-    ip: std::net::Ipv4Addr,
+async fn serve_async<S>(
```
gvaradarajan, line 204: I think this function is complex enough that it now warrants a general doc comment summarizing its procedure. I like that the cases have been outlined, but a quick statement on the expected behavior with each case would clarify things. I feel this way about `serve_async_with_external_network` as well.

gvaradarajan (changes requested):
This essentially looks good to me, just a few things that might be purely my confusion

micro-rdk/src/common/provisioning/server.rs
npmenard: the issue is that multiple bounds have to be added to make executor work with HyperExec trait  and provisioningExec traits so we would end up with a lot of bounds which make things harder to read. At this point we can merge NativeExecutor and Esp32Excutor so we should be able to make thing a bit easier to use. But that for some future work

micro-rdk/src/esp32/conn/mdns.rs
```diff
@@ -37,3 +34,35 @@ impl Mdns for Esp32Mdns {
         Ok(())
     }
 }
+
+impl Mdns for Esp32Mdns {
+    fn add_service(
+        &mut self,
+        instance_name: &str,
+        service_type: impl AsRef<str>,
+        proto: impl AsRef<str>,
```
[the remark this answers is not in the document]
npmenard: tcp/udp and alike

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -266,9 +331,19 @@ where
             .map_err(|e| ServerError::new(GrpcError::RpcInternal, Some(e.into())))?;
         debug_assert_eq!(buffer.len(), 5 + len);
         debug_assert_eq!(buffer.capacity(), 5 + len);
-        if self.wifi_manager.is_some() && self.storage.has_wifi_credentials() {
-            self.credential_ready.done();
+
+        match self.wifi_manager.as_ref() {
```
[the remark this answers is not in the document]
npmenard: yes credential ready is meant as a flag to let the provisioning server that it should go away and test sent credentials

micro-rdk/src/esp32/entry.rs
npmenard: it is used buy the http2 server if configured in the Viam Server Build

micro-rdk/src/esp32/entry.rs
npmenard: It's possible but I think it would be best to write as a macro in the mean time, We can do it in a later PR. A full generic approach is challenging because we have many type that are very similar but slightly different. In the future we can start an effort to de-duplicate code between native and esp32 but out of scope for this PR

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -1,117 +1,136 @@
...
-            .expect("could not get IP info")
-            .ip
+        let guard = esp32_get_wifi().map_or(None, |wifi| wifi.try_lock());
+
+        guard.map_or(Ipv4Addr::UNSPECIFIED, |guard| {
+            guard
+                .wifi()
+                .sta_netif()
+                .get_ip_info()
+                .map_or(Ipv4Addr::UNSPECIFIED, |ip_info| ip_info.ip)
```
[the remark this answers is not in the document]
npmenard: that was how you implemented the trait. I think returning 0.0.0.0 works well in situation where the esp32 is connected but for some reason no ip are assigned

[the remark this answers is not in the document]
gvaradarajan: Yeah I know, I thought maybe I had made a mistake. But if the unspecified ip is more suitable, that's fine by me

micro-rdk/src/esp32/entry.rs
gvaradarajan: Yeah, fair enough

micro-rdk/src/common/provisioning/server.rs
gvaradarajan: Sounds reasonable to me

micro-rdk/src/native/conn/mdns.rs
```diff
@@ -68,3 +60,40 @@ impl Mdns for NativeMdns {
...
+        txt: &[(&str, &str)],
+    ) -> Result<(), MdnsError> {
+        self.add_service(instance_name, service_type, proto, port, txt)
+    }
+    fn set_hostname(&mut self, hostname: &str) -> Result<(), MdnsError> {
+        self.set_hostname(hostname)
+    }
+}
+
+impl Mdns for &mut NativeMdns {
```
acmorrow, line 85: Asking for my own understanding: why implement the trait both for the `NativeMdns` type and a mutable ref to it? It isn't a pattern I've seen before.

micro-rdk/src/native/entry.rs
```diff
@@ -145,132 +153,124 @@ pub async fn serve_web_inner(
...
+    let builder = AppClientBuilder::new(Box::new(client), app_config.clone());
+
+    let _client = builder.build().await?;
+    Ok(())
+}
+
 #[cfg(feature = "provisioning")]
-async fn serve_provisioning_async<S>(
-    ip: Ipv4Addr,
+async fn serve_async_with_external_network<S>(
```
acmorrow, line 176: This function and the one below are named `_with_external_network`, but they accept `impl Network` which isn't necessarily an ExternalNetwork.

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -27,24 +36,60 @@ use prost::Message;
...
-#[derive(Default)]
-pub(crate) struct ProvisioningServiceBuilder<Wifi = ()> {
+async fn dns_server(ap_ip: Ipv4Addr) {
+    let socket = async_io::Async::<UdpSocket>::bind(([0, 0, 0, 0], 53)).unwrap();
+    loop {
+        let mut buf = [0_u8; 512];
+        let len = socket.recv_from(&mut buf).await.unwrap();
+        let buf = Bytes::copy_from_slice(&buf[..len.0]);
+        let mut ans = dns_message_parser::Dns::decode(buf);
+        if let Ok(ref mut msg) = ans {
```
acmorrow, line 46: Didn't know about `ref mut`.

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -64,19 +109,32 @@ impl<Wifi: WifiManager> ProvisioningServiceBuilder<Wifi> {
...
         self,
         storage: S,
-    ) -> ProvisioningService<S, Wifi> {
+    ) -> ProvisioningService<S, Wifi>
+    where
+        Exec: ProvisioningExecutor,
+        Wifi: WifiManager,
+    {
+        // Provisioning relies on DNS query to find the IP of the server. Specifically it will
+        // make a request for viam.setup. All other queries are answered failed to express the lack off
```
acmorrow, line 131: `lack of`

acmorrow (approved):

gvaradarajan (approved):
LGTM

micro-rdk/src/native/conn/mdns.rs
npmenard: the function  `serve_provisioning_async` takes a `mut mdns: M` as a param where M implements Mdns. But we pass a mutable reference to a `NativeMdns` or `Esp32Mdns` to the function. Therefore the Mdns trait has to be implemented for &mut T.  (note that we have to dereference because the argument if the same as `&mut self: &mut NativeMdns` otherwise we would recurse)
There is another case when trait method takes self as a receiver but you what to implemented for &T only

PR #210, opened 2024-05-20
hotfix: `write-flash` command

mattjperez:
This fixes the issues with `write-flash` and `--monitor` mentioned [here](https://github.com/viamrobotics/micro-rdk/pull/204#issuecomment-2120965559).

command should now work as expected. 

Note: the previous app-image and new app-image must both be from the same variant of `examples`, ie `esp32-with-cred`.
Since we publish `esp32-with-cred`, if the old app image was build manually with `make build-esp32-bin`, the `monitor` will show an error such as 
```
I (16) esp32_server_with_cred::esp32: load vars from NVS...
I (32) esp32_server_with_cred::esp32: get namespace...
I (33) esp32_server_with_cred::esp32: loading creds...
I (34) esp_idf_svc::nvs: EspNvs dropped
I (38) esp_idf_svc::nvs: NvsDefault dropped
thread 'main' panicked at examples/esp32-with-cred/esp32-server-with-cred.rs:159:45:
called `Result::unwrap()` on an `Err` value: NVSKeyError("WIFI_SSID")
```

This should not happen once #209 is merged though.

### DIFF
--- a/micro-rdk-installer/src/main.rs
+++ b/micro-rdk-installer/src/main.rs
@@ -33,6 +33,7 @@ const APP_IMAGE_PARTITION_NAME: &str = "factory";
 // taken from `espflash::cli::ReadFlashArgs` default values
 const DEFAULT_BLOCK_SIZE: u32 = 0x1000;
 const DEFAULT_MAX_IN_FLIGHT: u32 = 64;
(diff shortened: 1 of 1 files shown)


npmenard (approved):
LGTM

PR #211, opened 2024-05-21
implement final adjustments for first version of data management

gvaradarajan:
This is to address two follow-ups to the review of the previous [PR](https://github.com/viamrobotics/micro-rdk/pull/208) integrating data management and server code

- setting a more generous limit to the size of static memory data storage
- chunking large amount of data retrieved from the store into multiple upload requests

### DIFF
--- a/micro-rdk/src/common/data_manager.rs
+++ b/micro-rdk/src/common/data_manager.rs
@@ -22,6 +22,11 @@ use futures_util::lock::Mutex as AsyncMutex;
 use prost::Message;
 use thiserror::Error;
 
+// Maximum size in bytes of readings that should be sent in a single request
+// as recommended by Viam's data management team is 64K. To accommodate for
+// the smaller amount of available RAM, we've halved it
+static MAX_SENSOR_CONTENTS_SIZE: usize = 32000;
+
 #[derive(Debug, Error)]
 pub enum DataManagerError {
     #[error("no data collectors in manager")]
@@ -222,9 +227,9 @@ where
 #[derive(Debug, Error)]
 pub enum DataSyncError {
     #[error(transparent)]
-    DataStoreError(#[from] DataStoreError),
-    #[error(transparent)]
     MessageDecodingError(#[from] prost::DecodeError),
+    #[error(transparent)]
+    UploadError(#[from] AppClientError),
 }
 
 pub struct DataSyncTask<StoreType> {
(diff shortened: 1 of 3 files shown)


micro-rdk/src/common/data_store.rs
```diff
@@ -28,7 +28,7 @@ impl Default for WriteMode {
     }
 }
 
-static mut DATA_STORE: [MaybeUninit<u8>; 10240] = [MaybeUninit::uninit(); 10240];
+static mut DATA_STORE: [MaybeUninit<u8>; 111834] = [MaybeUninit::uninit(); 111834];
```
mattjperez, line 31: can we move this number to a constant with a comment/link?

micro-rdk/src/common/data_manager.rs
```diff
@@ -22,6 +22,8 @@ use futures_util::lock::Mutex as AsyncMutex;
 use prost::Message;
 use thiserror::Error;
 
+static MAX_SENSOR_CONTENTS_SIZE: usize = 64000;
```
mattjperez, line 25: Why this number? Is it an esp32 constraint? something in app?

micro-rdk/src/common/data_manager.rs
```diff
@@ -241,25 +243,40 @@ where
...
+        let mut current_chunk: Vec<BytesMut> = vec![];
+        let mut chunk_size_counter: usize = 0;
         let mut reader = store_lock.get_reader(collector_key)?;
         loop {
             let next_message = reader.read_next_message()?;
             if next_message.is_empty() {
                 break;
             }
-            raw_messages.push(next_message);
+            // If the amount of data becomes two big for a single upload, we want to start a new
```
mattjperez, line 257: typo: `too`

mattjperez (approved):
LGTM. just some [non-blocking] questions

micro-rdk/src/common/data_manager.rs
acmorrow: Let's document it, and also consider cutting it to something smaller than the upper bound, like 32k or 16k.

micro-rdk/src/common/data_manager.rs
```diff
@@ -241,25 +244,40 @@ where
     async fn read_messages_for_collector(
         &self,
         collector_key: &ResourceMethodKey,
-    ) -> Result<Vec<SensorData>, DataSyncError> {
+    ) -> Result<Vec<Vec<SensorData>>, DataSyncError> {
         let store_lock = self.store.lock().await;
-        let mut raw_messages: Vec<BytesMut> = vec![];
+        let mut chunks: Vec<Vec<SensorData>> = vec![];
```
acmorrow, line 249: Per discussion, I think this isn't currently achieving the goal of keeping the memory usage for each upload under a static limit.

acmorrow (changes requested):

micro-rdk/src/common/data_manager.rs
```diff
@@ -234,54 +239,111 @@ pub struct DataSyncTask<StoreType> {
...
+                        log::error!("error reading from store: {:?}", err);
+                        break;
+                    }
+                };
+                if next_message.is_empty() {
+                    break;
+                }
+                // If the amount of data becomes too big for a single upload, we want to start a new
+                // collection of upload data
+                if (chunk_size_counter + next_message.len()) > MAX_SENSOR_CONTENTS_SIZE {
```
acmorrow, line 299: Per discussion, this will drop a message on overflow. Additionally, there is no protection against encountering an entry that exceeds the max upload size, which would stall indefinitely if the above issue was fixed.

acmorrow (changes requested):
@gvaradarajan and I talked this over in person, we are going to reduce priority on this temporarily while we mull over some solutions: it isn't blocking anything yet.

gvaradarajan:
closing this PR in favor of work on a different branch that accounts for an upstream refactor

PR #212, opened 2024-05-22
[RSDK-7708] Update template

npmenard:


mattjperez (approved):
This is great! Awesome seeing how much boilerplate we can remove after your work on provisioning.

LGTM

templates/project/Cargo.toml
```diff
@@ -18,23 +18,11 @@ opt-level = "z"
 embedded-svc = "0.27"
 embedded-hal = { version = "0.2.7", features = ["unproven"]}
 log = "0.4"
-async-channel = "1.8.0"
-futures-lite = "1.12.0"
-micro-rdk = {version = "0.1.8", git = "https://github.com/viamrobotics/micro-rdk.git", features = ["esp32", "binstart"], rev = "c669b06" }
+async-channel = "2"
+futures-lite = "1"
+micro-rdk = {version = "0.1.8", git = "https://github.com/viamrobotics/micro-rdk.git", features = ["esp32", "binstart","provisioning"], rev = "e5ba39f" }
```
acmorrow, line 23: The module template also has a `rev` of `c669b06` - I suspect it should get the same update.

templates/project/Cargo.toml
```diff
@@ -18,23 +18,11 @@ opt-level = "z"
...
 
 [build-dependencies]
 anyhow = "1"
-cargo_metadata = "0.18.1"
-const-gen = "1.3.0"
+cargo_metadata = "0.18"
```
acmorrow, line 27: ⬆️ - Do you want to drop the `anyhow` dependency here too?

templates/project/build.rs
```diff
@@ -1,277 +1,66 @@
...
-        const_declaration!(
-            #[allow(clippy::redundant_static_lifetimes, dead_code)]
-            ROBOT_SRV_DER_KEY = key
-        ),
-    ]
-    .join("\n");
-    fs::write(dest_path, robot_decl).unwrap();
 
-    let metadata = MetadataCommand::new()
+    let metadata = cargo_metadata::MetadataCommand::new()
```
acmorrow, line 11: I don't mind it, I'm just curious why you changed from `use cargo_metadata::{MetadataCommand, CargoOpt}` to long names for these types.

templates/project/sdkconfig.defaults
```diff
@@ -11,8 +11,8 @@ CONFIG_ESP_SYSTEM_EVENT_TASK_STACK_SIZE=4096
 CONFIG_PTHREAD_TASK_STACK_SIZE_DEFAULT=8192
 
 CONFIG_ESP32_DEFAULT_CPU_FREQ_240=y
-CONFIG_ESPTOOLPY_FLASHFREQ_80M=y
-CONFIG_ESPTOOLPY_FLASHMODE_DIO=y
+CONFIG_ESPTOOLPY_FLASHFREQ_80M=n
+CONFIG_ESPTOOLPY_FLASHMODE_DIO=n
```
acmorrow, line 15: I can't find a lot of meaningful documentation on what these mean. Can you clarify why the template should have this as `n`? The server still has these as `y` so they will diverge.

templates/project/sdkconfig.defaults
```diff
@@ -23,11 +23,11 @@ CONFIG_SPIRAM_SUPPORT=y
 CONFIG_ESP32_SPIRAM_SUPPORT=y
 CONFIG_ESP32S2_SPIRAM_SUPPORT=n
 CONFIG_ESP32S3_SPIRAM_SUPPORT=n
-CONFIG_SPIRAM_SPEED_80M=y
+CONFIG_SPIRAM_SPEED_80M=n
```
acmorrow, line 26: As above re `FLASH[FREQ,MODE]`

templates/project/cargo-generate.toml
```diff
@@ -9,17 +9,7 @@ prompt = "MCU"
...
-type = "string"
-prompt = "Enter wifi name (only 2.4 gHz network are supported)"
-default = "Viam-2G"
-
-[placeholders.pwd]
-type = "string"
-prompt = "Enter the wifi password (leave blank if you don't want to store it)"
-default = ""
+prompt = "Enter fragment ID"
+default = "d385b480-3d19-4fad-a928-b5c18a58d0ed"
```
acmorrow, line 15: Could we add a comment linking to `https://app.viam.com/fragment?id=d385b480-3d19-4fad-a928-b5c18a58d0ed` so it is easy to jump to the actual fragment definition from the source? Should the fragment actually represented by the URL?

acmorrow (changes requested):

templates/project/Cargo.toml
npmenard: yes!

templates/project/Cargo.toml
npmenard: done

templates/project/build.rs
npmenard: I was lazy :P

templates/project/cargo-generate.toml
npmenard: We can maybe give more context? I added it here so provisioning has a default config to be set. But in reality we need more discussion about what a default fragment should look like if we want any.
For now Viam app doesn't seem to honor fragment id

acmorrow (approved):
LGTM!

PR #214, opened 2024-05-24
[RSDK-7717] windows micro-rdk-installer 

mattjperez:
This adds necessary changes to the source code needed to support windows builds, ~however this does not include the CI changes necessary to build and publish it.~ including changes to CI to build and publish.

~you can use [cross](https://github.com/cross-rs/cross) to compile with the command `cross build --target x86_64-pc-windows-gnu`~

confirmed locally you can use canon with micro-rdk-dev image via

```sh
$ sudo apt update
$ sudo apt install gcc-mingw-w64
$ cargo build -p micro-rdk-installer --target x86_64-pc-windows-gnu`
```


npmenard (changes requested):
done with first pass

PR #215, opened 2024-05-28
Add installer target to gitignore

stevebriskin:


.gitignore
```diff
@@ -1,4 +1,5 @@
 /target
+/micro-rdk-installer/target
```
npmenard, line 2: wit the workspace update there shouldn't be a target folder created under micro-rdk-installer. Under which condition do you end up in this situation?

stevebriskin: ack. i'll double check. this was an old fork

mattjperez:
@stevebriskin are you still having this issue on an updated build?

PR #216, opened 2024-05-30
RSDK-7777 - fix data manager not flushing reader on successful upload

gvaradarajan:


micro-rdk/src/common/data_manager.rs
```diff
@@ -234,32 +234,46 @@ pub struct DataSyncTask<StoreType> {
     part_id: String,
 }
 
+async fn read_messages_for_collector(
```
acmorrow, line 237: This isn't `async` now

acmorrow (approved):
LGTM with one small thing

PR #218, opened 2024-06-04
chunk upload requests in data management

gvaradarajan:
This also addresses the bug of data collection pausing when the ESP32 is offline because the attempt to upload data hangs, never releasing the lock on the store.

### DIFF
--- a/micro-rdk/src/common/data_manager.rs
+++ b/micro-rdk/src/common/data_manager.rs
@@ -18,9 +18,15 @@ use super::robot::{LocalRobot, RobotError};
 use async_io::Timer;
 use bytes::BytesMut;
 use futures_lite::prelude::Future;
+use futures_lite::FutureExt;
 use futures_util::lock::Mutex as AsyncMutex;
 use prost::Message;
 use thiserror::Error;
+
+// Maximum size in bytes of readings that should be sent in a single request
+// as recommended by Viam's data management team is 64K. To accommodate for
+// the smaller amount of available RAM, we've halved it
+static MAX_SENSOR_CONTENTS_SIZE: usize = 32000;
 
 #[derive(Debug, Error)]
 pub enum DataManagerError {
@@ -234,22 +240,6 @@ pub struct DataSyncTask<StoreType> {
     part_id: String,
 }
 
-fn read_messages_for_collector(
-    reader: &mut impl DataStoreReader,
-) -> Result<Vec<SensorData>, DataSyncError> {
-    let mut raw_messages: Vec<BytesMut> = vec![];
-    loop {
-        let next_message = reader.read_next_message()?;
-        if next_message.is_empty() {
-            break;
-        }
-        raw_messages.push(next_message);
-    }
-    let data: Result<Vec<SensorData>, prost::DecodeError> =
-        raw_messages.into_iter().map(SensorData::decode).collect();
-    Ok(data?)
-}
-
 impl<StoreType> DataSyncTask<StoreType>
 where
     StoreType: DataStore,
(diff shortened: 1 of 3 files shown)


micro-rdk/src/common/data_manager.rs
```diff
@@ -262,43 +252,128 @@ where
...
+            loop {
+                let mut reader = match store_lock.get_reader(collector_key) {
+                    Ok(reader) => reader,
+                    Err(err) => {
+                        log::error!(
+                            "error acquiring reader for collector key ({:?}): {:?}",
+                            collector_key,
+                            err
+                        );
+                        continue;
```
acmorrow, line 264: Does this mean to continue to the next collector? It looks to me like it will just immediately retry `get_reader(collector_key)`.  If you meant to continue with the next collector, just label the loops and then make clear on `break`s and `continue`s which one you are referring to.

micro-rdk/src/common/data_manager.rs
```diff
@@ -262,43 +252,128 @@ where
...
+                    Ok(msg) => msg,
+                    Err(err) => {
+                        log::error!(
+                            "error reading message from store for collector key ({:?}): {:?}",
+                            collector_key,
+                            err
+                        );
+                        // we don't want to panic, and creating an AppClientError variant for this case
+                        // feels too specific, so we'll return to exit the function early
+                        break;
```
acmorrow, line 278: Comment says "return", code says "break", but it looks to me like it actually moves to the next collector.

micro-rdk/src/common/data_manager.rs
```diff
@@ -262,43 +252,128 @@ where
...
+                        }
+                    };
+                    if next_message.len() > MAX_SENSOR_CONTENTS_SIZE {
+                        log::error!(
+                            "message encountered that was too large (>32K bytes) for collector {:?}",
+                            collector_key
+                        );
+                        continue;
+                    }
+                    let current_chunk_size: usize = current_chunk.iter().map(|m| m.len()).sum();
```
acmorrow, line 316: I think this is N^2 in number of messages that end up in the chunk since you will re-compute the sum after each iteration. It is definitely safer and I hesitate to add more complexity to this already fairly complex method, but it does seem like you could avoid the N^2 by tracking the current total.

micro-rdk/src/common/data_manager.rs
```diff
@@ -262,43 +252,128 @@ where
...
+                            Ok(_) => {
+                                current_chunk.push(next_message);
+                                break true;
+                            }
+                            // If the request takes too long to elicit a response, we don't know whether
+                            // the upload was successful on app's side. We've decided that we'd rather
+                            // potentially lose some data than upload duplicate data, and so we opt to
+                            // assume the best and consume the messages
+                            Err(AppClientError::AppClientRequestTimeout) => {
+                                // current_chunk_size = next_message.len();
```
acmorrow, line 361: Remove

micro-rdk/src/common/data_manager.rs
```diff
@@ -262,43 +252,128 @@ where
...
+                let mut current_chunk: Vec<BytesMut> = vec![];
+                let next_message = match reader.read_next_message() {
+                    Ok(msg) => msg,
+                    Err(err) => {
+                        log::error!(
+                            "error reading message from store for collector key ({:?}): {:?}",
+                            collector_key,
+                            err
+                        );
+                        // we don't want to panic, and creating an AppClientError variant for this case
```
acmorrow, line 276: Personal style nit: add a blank line after code before starting a comment.

micro-rdk/src/common/data_manager.rs
```diff
@@ -262,43 +252,128 @@ where
...
+                                r#type: DataType::TabularSensor.into(),
+                                component_name: collector_key.r_name.clone(),
+                                method_name: collector_key.method.to_string(),
+                                ..Default::default()
+                            }),
+                            sensor_contents: data,
+                        };
+                        match app_client
+                            .upload_data(upload_request)
+                            .or(async {
```
acmorrow, line 346: Just confirming my understanding: we will have the lock on the store during this upload, and that is desired.

acmorrow (changes requested):

micro-rdk/src/common/data_manager.rs
gvaradarajan: yes, the timeout will ensure that we are not holding on to the lock for very long

micro-rdk/src/common/data_manager.rs
```diff
@@ -262,43 +252,140 @@ where
...
+
+                    // skip this message if it's too big
+                    if next_message.len() > MAX_SENSOR_CONTENTS_SIZE {
+                        log::error!(
+                            "message encountered that was too large (>32K bytes) for collector {:?}",
+                            collector_key
+                        );
+                        continue;
+                    }
+                    // let current_chunk_size: usize = current_chunk.iter().map(|m| m.len()).sum();
```
mattjperez, line 324: guessing this isn't needed anymore

micro-rdk/src/common/data_manager.rs
```diff
@@ -262,43 +252,139 @@ where
...
+                            Err(err) => return Err(err),
+                        };
+                    } else {
+                        current_chunk_size += next_message.len();
+                        current_chunk.push(next_message);
+                    }
                 };
-                app_client.upload_data(upload_request).await?;
+
+                // all of the data in the current chunk has been successfully uploaded, so we
```
acmorrow, line 382: This is outside the inner loop, so it isn't actually flushing the current chunk, it is all chunks. It feels related to my above concern about the Nth chunk.

micro-rdk/src/common/data_manager.rs
acmorrow: Let's add a comment making clear that we are intentionally doing the RPC with the lock held and why we believe it is safe to do so.

micro-rdk/src/common/data_manager.rs
```diff
@@ -262,43 +252,139 @@ where
...
+                            // If the request takes too long to elicit a response, we don't know whether
+                            // the upload was successful on app's side. We've decided that we'd rather
+                            // potentially lose some data than upload duplicate data, and so we opt to
+                            // assume the best and consume the messages
+                            Err(AppClientError::AppClientRequestTimeout) => {
+                                current_chunk_size = next_message.len();
+                                current_chunk.push(next_message);
+                                break true;
+                            }
+                            Err(err) => return Err(err),
```
acmorrow, line 374: If we hit this case on the Nth chunk but we have actually sent the prior N-1 chunks OK, doesn't this direct exit from the function mean we will not flush? So those other chunks will get resent on the next cycle?

acmorrow (changes requested):

micro-rdk/src/common/data_manager.rs
gvaradarajan: I don't think that's true? The inner loop (L. 300) represents the processing of single chunk and the outer loop (L. 257) is looping over the region one chunk at a time. Since the flush is happening at the end of an iteration of the outer loop, it is happening at the end of each single chunk processed. I did this to prompt each chunk to have its own reader and so that returning here would exit the entire sync loop (so we don't keep requesting when the network is down).

This confusion is a definite indicator that I need to refactor this code for clarity

acmorrow: Having re-read it, you are correct!

micro-rdk/src/common/data_manager.rs
acmorrow: As above.

acmorrow (approved):

PR #222, opened 2024-06-06
RSDK-7830 - fix panic occurring when capturing frequency is zero

gvaradarajan:


acmorrow:
@gvaradarajan - I see that there is a now-avoided division-by-zero issue here, but it isn't clear to me how we reached that code path to begin with. According to the ticket, the issue arises if there is a data manager but no configured components. Shouldn't that mean that no collectors are even created? So how was the div-by-zero reachable.

acmorrow:
Also, per the gist in the ticket, the div-by-zero crash was:
``
thread 'main' panicked at /host/.micro-rdk-docker-caches/cargo-registry/git/checkouts/micro-rdk-5c871d18509c9d7e/4f0814a/micro-rdk/src/common/data_store.rs:162
``` (https://github.com/viamrobotics/micro-rdk/blob/d2f1817406ff4984f413e2b1c3f929f94216b36e/micro-rdk/src/common/data_store.rs#L162)

That looks like a div-by-zero that makes more sense in the no-collectors case.

gvaradarajan:
@acmorrow well now, I'm more confused, because this code

https://github.com/viamrobotics/micro-rdk/blob/main/micro-rdk/src/common/data_manager.rs#L97

should have covered for this case, no?

gvaradarajan:
actual bug was found, re-requesting review

PR #223, opened 2024-06-10
Remove support for writing certificates, fqdns etc..in the NVS

npmenard:
Mostly code removal since provisioning is a default builtin in micro-rdk builds


micro-rdk-installer/Cargo.toml
```diff
@@ -14,23 +14,17 @@ version.workspace = true
 anyhow = { workspace = true }
 clap = { workspace = true, features = ["derive"] }
 crc32fast.workspace = true
-der.workspace = true
```
acmorrow, line 17: Are any of the removals from this file the last ref to a package in the top-level `Cargo.toml`? If so they should probably be removed there too.

acmorrow (approved):
This all looks fine to me, but I wasn't closely involved with the installer development, so probably good if @gvaradarajan has a look as well.

micro-rdk-installer/Cargo.toml
npmenard: yup, will do that

micro-rdk-installer/src/nvs/request.rs
```diff
@@ -1,37 +1,30 @@
...
-            let rest_of_url = format!("/download/{version}/micro-rdk-server-esp32.bin");
-            release_url.push_str(&rest_of_url)
-        }
-        None => release_url.push_str("latest/download/micro-rdk-server-esp32.bin"),
-    };
+    let release_url = version.map_or_else(
+        || {
+            format!(
+                "{}/{}/{}",
+                RELEASES_BASE_URL, "latest/download/", BINARY_NAME
```
gvaradarajan, line 22: looks like this will result in two slashes after the word "download"? Or am I missing something?

gvaradarajan (changes requested):
LGTM, but found one potential bug?

micro-rdk-installer/src/nvs/request.rs
npmenard: yeah the previous piece of code had it although it was before the download know that I re read it (L30) should I put it there? I don't think the // matter but if it's not needed then we should probably remove it.

gvaradarajan (approved):
LGTM, pending the removal of now unused packages from the workspace level Cargo.toml

PR #224, opened 2024-06-10
fix data store test

npmenard:
in `StaticMemoryDataStore::new` if no collector were configured then the `fetch_or` operation on `DATA_STORE_INITIALIZED` will still complete and incorrectly store true. Further call to `StaticMemoryDataStore::new` will always fail


gvaradarajan (approved):
LGTM, my bad

PR #225, opened 2024-06-11
fix compile error when building without default features

gvaradarajan:
Also includes a fix to data store tests


acmorrow (approved):
LGTM

PR #226, opened 2024-06-11
[RSDK-7813] add FakeCamera

mattjperez:
These changes let you register a `FakeCamera` on app.viam and be able to connect to it via SDK.

still TODO:
- [x] Add a static image to `FakeCamera` for verifying `GetImage`
- [x] Confirm image grabbed with GetImage on SDK over HTTP2
  - confirmed on go and python

Ended up going a little further than originally planned for this PR. It enables (for native-only):
- GetImage
- ~RenderFrame (mimics GetImage, displays on app.viam)~ deprecated except for RC page. will maybe revisit in later ticket/PR
- DoCommand


micro-rdk-server/Cargo.toml
```diff
@@ -17,7 +17,7 @@ qemu = []
 [target.'cfg(not(target_os = "espidf"))'.dependencies]
 env_logger.workspace = true
 local-ip-address.workspace = true
-micro-rdk = { workspace = true, features = ["native", "provisioning"], default-features = true }
+micro-rdk = { workspace = true, features = ["native", "provisioning", "camera"], default-features = true }
```
mattjperez, line 20: added for development convenience, remove before PR is merged

micro-rdk/src/common/robot.rs
```diff
@@ -183,10 +185,10 @@ impl LocalRobot {
             let cfg = &mut components[iter.next().unwrap()];
             if let Some(cfg) = cfg.as_ref() {
                 // capture the error and make it available to LocalRobot so it can be pushed in the logs?
-                if self
-                    .build_resource(cfg, board.clone(), board_key.clone(), &mut registry)
-                    .is_err()
+                if let Err(e) =
+                    self.build_resource(cfg, board.clone(), board_key.clone(), &mut registry)
                 {
+                    log::error!("{:?}", e);
```
mattjperez, line 191: Neither the `RUST_LOG=debug` errors nor app.viam logs were helpful when troubleshooting init failure for camera. Not directly related to the ticket but may be useful to add either here or separate PR

Makefile
```diff
@@ -37,7 +37,7 @@ build-native:
 	cargo build -p micro-rdk-server --bin micro-rdk-server-native
 
 native:
-	cargo run -p micro-rdk-server --bin micro-rdk-server-native
+	MICRO_RDK_WIFI_PASSWORD=0 cargo run -p micro-rdk-server --bin micro-rdk-server-native
```
mattjperez, line 40: also for development convenience. used to bypass provisioning and use the local `viam.json`. 
will remove before merging

micro-rdk/src/common/camera.rs
```diff
@@ -17,16 +24,53 @@ pub enum CameraError {
...
+        .is_err()
+    {
+        log::error!("fake camera type is already registered");
+    }
+}
 
 pub trait Camera {
     fn get_frame(&mut self, buffer: BytesMut) -> Result<BytesMut, CameraError>;
+    fn get_frames(&mut self, _buffer: BytesMut) -> Result<BytesMut, CameraError> {
+        unimplemented!();
```
npmenard, line 40: should not be `unimplemented!()` either without a body or returning an error

micro-rdk/src/common/camera.rs
```diff
@@ -17,16 +24,53 @@ pub enum CameraError {
...
 
+impl FakeCamera {
+    pub fn new() -> Self {
+        FakeCamera {}
+    }
+    pub(crate) fn from_config(
+        _cfg: ConfigType,
+        _: Vec<Dependency>,
+    ) -> Result<CameraType, CameraError> {
+        log::info!("making new camera");
```
npmenard, line 69: should be removed once done with dev

micro-rdk/src/common/grpc.rs
```diff
@@ -1246,6 +1251,11 @@ where
         Err(ServerError::from(GrpcError::RpcUnavailable))
     }
 
+    #[cfg(feature = "camera")]
+    fn camera_get_frames(&mut self, _message: &[u8]) -> Result<(), ServerError> {
```
npmenard, line 1255: if theses functions (including point cloud, render frame etc.. are going to remain unimplemented we should remove the code because by default we return unimplemented on all unknown RPCs

micro-rdk/src/common/grpc.rs
```diff
@@ -220,6 +220,8 @@ where
             #[cfg(feature = "camera")]
             "/viam.component.camera.v1.CameraService/GetImage" => self.camera_get_frame(payload),
```
npmenard, line 221: rename frame->image

micro-rdk/src/common/camera.rs
```diff
@@ -17,16 +24,53 @@ pub enum CameraError {
...
+    if registry
+        .register_camera("fake", &FakeCamera::from_config)
+        .is_err()
+    {
+        log::error!("fake camera type is already registered");
+    }
+}
 
 pub trait Camera {
     fn get_frame(&mut self, buffer: BytesMut) -> Result<BytesMut, CameraError>;
```
npmenard, line 38: rename frame->image

micro-rdk/src/common/robot.rs
npmenard: format the error reporting better in this case so there is a bit more context

npmenard (changes requested):
done with first pass

micro-rdk/src/common/grpc.rs
mattjperez: So remove unimplemented rpcs from `grpc.rs` but leave the trait methods in `camera.rs` (returning an error/leaving them blank)?

micro-rdk/src/common/webrtc/grpc.rs
```diff
@@ -93,7 +93,7 @@ where
             stream: None,
             headers: None,
             streams: HashMap::new(),
-            buffer: BytesMut::zeroed(1650),
+            buffer: BytesMut::zeroed(15000),
```
mattjperez, line 96: @npmenard I used the Python SDK with `disable_webrtc=True` but the native server still went down this path with webrtc. I need to try again with the Go SDK, but this change did allow me to get everything working.

Not sure what this value should be if it's the limiting factor. We could feature-gate whatever the value is as well.

micro-rdk/src/lib.rs
```diff
@@ -26,6 +26,10 @@ pub use micro_rdk_macros::PowerSensorReadings;
 
 /// gRPC protobuf utilities, auto-generated
 pub mod google {
+    pub mod api {
```
mattjperez, line 29: needed for `google.api.HttpBody`, the response type for `RenderFrame`

mattjperez:
Ended up going a little further than originally planned for this PR. It enables:
- GetImage
- RenderFrame
- DoCommand

These should be everything I need for e2e testing for the rest of the epic.

micro-rdk/src/lib.rs
```diff
@@ -26,6 +26,11 @@ pub use micro_rdk_macros::PowerSensorReadings;
 
 /// gRPC protobuf utilities, auto-generated
 pub mod google {
+    #[cfg(feature = "camera")]
+    pub mod api {
+        #![allow(clippy::derive_partial_eq_without_eq)]
+        include!("gen/google.api.rs");
```
mattjperez, line 32: for `RenderFrame`'s `HttpBody` response

micro-rdk/src/common/camera.rs
```diff
@@ -52,11 +71,67 @@ impl Default for FakeCamera {
...
+        let image = BytesMut::new();
+        let msg = camera::v1::GetImageResponse {
+            mime_type: "image/jpeg".to_string(),
+            image: image.into(),
+        };
+        msg.encode(&mut buffer).unwrap();
+        Ok(buffer)
+    }
+    fn get_images(&mut self, _buffer: BytesMut) -> Result<BytesMut, CameraError> {
+        unimplemented!();
```
mattjperez, line 88: Will change these to return errors

micro-rdk/src/common/camera.rs
```diff
@@ -52,11 +71,67 @@ impl Default for FakeCamera {
...
+    fn render_frame(&mut self, mut buffer: BytesMut) -> Result<BytesMut, CameraError> {
+        #[cfg(feature = "camera")]
+        let msg = google::api::HttpBody {
+            content_type: "image/jpeg".to_string(),
+            data: FAKE_JPEG.into(),
+            ..Default::default()
+        };
+        #[cfg(not(feature = "camera"))]
+        let msg = google::protobuf::Struct::default();
+        msg.encode(&mut buffer).unwrap();
```
npmenard, line 105: remove the unwrap and return an error

micro-rdk/src/common/grpc.rs
```diff
@@ -1218,32 +1224,42 @@ where
...
+        buffer[1] = len[0];
+        buffer[2] = len[1];
+        buffer[3] = len[2];
+        buffer[4] = len[3];
+        buffer.unsplit(msg);
+        self.response.put_data(buffer.freeze());
+        return Ok(());
+    }
+
+    /*
```
npmenard, line 1259: remove comments

micro-rdk/src/common/camera.rs
```diff
@@ -52,11 +71,67 @@ impl Default for FakeCamera {
...
+    }
+    fn get_properties(&mut self, _buffer: BytesMut) -> Result<BytesMut, CameraError> {
+        unimplemented!();
+    }
+    fn render_frame(&mut self, mut buffer: BytesMut) -> Result<BytesMut, CameraError> {
+        #[cfg(feature = "camera")]
+        let msg = google::api::HttpBody {
+            content_type: "image/jpeg".to_string(),
+            data: FAKE_JPEG.into(),
+            ..Default::default()
```
npmenard, line 101: do we make any use of HTTP extensions?

npmenard (changes requested):
Done with Pass! Nice work

micro-rdk/src/common/camera.rs
mattjperez: not in [rdk camera server](https://github.com/viamrobotics/rdk/blob/04037e3ddc9abd2740b1e3d5f3e8152e04e3b121/components/camera/server.go#L194) at least

micro-rdk/src/common/camera.rs
```diff
@@ -1,12 +1,23 @@
...
 
+#[cfg(feature = "camera")]
+use {
+    super::{config::ConfigType, registry::ComponentRegistry, registry::Dependency},
+    std::sync::Arc,
+};
+
+pub static COMPONENT_NAME: &str = "camera";
+#[cfg(feature = "camera")]
+pub static FAKE_JPEG: &[u8] = include_bytes!("../../../etc/assets/symbol.jpg");
```
acmorrow, line 19: Any reason not to just put this in the same directory as `camera.rs`, or under an `assets` directory there rather than at the top?

And `symbol` is a pretty generic name. Maybe `fake_camera_image`?.

mattjperez: I was avoiding putting artifacts directly in the micro-rdk core repo.
fine with putting it here though. 

I'm tempted to move the FakeCamera and the image into the native directory and leave just the trait definition here. 

As for the name, I had several candidate files when testing, (symbol, checkers, etc). will rename it

acmorrow: Yes, it should live next to whatever ends up using it, wherever that lives.

micro-rdk/src/lib.rs
mattjperez: render frame no longer being implemented this PR

npmenard (approved):
LGTM pending tests over HTTP2

micro-rdk/src/common/registry.rs
```diff
@@ -152,6 +165,8 @@ impl Default for ComponentRegistry {
             crate::common::generic::register_models(&mut r);
             crate::common::ina::register_models(&mut r);
             crate::common::wheeled_base::register_models(&mut r);
+            #[cfg(feature = "camera")]
```
gvaradarajan, line 168: `fake_camera` only exists if the "native" feature is activated, but here we attempt to register models even if the native feature is not active. Your PR seems to be passing CI checks despite this but I'm not sure how?

EDIT: it's because we don't pass the camera feature in any of our CI checks (which I'm not necessarily advocating for doing). I've confirmed that the following fails
```
cargo +esp clippy -p micro-rdk  --features esp32,provisioning,camera  --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort -- -Dwarnings
```
The follow-up question here is: should we seek a more graceful compilation failure via [compile_error](https://doc.rust-lang.org/std/macro.compile_error.html)?

gvaradarajan (changes requested):
LGTM, but I would like the one thing I noticed addressed somehow (or a decision made on it in consultation with @npmenard and @acmorrow )

micro-rdk/src/common/registry.rs
mattjperez: For the `Esp32Camera`, this ticket only covers native. In the [epic](https://viam.atlassian.net/browse/RSDK-7813?atlOrigin=eyJpIjoiYzFmMWNkNmZjNzkxNDVmMjgxNzY0MDNiNTllZWMyMDIiLCJwIjoiaiJ9), esp32 implementations and testings come later.

mattjperez: I can do an immediate PR after so it passes for esp32,camera. Or if you don't mind waiting a week and I fix it when I get to the esp32 stuff

gvaradarajan: Could you put the native flag above this line though? Otherwise even when you fix `Esp32Camera` it will still fail

gvaradarajan (approved):
LGTM

micro-rdk/src/common/grpc.rs
```diff
@@ -35,7 +35,7 @@ use thiserror::Error;
 use super::webrtc::grpc::WebRtcGrpcService;
 
 #[cfg(feature = "camera")]
-static GRPC_BUFFER_SIZE: usize = 10240;
+static GRPC_BUFFER_SIZE: usize = 1024 * 1024; // 1MB
```
mattjperez, line 38: grpc spec has an incoming message limit of 4MB. the current `fake_image.jpeg` is about 10KB, the rdk's `FakeCamera` saves a jpeg of ~20KB.

micro-rdk/src/common/grpc.rs
```diff
@@ -1218,47 +1208,59 @@ where
...
+
+        let msg_buf = camera
+            .lock()
+            .unwrap()
+            .get_image(msg_buf)
+            .map_err(|err| ServerError::new(GrpcError::RpcInternal, Some(err.into())))?;
+
+        // NOTE: can't use buffer.capacity() here, would return 5, may be due to multiple RefCells
+        if 5 + msg_buf.len()
+            > GRPC_BUFFER_SIZE
```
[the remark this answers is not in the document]
mattjperez: Should I change the line in [`encode_message`](https://github.com/viamrobotics/micro-rdk/blob/26dc4a70874461f5679d76356afd2b186668aaf2/micro-rdk/src/common/grpc.rs#L1279) as well for consistency?

micro-rdk/src/common/grpc.rs
```diff
@@ -1218,47 +1208,59 @@ where
...
+        let mut buffer = RefCell::borrow_mut(&self.buffer).split_off(0);
+        let msg_buf = buffer.split_off(5);
+
+        let msg_buf = camera
+            .lock()
+            .unwrap()
+            .get_image(msg_buf)
+            .map_err(|err| ServerError::new(GrpcError::RpcInternal, Some(err.into())))?;
+
+        // NOTE: can't use buffer.capacity() here, would return 5, may be due to multiple RefCells
```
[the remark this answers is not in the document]
mattjperez: for example

```rs
    let mut first = BytesMut::with_capacity(9000);
    println!("first capacity: {}", first.capacity());

    let second = first.split_off(100);
    println!("first capacity: {}", first.capacity());
    println!("second capacity: {}", second.capacity());
```

output:
```sh
first capacity: 9000
first capacity: 100
second capacity: 8900
```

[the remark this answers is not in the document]
npmenard: yes seems logical if you split the buffer with split_off. left would have [0,100) and right [100,9000) so capacity returned is correct

micro-rdk/src/common/grpc.rs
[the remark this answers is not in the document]
npmenard: In encode_message we check if data will fit before we do the actual encoding. (hence the call tne encoded_len)

micro-rdk/src/common/grpc.rs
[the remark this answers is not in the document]
mattjperez: since I'm removing the following check this comment will go as well.

PR #228, opened 2024-06-12
Add a version check

npmenard:
Add version check so that a user cannot use the installer to flash micro-rdk version < 0.1.9


mattjperez (approved):
LGTM

PR #229, opened 2024-06-12
[RSDK-7685] consolidate example folders into micrordk

allisonschiang:
Pt 1 (duplicating folders into micrordk)


mattjperez:
You can create (convert this to) a draft PR to verify your changes or until it's ready for an official review.

mattjperez:
I think it's best that we don't include the target directories. You can delete them locally then push again.

PR #230, opened 2024-06-12
Convert RSSI to signal strenght

npmenard:
Two changes :
 1) SSID so it's more explicit
 2)  Convert RSSI to signal strength as this is what Provisioning API expects


acmorrow (approved):
LGTM

PR #233, opened 2024-06-13
[RSDK-7685] move examples into modular-drivers

allisonschiang:
pr for just file movement!

### DIFF
--- a/examples/micro-rdk-esp32-sensor-examples/Cargo.toml
+++ b/examples/micro-rdk-esp32-sensor-examples/Cargo.toml
@@ -1,13 +1,16 @@
 [package]
 name = "micro-rdk-esp32-sensor-examples"
-version = "0.1.0"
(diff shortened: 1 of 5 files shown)


examples/micro-rdk-esp32-sensor-examples/README.md
```diff
@@ -1,143 +0,0 @@
-# Viam Micro-RDK ESP32 Sensor Modules Example
```
acmorrow, line 1: I think this would be the README.md to keep, since it has content, unlike the one from micro-rdk-modular-driver-examples.

acmorrow (changes requested):
I think that rather than deleting the files that collide, you should just move one and leave the other in place. Then in the next review where things get resolved we will be able to see the deletion in one place and the addition in the other.

acmorrow (approved):

PR #235, opened 2024-06-17
consolidate cargo.toml, lib.rs, and add into workspace

allisonschiang:


examples/modular-drivers/src/lib.rs
```diff
@@ -2,9 +2,13 @@ use micro_rdk::common::registry::{ComponentRegistry, RegistryError};
...
 pub mod free_heap_sensor;
 pub mod wifi_rssi_sensor;
+pub mod moisture_sensor;
+pub mod water_pump;
 
 pub fn register_models(registry: &mut ComponentRegistry) -> Result<(), RegistryError> {
     wifi_rssi_sensor::register_models(registry)?;
     free_heap_sensor::register_models(registry)?;
+    moisture_sensor::register_models(registry)?;
+    water_pump::register_models(registry)?;
```
acmorrow, line 12: Ditto alphabetize

examples/modular-drivers/Cargo.toml
```diff
@@ -1,16 +1,16 @@
...
+license.workspace = true
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
 
 [dependencies]
-log = "0.4"
-micro-rdk = {version = "0.2.0", git = "https://github.com/viamrobotics/micro-rdk.git", features = ["esp32"], rev = "ad31cf1"}
+log.workspace = true
+micro-rdk = { workspace = true, features = ["esp32"] }
```
acmorrow, line 13: Two of the sensors are esp32 specific (heap, wifi) and the other two are not. So, this crate needs to have an esp32 feature, and it needs to interact with the micro-rdk feature flag for esp32 support.

examples/modular-drivers/Cargo.toml
```diff
@@ -1,16 +1,16 @@
...
-edition = "2021"
-authors = ["Viam, inc. <contact@viam.com>"]
-resolver = "2"
-rust-version = "1.75"
-license = "AGPL-3.0"
+name = "modular-drivers"
+authors.workspace = true
+description.workspace = true
+edition.workspace = true
+license.workspace = true
```
acmorrow, line 6: Interestingly, the example modules do *not* have the same license as the overall workspace.

The micro-rdk is mostly AGPL, but the example modules are Apache.

acmorrow (changes requested):

Cargo.toml
```diff
@@ -6,6 +6,7 @@ members = [
   "micro-rdk-installer",
   "micro-rdk-macros",
   "micro-rdk-server",
+  "examples",
```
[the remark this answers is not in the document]
allisonschiang: examples/modular-drivers is correct

examples/modular-drivers/Cargo.toml
```diff
@@ -1,10 +1,9 @@
 [package]
-name = "micro-rdk-esp32-sensor-examples"
+name = "modular-drivers"
```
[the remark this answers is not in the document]
allisonschiang: I'll edit this in the next PR (to separate "moving files" and editing files)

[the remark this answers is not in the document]
acmorrow: Sorry, I was perhaps unclear. I'm not proposing moving any files here: just saying `name = micro-rdk-modular-driver-example`. So it is just an edit and should be fine to include here.

Also, FYI: Usually our workflow is to have the author of a comment resolve the comment when they feel it has been addressed, rather than resolving them yourself. Otherwise it can be hard for a reviewer to keep track of whether things changed as desired or not. I'll re-open some of the conversations that I think are still live.

examples/modular-drivers/Cargo.toml
acmorrow: I may have been too oblique here: please change this to reflect the actual license of the examples (Apache)

examples/modular-drivers/Cargo.toml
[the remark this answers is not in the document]
allisonschiang: oops, thanks for the workflow info!

examples/modular-drivers/src/lib.rs
```diff
@@ -1,10 +1,16 @@
...
+pub mod water_pump;
+#[cfg(feature = "esp32")]
 pub mod wifi_rssi_sensor;
 
 pub fn register_models(registry: &mut ComponentRegistry) -> Result<(), RegistryError> {
-    wifi_rssi_sensor::register_models(registry)?;
     free_heap_sensor::register_models(registry)?;
+    moisture_sensor::register_models(registry)?;
+    water_pump::register_models(registry)?;
+    wifi_rssi_sensor::register_models(registry)?;
```
mattjperez, line 14: the wifi_rssi_sensor and free_heap_sensor should also be feature gated here for esp32. Otherwise, running this on native would not compile.

mattjperez (changes requested):
Looks great!
Just a couple more changes.

examples/modular-drivers/Cargo.toml
```diff
@@ -1,16 +1,19 @@
...
+authors.workspace = true
+description.workspace = true
+edition.workspace = true
+license = "Apache-2.0"
+repository.workspace = true
+rust-version.workspace = true
+version.workspace = true
+
+[features]
+esp32 = []
```
acmorrow, line 12: I think enabling the `esp32` feature her should enable it in `micro-rdk` as well. Something like:
```
[features]
esp32 = ['micro-rdk/esp32']
```

examples/modular-drivers/Cargo.toml
acmorrow: I think this shouldn't be unconditionally enabling the `esp32` feature in `micro-rdk`, per above.

examples/modular-drivers/Cargo.toml
```diff
@@ -1,16 +1,19 @@
...
-resolver = "2"
-version = "0.2.0"
-edition = "2021"
-authors = ["Viam, inc. <contact@viam.com>"]
-resolver = "2"
-rust-version = "1.75"
-license = "AGPL-3.0"
+name = "micro-rdk-modular-driver-example"
+authors.workspace = true
+description.workspace = true
```
acmorrow, line 4: Maybe this crate warrants its own description.

acmorrow (changes requested):
I think this is very close to ready to merge.

acmorrow (approved):
LGTM

mattjperez (commented):
LGTM! will defer to the others for final approval

PR #237, opened 2024-06-20
fix memory leak occurring when wifi is connected but app is unreachable

gvaradarajan:
we've added a wrapper around the foreign `esp_tls_t` pointer and implemented Drop to ensure that it is always cleaned up in appropriate cases

### DIFF
--- a/micro-rdk/src/common/conn/server.rs
+++ b/micro-rdk/src/common/conn/server.rs
@@ -452,9 +452,7 @@ where
                     futures_lite::future::Pending<Result<AppSignaling, AppClientError>>,
                 >::default())
             };
-
             let listener = self.http_listener.next_conn();
-
             log::info!("waiting for connection");
 
             let connection = futures_lite::future::or(
--- a/micro-rdk/src/esp32/tls.rs
+++ b/micro-rdk/src/esp32/tls.rs
@@ -15,8 +15,8 @@ use futures_lite::{ready, AsyncRead, AsyncWrite};
 use std::{
     fmt::Debug,
     io::{Read, Write},
-    mem::ManuallyDrop,
     net::TcpStream,
+    ops::Deref,
     os::{fd::FromRawFd, raw::c_char, unix::prelude::AsRawFd},
     task::Poll,
 };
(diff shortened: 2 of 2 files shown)


micro-rdk/src/esp32/tls.rs
```diff
@@ -199,17 +207,16 @@ impl Esp32TLSStream {
                 "couldn't allocate tls context",
             ));
         }
-        let tls_context = ManuallyDrop::new(p);
+        let tls_context = TLSContext(ManuallyDrop::new(p));
```
acmorrow, line 210: Could the `TLSContext` wrapper just take care of calling `esp_tls_init`? Then there would be no window where it could go wrong. As written, `p` comes into existence at line 203 but it isn't truly under proper management until after 210.

I think it would mostly amount to just adding a `TLSContext::new() -> Result<Self, SomeErrorHere>` and then calling it like `let tls_context = TLSContext::new()?` here.

micro-rdk/src/common/conn/server.rs
```diff
@@ -452,9 +452,7 @@ where
                     futures_lite::future::Pending<Result<AppSignaling, AppClientError>>,
                 >::default())
             };
-
```
acmorrow, line 455: Some stray whitespace changes here

micro-rdk/src/esp32/tls.rs
```diff
@@ -309,15 +326,6 @@ impl Esp32TLSStream {
     }
 }
 
-impl Drop for Esp32TLSStream {
-    fn drop(&mut self) {
-        log::error!("dropping the tls stream");
-        if let Some(err) = EspError::from(unsafe { esp_tls_conn_destroy(*self.tls_context) }) {
-            log::error!("error while dropping the tls connection '{}'", err);
```
acmorrow, line 316: I'm not sure how important these log messages are, but they will no longer happen. If they are valuable then they should migrate somewhere else.

acmorrow (changes requested):

micro-rdk/src/esp32/tls.rs
gvaradarajan: I'll bring them back

micro-rdk/src/esp32/tls.rs
gvaradarajan: yeah that's fair, I'll add that

micro-rdk/src/esp32/tls.rs
```diff
@@ -199,17 +207,16 @@ impl Esp32TLSStream {
...
+        let tls_context = TLSContext(ManuallyDrop::new(p));
         match tls_cfg {
             Either::Left(tls_cfg) => {
                 let fd = socket.as_ref().unwrap().as_raw_fd();
                 unsafe {
                     if let Some(err) = EspError::from(esp_tls_server_session_create(
                         &mut **tls_cfg,
                         fd,
-                        *tls_context,
+                        *(tls_context.0),
```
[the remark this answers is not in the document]
npmenard: what do you have in mind here?

micro-rdk/src/esp32/tls.rs
```diff
@@ -47,9 +47,17 @@ impl TlsClientConnector for Esp32TLS {
     }
 }
 
+struct TLSContext(ManuallyDrop<*mut esp_tls_t>);
+
+impl Drop for TLSContext {
```
npmenard, line 52: nit: Esp32TLSContext

micro-rdk/src/esp32/tls.rs
```diff
@@ -47,9 +47,17 @@ impl TlsClientConnector for Esp32TLS {
     }
 }
 
+struct TLSContext(ManuallyDrop<*mut esp_tls_t>);
```
npmenard, line 50: we may not need the ManuallyDrop here

npmenard (approved):
LGTM just some nits

acmorrow (approved):
LGTM, though if it works (and per discussion it seems it does), I think you can drop the `(*self.tls_context).thing` in favor of just `self.tls_context.thing` because the `Deref` trait should kick in automagically.

PR #238, opened 2024-06-20
APP-5312 - serialize data collection methods in upper camel case

gvaradarajan:


mattjperez (approved):
LGTM

maxhorowitz:
only concern is that this change should be made for all built-in sensor data capture method types - are those 4 the only that exist?

gvaradarajan:
> only concern is that this change should be made for all built-in sensor data capture method types - are those 4 the only that exist?

yes, for now. I'll add a comment so that future methods are done the right way

micro-rdk/src/common/data_collector.rs
```diff
@@ -69,12 +69,14 @@ pub enum CollectionMethod {
 
 impl Display for CollectionMethod {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        // NOTE: Future added methods need to follow the same upper camel case
```
maxhorowitz, line 72: Thank you!

maxhorowitz (commented):
:shipit:

maxhorowitz (approved):
:shipit:

PR #240, opened 2024-06-24
[RSDK-7992] Add keepalive

npmenard:
Add a keep alive timeout for both the HTTP2 layer and the Socket layer

### DIFF
--- a/micro-rdk/src/esp32/tls.rs
+++ b/micro-rdk/src/esp32/tls.rs
@@ -32,6 +32,10 @@ use super::tcp::Esp32Stream;
 unsafe impl Sync for Esp32TLS {}
 unsafe impl Send for Esp32TLS {}
 
+const TCP_KEEPINTVL_S: i32 = 60; // seconds
+const TCP_KEEPCNT_N: i32 = 4;
+const TCP_KEEPIDLE_S: i32 = 120; // seconds
+
 /// structure to store tls configuration
 #[derive(Clone)]
 pub struct Esp32TLS {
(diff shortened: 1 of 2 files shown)


micro-rdk/src/esp32/tls.rs
```diff
@@ -261,6 +264,56 @@ impl Esp32TLSStream {
                             let mut fd: i32 = 0;
                             esp_idf_svc::sys::esp!(esp_tls_get_conn_sockfd(*tls_context, &mut fd))
                                 .map_err(|e| std::io::Error::new(std::io::ErrorKind::Other, e))?;
+
+                            // set socket keep alive properties as such
+                            // KEEPIDLE is set to 120 second (time before a first keepalive probe is sent)
+                            // KEEPINTVL, KEEPCNT are set to 60 and 4 respectively
+                            // Total time before an IDLE and DEAD connection is closed =  360s
+                            let var: i32 = 1;
+                            if lwip_setsockopt(
```
mattjperez, line 273: what does this first option set? enables keepalive?

micro-rdk/src/esp32/tls.rs
```diff
@@ -261,6 +264,56 @@ impl Esp32TLSStream {
...
+                            {
+                                return Err(std::io::Error::last_os_error());
+                            }
+                            let var: i32 = 60;
+                            if lwip_setsockopt(
+                                fd,
+                                IPPROTO_TCP as i32,
+                                TCP_KEEPINTVL as i32,
+                                &var as *const i32 as *const _,
+                                std::mem::size_of::<i32>() as socklen_t,
```
mattjperez, line 289: Is it possible to declare these values as `const` somewhere (ie `const KEEPALIVE_INTVL_VAL`) and reference them directly (ie on L288 `&KEEPALIVE_INVL_VAL as *const i32 as *const _,...`)

mattjperez (approved):
some non-blocking questions, but LGTm

acmorrow (approved):
LGTM

micro-rdk/src/esp32/tls.rs
npmenard: yup

micro-rdk/src/esp32/tls.rs
npmenard: make sense

PR #244, opened 2024-06-25
RSDK-8036 - change data upload strategy to lose messages that fail to upload

gvaradarajan:
Justification available here: https://viam.atlassian.net/browse/RSDK-8036


npmenard (approved):
LGTM

PR #247, opened 2024-06-26
[RSDK-7876] Update CI for FFI

npmenard:


.github/workflows/publish.yml
```diff
@@ -134,41 +134,59 @@ jobs:
...
       with:
         name: micro-rdk-installer-macos
     - name: Download windows installer
-      uses: actions/download-artifact@v3
+      uses: actions/download-artifact@v4
       with:
         name: micro-rdk-installer-windows.exe
-    - name: Download micro-RDK binary
-      uses: actions/download-artifact@v3
+    - name: Download mMicro-RDK
```
mattjperez, line 152: typo?

.github/workflows/test.yml
```diff
@@ -16,29 +16,30 @@ jobs:
...
+      uses: actions/checkout@v4
       with:
         ref: ${{ github.event.pull_request.head.sha }}
         fetch-depth: 2
     - run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
     - name: path filter
-      uses: dorny/paths-filter@v2
-      id: changes
+      uses: dorny/paths-filter@v3
+      id: filter
```
mattjperez, line 42: What is the effect of this?
The example [here](https://github.com/dorny/paths-filter?tab=readme-ov-file#example) still uses `id: changes` for v3

etc/docker/Dockerfile
```diff
@@ -65,6 +66,11 @@ COPY --from=builder /usr/ldproxy /usr/bin
 COPY --from=builder /usr/espup /usr/bin
 COPY --from=builder /usr/sccache /usr/bin
 
+RUN case "${TARGETPLATFORM}" in \
```
acmorrow, line 69: Why this way rather than with `cargo install`?

acmorrow (approved):
LGTM

.github/workflows/test.yml
npmenard: that just change the key to access the output so `steps.filter` instead of  `steps.changes`

etc/docker/Dockerfile
npmenard: actually we don't need it anymore since we use build.rs to generate bindings. But the idea was that I didn't want to include it in arm version of the docker image

PR #248, opened 2024-06-26
RSDK-8038 - calculate actual timestamps for data before upload

gvaradarajan:


gvaradarajan:
Will test more upon completion of https://viam.atlassian.net/browse/RSDK-7043 (and rebase of changes thereof). Waiting on merge until then, but would still like a code review

micro-rdk/src/common/data_manager.rs
```diff
@@ -107,13 +114,15 @@ where
...
             part_id,
+            robot_start_time,
         })
     }
 
     pub fn from_robot_and_config(
         cfg: &ConfigResponse,
         app_config: &AppClientConfig,
         robot: Arc<Mutex<LocalRobot>>,
+        robot_start_time: Instant,
```
acmorrow, line 125: If `LocalRobot` owns its start time, you won't need this to be passed as an additional parameter and it can just be read out of `LocalRobot`.

micro-rdk/src/common/data_manager.rs
```diff
@@ -327,8 +398,10 @@ where
                     if next_message.is_empty()
                         || ((next_message.len() + current_chunk_size) > MAX_SENSOR_CONTENTS_SIZE)
                     {
-                        let data: Result<Vec<SensorData>, prost::DecodeError> =
-                            current_chunk.drain(..).map(SensorData::decode).collect();
+                        let data: Result<Vec<SensorData>, DataSyncError> = current_chunk
+                            .drain(..)
+                            .map(time_corrected_reading_handler(self.robot_start_time))
```
acmorrow, line 403: What happens here if the function returned by time_corrected_reading_handler returns an error?

micro-rdk/src/common/data_manager.rs
```diff
@@ -231,13 +247,68 @@ pub enum DataSyncError {
...
+    stored_time: Timestamp,
+) -> Result<chrono::Duration, DataSyncError> {
+    let stored_time_dur = Duration::new(stored_time.seconds as u64, stored_time.nanos as u32);
+    let time_to_subtract = robot_start_time.elapsed() - stored_time_dur;
+    let time_to_subtract = chrono::Duration::new(
+        time_to_subtract.as_secs() as i64,
+        time_to_subtract.subsec_nanos(),
+    )
+    .ok_or(DataSyncError::TimeOutOfBoundsError)?;
+    Ok(time_to_subtract)
```
acmorrow, line 275: Is there some from/into type magic you can use here to traverse the three types at play here (`Duration`, `Instant`, and `Timestamp`) so we can absolve ourselves of needing to get it right?

micro-rdk/src/common/data_manager.rs
```diff
@@ -231,13 +247,68 @@ pub enum DataSyncError {
...
+}
+
+fn time_corrected_reading_handler(
+    robot_start_time: Instant,
+) -> impl FnMut(BytesMut) -> Result<SensorData, DataSyncError> {
```
acmorrow, line 280: I'd expect the returned function to capture `robot_start_time` by value, so maybe it can just be `Fn` ?

micro-rdk/src/common/data_manager.rs
```diff
@@ -231,13 +247,68 @@ pub enum DataSyncError {
...
+    let time_to_subtract = robot_start_time.elapsed() - stored_time_dur;
+    let time_to_subtract = chrono::Duration::new(
+        time_to_subtract.as_secs() as i64,
+        time_to_subtract.subsec_nanos(),
+    )
+    .ok_or(DataSyncError::TimeOutOfBoundsError)?;
+    Ok(time_to_subtract)
+}
+
+fn time_corrected_reading_handler(
```
acmorrow, line 278: Let's call this function `make_x` or `create_y`: it isn't actually the handler: it is a factory for the handler.

acmorrow (changes requested):
Looks great! Some small questions and comments.

micro-rdk/src/common/data_collector.rs
```diff
@@ -1,5 +1,5 @@
 use std::fmt::Display;
-use std::time::Duration;
+use std::time::{Duration, Instant};
```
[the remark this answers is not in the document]
gvaradarajan: I believe so, but I'll double-check to confirm the behavior of `Instant` is as expected. I think it's worth noting that, since we're using elapsed, it doesn't really matter whether `Instant::now` returns zero.

micro-rdk/src/common/data_manager.rs
```diff
@@ -87,6 +90,9 @@ pub struct DataManager<StoreType> {
     sync_interval: Duration,
     min_interval: Duration,
     part_id: String,
+    // used for time correcting stored data before upload, see DataSyncTask::run
+    // and time_corrected_reading_handler below
+    robot_start_time: Instant,
```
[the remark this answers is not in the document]
gvaradarajan: when you move the `DataManager` to Robot, my plan is to make `robot_start_time` an argument to `get_sync_task`

micro-rdk/src/common/data_manager.rs
gvaradarajan: I didn't see any, but I'll double-check

micro-rdk/src/common/data_manager.rs
gvaradarajan: I'm not sure I understand what you mean?

micro-rdk/src/esp32/entry.rs
```diff
@@ -98,6 +98,8 @@ pub async fn serve_web_inner(
         let (cfg_response, cfg_received_datetime) =
             client.get_config(network.get_ip()).await.unwrap();
 
+        let robot_start_time = Instant::now();
```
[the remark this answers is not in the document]
gvaradarajan: My plan was to do so when your change to move DataManager to Robot goes in, I wanted to keep it separate for now because I thought it might minimize merge conflicts

micro-rdk/src/common/data_manager.rs
gvaradarajan: if it's to change the signature, I had to make it return `FnMut` so it could be provided to `map`

acmorrow: OK, yeah I see that `std::vec::Drain::map` declares that it takes `FnMut`: https://doc.rust-lang.org/std/vec/struct.Drain.html#method.map, but shouldn't you always be able to pass a `Fn` where a `FnMut` is needed, since `FnMut` is a supertrait of `Fn`?

micro-rdk/src/common/data_manager.rs
acmorrow: Might as well hold off on any changes here until we have a resolution regarding `Instant`

micro-rdk/src/common/data_manager.rs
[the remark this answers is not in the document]
acmorrow: Sounds good.

acmorrow (changes requested):
Overall this looks like the right thing, but:

- I have some reservations about use of `Instant`, please see my comment there.
- You and I should decide ahead of time if this PR or #251 should merge first. If mind goes first I think you should probably fold in the work to attach the start time to the Robot under this PR.

micro-rdk/src/common/data_manager.rs
gvaradarajan: but an `Fn` can't be guaranteed to satisfy `FnMut`, only vice versa, right?

acmorrow: I'm getting myself confused now, or maybe I was originally. But this language from https://doc.rust-lang.org/std/ops/trait.Fn.html is what I was going by:

> Since both [FnMut](https://doc.rust-lang.org/std/ops/trait.FnMut.html) and [FnOnce](https://doc.rust-lang.org/std/ops/trait.FnOnce.html) are supertraits of Fn, any instance of Fn can be used as a parameter where a [FnMut](https://doc.rust-lang.org/std/ops/trait.FnMut.html) or [FnOnce](https://doc.rust-lang.org/std/ops/trait.FnOnce.html) is expected.

So if `Drain::map` wants a `FnMut` a `Fn` should be fine, and then this factory should be able to return `Fn`.

But really, this is unimportant. If the compiler is telling you it needs to be `FnMut`, then by all means listen to the compiler and ignore me.

micro-rdk/src/common/data_collector.rs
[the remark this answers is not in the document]
gvaradarajan: So, for UNIX systems, `Instant::now` uses [clock_gettime()](https://doc.rust-lang.org/nightly/src/std/time.rs.html#113). And I believe for the xtensa targets, UNIX convention is mostly used because ESP-IDF defines many of the same kernel functions. Following the thread from [here](https://github.com/espressif/esp-idf/blob/release/v4.4/components/newlib/time.c#L269) through [here](https://github.com/espressif/esp-idf/blob/release/v4.4/components/newlib/port/esp_time_impl.c#L90), it would seem ESP-IDF uses the RTC timer (which advances during light sleep) in its implementation of `clock_gettime`, meaning light sleep should probably not interfere with the behavior of Instant

micro-rdk/src/common/data_manager.rs
gvaradarajan: Oh I misunderstood how that works, `Fn` worked fine, changing

gvaradarajan:
Following hardware cases tested
1) time corrections work when micro-RDK is online the entire time
2) with network interruptions
3) with initial wifi connection but no internet, with internet turned on later

micro-rdk/src/common/data_collector.rs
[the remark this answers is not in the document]
acmorrow: OK, I'm reasonably convinced. I wouldn't mind seeing a test added that puts the ESP32 into light sleep and validates that time advances plausibly on wakeup, but I'd also be willing to defer that into a ticket. Could you please file one? I just want to ensure that if the behavior changes on us, we find out in our tests, not the field.

micro-rdk/src/common/data_manager.rs
```diff
@@ -269,13 +288,70 @@ pub enum DataSyncError {
...
+    let time_to_subtract = robot_start_time.elapsed() - stored_time_dur;
+    let time_to_subtract = chrono::Duration::new(
+        time_to_subtract.as_secs() as i64,
+        time_to_subtract.subsec_nanos(),
+    )
+    .ok_or(DataSyncError::TimeOutOfBoundsError)?;
+    Ok(time_to_subtract)
+}
+
+fn create_time_corrected_reading(
```
acmorrow, line 321: This function doesn't create a reading though: it creates a thing which adjusts readings. `create_reading_time_corrector`?

micro-rdk/src/common/data_manager.rs
```diff
@@ -369,10 +445,16 @@ where
                     if next_message.is_empty()
                         || ((next_message.len() + current_chunk_size) > MAX_SENSOR_CONTENTS_SIZE)
                     {
-                        let data: Result<Vec<SensorData>, prost::DecodeError> =
-                            current_chunk.drain(..).map(SensorData::decode).collect();
+                        let data: Result<Vec<SensorData>, DataSyncError> = current_chunk
+                            .drain(..)
+                            .map(create_time_corrected_reading(self.robot_start_time))
```
acmorrow, line 450: Unless rust is able to see this and hoist this call automatically, this will create a new time adjuster every time in the loop, and then probably has to do dynamic dispatch under the `Fn`, but as far as I can see `robot_start_time` never changes. Do you even need it anymore? Could it just be a regular function on `self`? At minimum I think you could hoist the creation of the adjuster outside the loop.

micro-rdk/src/common/data_manager.rs
acmorrow: Now that things are settled on `Instant`, let's have another quick look at this (like no more than 10 minutes) to see if we can make the existing libraries do the work for us.

acmorrow (changes requested):
This looks good, I just have a naming quibble, a question about whether something can be lifted out of an inner loop (or even eliminated), and a faint hope about simplifying the routine which manipulates three time related types.

micro-rdk/src/common/data_manager.rs
gvaradarajan: Unfortunately, the answer appears to be no.

acmorrow (approved):
LGTM

PR #249, opened 2024-06-27
RSDK-8042 - fix panic when adjusting PWM after setting bad frequency

gvaradarajan:


micro-rdk/src/esp32/pwm.rs
```diff
@@ -219,14 +219,27 @@ impl<'a> LedcTimerWrapper<'a> {
...
+            Ok(driver) => {
+                let _ = self.timer.set(driver);
+                self.frequency = frequency_hz;
+                Ok(())
+            }
+            Err(err) => {
+                // if there is a failure, or the frequency is bad, we want to revert back
+                // to the last working frequency before raising the error
+                let timer_config = TimerConfig::default().frequency(self.frequency.Hz());
+                let _ = self.timer.set(create_timer_driver(id, &timer_config)?);
```
acmorrow, line 239: What should happen if this restoring call to `create_timer_driver` *also* fails. Right now it will return the error from trying and failing to restore the timer, and not the error from the attempt to create the timer with the new frequency. It will also leave the timer in what I believe is still an invalid state.

gvaradarajan: So I actually don't believe this is possible, because in order for an `LedcTimerWrapper` to exist, this needs to have been called with a proper frequency at some point prior (because of the code in `new`).

acmorrow: OK. I'd say `?` isn't the right thing then. If this is believed to be known to be impossible (by us), we should probably panic! if if happens. So, maybe `expect`?

gvaradarajan: yeah this is reasonable, I'll change it to an expect

acmorrow (approved):
LGTM, maybe get a second opinion from @npmenard though.

stevebriskin:
@gvaradarajan Can you please open a docs ticket to define what valid frequencies and duty cycles are -- https://docs.viam.com/build/micro-rdk/board/esp32/#pwm-signals-on-esp32-pins

gvaradarajan:
> @gvaradarajan Can you please open a docs ticket to define what valid frequencies and duty cycles are -- https://docs.viam.com/build/micro-rdk/board/esp32/#pwm-signals-on-esp32-pins

Is this our responsibility if it's detailed in Espressif's docs?

micro-rdk/src/esp32/pwm.rs
```diff
@@ -219,14 +219,31 @@ impl<'a> LedcTimerWrapper<'a> {
...
+                Ok(())
+            }
+            Err(err) => {
+                // if there is a failure, or the frequency is bad, we want to revert back
+                // to the last working frequency before raising the error
+                let timer_config = TimerConfig::default().frequency(self.frequency.Hz());
+                let _ =
+                    self.timer
+                        .set(create_timer_driver(id, &timer_config).unwrap_or_else(|_| {
+                            panic!("bad frequency previously set on timer {:?}", id)
```
mattjperez, line 242: to confirm, this theoretically can't happen, right?

mattjperez (approved):
LGTM

micro-rdk/src/esp32/pwm.rs
gvaradarajan: yes

micro-rdk/src/esp32/pwm.rs
```diff
@@ -219,14 +219,31 @@ impl<'a> LedcTimerWrapper<'a> {
...
+                self.frequency = frequency_hz;
+                Ok(())
+            }
+            Err(err) => {
+                // if there is a failure, or the frequency is bad, we want to revert back
+                // to the last working frequency before raising the error
+                let timer_config = TimerConfig::default().frequency(self.frequency.Hz());
+                let _ =
+                    self.timer
+                        .set(create_timer_driver(id, &timer_config).unwrap_or_else(|_| {
```
npmenard, line 241: nit: replace with an expect here

micro-rdk/src/esp32/pwm.rs
```diff
@@ -219,14 +219,31 @@ impl<'a> LedcTimerWrapper<'a> {
         if self.count > 0 {
             return Err(Esp32PwmError::OtherChannelsBindToTimer);
         }
+
         let id = {
```
npmenard, line 223: can you add the following comment:

```
// The configured clock source for the timer may not be able to achieve the target frequency,
	// We have to reconfigure the timer so the appropriate clock source for that frequency may be
	// selected. If no appropriate clock source exists the previous timer frequency
        // will be retained
```

npmenard (approved):
LGTM
I am not to comfortable with the flow seems a bit contrived but there are no other api that gives use the flexibility we seek

micro-rdk/src/esp32/pwm.rs
gvaradarajan: I have this commented further down, but it's not worded as well so I'll use this instead

micro-rdk/src/esp32/pwm.rs
gvaradarajan: I tried `.expect(format!(...))` but clippy got mad at me and insisted I do it this way

npmenard: damm clippy :P

PR #251, opened 2024-07-01
RSDK-7582 Robot configuration caching

acmorrow:
This adds support for caching robot configurations.

- Support for storing a robot configuration is added to the NVS and RAM storage infra
- The `serve_web_async` family of functions will create a robot from the cached config during startup.
- The created robot will be live while the startup code attempts to connect to a network and validate credentials.
- Once the network is available, if the credentials validate, the already created robot is used. If not, it is discarded.
- The app client used to validate credentials is re-used all the way down into ViamServer::serve.

Overall, there are several things about this code that leave me less than happy. I'll note them as review comments. However, it does seem to work with the testing I've done so far. I intend to do more, but I'd like to get feedback on the overall approach here before I invest further in any testing in case a redesign or a different approach is desired.

Due to the fairly large changes required to some of the initialization code this may unfortunately be somewhat hard to review.

### DIFF
--- a/micro-rdk-server/esp32/main.rs
+++ b/micro-rdk-server/esp32/main.rs
@@ -27,6 +27,7 @@ mod esp32 {
     }
 
     use micro_rdk::common::registry::ComponentRegistry;
+    use micro_rdk::esp32::nvs_storage::NVSStorage;
 
     pub(crate) fn main_esp32() {
         micro_rdk::esp32::esp_idf_svc::sys::link_patches();
@@ -57,21 +58,44 @@ mod esp32 {
         // then the entire provisioning step can be skipped
         #[cfg(has_robot_config)]
         {
-            use micro_rdk::common::provisioning::storage::RAMStorage;
-            if SSID.is_some() && PASS.is_some() && ROBOT_ID.is_some() && ROBOT_SECRET.is_some() {
-                let ram_storage = RAMStorage::new(
-                    SSID.unwrap(),
-                    PASS.unwrap(),
-                    ROBOT_ID.unwrap(),
-                    ROBOT_SECRET.unwrap(),
-                );
-                serve_web(None, repr, max_connection, ram_storage);
-            }
+            use micro_rdk::common::provisioning::storage::RobotConfigurationStorage;
+            use micro_rdk::common::provisioning::storage::RobotCredentials;
+            use micro_rdk::common::provisioning::storage::WifiCredentialStorage;
+            use micro_rdk::common::provisioning::storage::WifiCredentials;
+
+            log::warn!("Unconditionally using build-time WiFi and robot configuration");
+
+            let storage = NVSStorage::new("nvs").unwrap();
+            log::info!("Storing static values from build time wifi configuration to NVS");
+            storage
+                .store_wifi_credentials(WifiCredentials::new(
+                    SSID.expect("[cfg(has_robot_config)]: missing WiFi SSID")
+                        .to_string(),
+                    PASS.expect("[cfg(has_robot_config)]: missing WiFi password")
+                        .to_string(),
+                ))
+                .expect("Failed to store WiFi credentials to NVS");
+
+            log::info!("Storing static values from build time robot configuration to NVS");
+            storage
+                .store_robot_credentials(
+                    RobotCredentials::new(
+                        ROBOT_ID
+                            .expect("[cfg(has_robot_config)]: missing robot id")
+                            .to_string(),
+                        ROBOT_SECRET
+                            .expect("[cfg(has_robot_config)]: missing robot secret")
+                            .to_string(),
+                    )
+                    .into(),
+                )
+                .expect("Failed to store robot credentials to NVS");
+
+            serve_web(None, repr, max_connection, storage);
         }
         #[cfg(not(has_robot_config))]
         {
             use micro_rdk::common::provisioning::server::ProvisioningInfo;
-            use micro_rdk::esp32::nvs_storage::NVSStorage;
             let mut info = ProvisioningInfo::default();
             info.set_manufacturer("viam".to_owned());
             info.set_model("test-esp32".to_owned());
--- a/micro-rdk/src/common/app_client.rs
+++ b/micro-rdk/src/common/app_client.rs
@@ -254,9 +254,9 @@ impl AppClient {
     // taken from its header for the purposes of timestamping configuration logs and returning
     // `last_reconfigured` values for resource statuses.
     pub async fn get_config(
-        &self,
+        &mut self,
         ip: Ipv4Addr,
-    ) -> Result<(Box<ConfigResponse>, Option<DateTime<FixedOffset>>), AppClientError> {
+    ) -> Result<(AppClientConfig, Box<ConfigResponse>, Option<DateTime<FixedOffset>>), AppClientError> {
         let agent = AgentInfo {
             os: "esp32".to_string(),
             host: "esp32".to_string(),
@@ -293,9 +293,14 @@ impl AppClient {
             None
         };
 
-        let r = r.split_off(5);
+        let cfg_response = ConfigResponse::decode(r.split_off(5))?;
+        if let Some(robot_config) = cfg_response.config.as_ref() {
+            if let Some(cloud_config) = robot_config.cloud.as_ref() {
+                self.config.set_rpc_host(cloud_config.fqdn.clone());
+            }
+        }
 
-        Ok((Box::new(ConfigResponse::decode(r)?), datetime))
+        Ok((self.config.clone(), Box::new(cfg_response), datetime))
     }
 
     pub async fn push_logs(&self, logs: Vec<LogEntry>) -> Result<(), AppClientError> {
--- a/micro-rdk/src/common/conn/server.rs
+++ b/micro-rdk/src/common/conn/server.rs
@@ -363,7 +363,7 @@ where
 
         // Let the robot register any periodic app client tasks it may
         // have based on its configuration.
-        self.app_client_tasks.append(&mut robot.lock().unwrap().generate_periodic_app_client_tasks());
+        self.app_client_tasks.append(&mut robot.lock().unwrap().get_periodic_app_client_tasks());
 
         // Convert each `PeriodicAppClientTask` implementer into an async task spawned on the
         // executor, and collect them all into `_app_client_tasks` so we don't lose track of them.
--- a/micro-rdk/src/common/conn/server.rs
+++ b/micro-rdk/src/common/conn/server.rs
@@ -361,6 +361,10 @@ where
             self.app_client.write().await.replace(app_client);
         }
 
+        // Let the robot register any periodic app client tasks it may
+        // have based on its configuration.
+        self.app_client_tasks.append(&mut robot.lock().unwrap().generate_periodic_app_client_tasks());
+
         // Convert each `PeriodicAppClientTask` implementer into an async task spawned on the
         // executor, and collect them all into `_app_client_tasks` so we don't lose track of them.
         let _app_client_tasks: Vec<_> = self
--- a/micro-rdk/src/common/conn/server.rs
+++ b/micro-rdk/src/common/conn/server.rs
@@ -223,8 +223,6 @@ where
             .unwrap()
             .into();
 
-        self.app_config.set_rpc_host(cfg.fqdn.clone());
-
         self.mdns
             .set_hostname(&cfg.name)
             .map_err(|e| ServerError::Other(e.into()))?;
@@ -356,8 +354,12 @@ where
             network,
         }
     }
-    pub async fn serve(&mut self, robot: Arc<Mutex<LocalRobot>>) {
+    pub async fn serve(&mut self, robot: Arc<Mutex<LocalRobot>>, app_client: Option<AppClient>) {
         let cloned_robot = robot.clone();
+
+        if let Some(app_client) = app_client {
+            self.app_client.write().await.replace(app_client);
+        }
 
         // Convert each `PeriodicAppClientTask` implementer into an async task spawned on the
         // executor, and collect them all into `_app_client_tasks` so we don't lose track of them.
--- a/micro-rdk/src/common/provisioning/storage.rs
+++ b/micro-rdk/src/common/provisioning/storage.rs
@@ -3,7 +3,10 @@ use std::{convert::Infallible, error::Error, fmt::Debug, rc::Rc, sync::Mutex};
 
 use crate::{
     common::grpc::ServerError,
-    proto::provisioning::v1::{CloudConfig, SetNetworkCredentialsRequest},
+    proto::{
+        app::v1::RobotConfig,
+        provisioning::v1::{CloudConfig, SetNetworkCredentialsRequest},
+    },
 };
 
 #[derive(Clone, Default, Debug)]
@@ -63,17 +66,23 @@ pub trait WifiCredentialStorage {
     fn reset_wifi_credentials(&self) -> Result<(), Self::Error>;
 }
 
-pub trait RobotCredentialStorage {
+pub trait RobotConfigurationStorage {
     type Error: Error + Debug + Into<ServerError>;
-    fn has_stored_credentials(&self) -> bool;
+    fn has_robot_credentials(&self) -> bool;
     fn store_robot_credentials(&self, cfg: CloudConfig) -> Result<(), Self::Error>;
     fn get_robot_credentials(&self) -> Result<RobotCredentials, Self::Error>;
     fn reset_robot_credentials(&self) -> Result<(), Self::Error>;
+
+    fn has_robot_configuration(&self) -> bool;
+    fn store_robot_configuration(&self, cfg: RobotConfig) -> Result<(), Self::Error>;
+    fn get_robot_configuration(&self) -> Result<RobotConfig, Self::Error>;
+    fn reset_robot_configuration(&self) -> Result<(), Self::Error>;
 }
 
 #[derive(Default)]
 struct RAMCredentialStorageInner {
-    config: Option<RobotCredentials>,
+    robot_creds: Option<RobotCredentials>,
+    robot_config: Option<RobotConfig>,
     wifi_creds: Option<WifiCredentials>,
 }
 
@@ -83,7 +92,7 @@ pub struct RAMStorage(Rc<Mutex<RAMCredentialStorageInner>>);
 
 impl RAMStorage {
     pub fn new(ssid: &str, pass: &str, robot_id: &str, robot_secret: &str) -> Self {
-        let config = RobotCredentials {
+        let robot_creds = RobotCredentials {
             robot_id: robot_id.to_owned(),
             robot_secret: robot_secret.to_owned(),
         };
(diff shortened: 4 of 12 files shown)


micro-rdk/src/common/robot.rs
```diff
@@ -87,12 +101,22 @@ impl ResourceType {
     }
 }
 
+#[cfg(feature = "native")]
```
gvaradarajan, line 104: Could we combine these with L.11 a la `type Executor = crate::<esp32/native>::exec::<executor_name>`?

micro-rdk/src/common/registry.rs
```diff
@@ -133,6 +133,7 @@ type GenericComponentConstructor =
 
 type DependenciesFromConfig = dyn Fn(ConfigType) -> Vec<ResourceKey>;
 
+#[derive(Clone)]
```
gvaradarajan, line 136: why does this now need to be cloneable?

gvaradarajan (approved):
It actually LGTM largely, but I have some questions that I don't want overlooked

micro-rdk/src/common/robot.rs
```diff
@@ -242,6 +275,31 @@ impl LocalRobot {
             components.map_err(RobotError::RobotParseConfigError)?,
             registry,
         )?;
+
+        // TODO: When cfg's on expressions are valid, remove the outer scope.
+        #[cfg(feature = "data")]
+        {
+            match DataManager::<StaticMemoryDataStore>::from_robot_and_config(&robot, config_resp) {
+                Ok(Some(mut data_manager)) => {
```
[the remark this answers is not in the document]
acmorrow: Ownership of the DataManager did prove troublesome and I wouldn't be opposed to a refactoring, but probably not as part of this review. Note that things still aren't entirely right maybe, as I'm currently getting a clippy error for the handling of `data_manager_collection_task`.

micro-rdk/src/common/robot.rs
```diff
@@ -446,6 +504,17 @@ impl LocalRobot {
...
 
+    pub fn get_periodic_app_client_tasks(&mut self) -> Vec<Box<dyn PeriodicAppClientTask>> {
+        let mut tasks = Vec::<Box<dyn PeriodicAppClientTask>>::new();
+
+        #[cfg(feature = "data")]
+        if let Some(dm_sync_task) = self.data_manager_sync_task.take() {
+            tasks.push(dm_sync_task);
+        }
+
+        tasks
```
mattjperez, line 515: will we be moving the restart and config-change monitors from the entrypoints here as well?

micro-rdk/src/common/robot.rs
```diff
@@ -217,17 +241,26 @@ impl LocalRobot {
...
             resources: ResourceMap::new(),
             // Use date time pulled off gRPC header as the `build_time` returned in the status of
             // every resource as `last_reconfigured`.
             build_time,
+
             #[cfg(feature = "data")]
             data_collector_configs: vec![],
+            #[cfg(feature = "data")]
+            data_manager_sync_task: None,
+            #[cfg(feature = "data")]
```
mattjperez, line 262: this isn't necessary since the field isn't feature-gated anywhere else and is referenced in `Drop`

micro-rdk/src/common/conn/server.rs
```diff
@@ -356,9 +354,18 @@ where
...
         let cloned_robot = robot.clone();
 
+        if let Some(app_client) = app_client {
+            self.app_client.write().await.replace(app_client);
+        }
+
+        // Let the robot register any periodic app client tasks it may
+        // have based on its configuration.
+        self.app_client_tasks
+            .append(&mut robot.lock().unwrap().get_periodic_app_client_tasks());
```
mattjperez, line 367: So is the idea to load feature-gated tasks within this function in `LocalRobot`, with user-defined tasks loaded through the entrypoint?

mattjperez (commented):
Looks great! Just some questions

micro-rdk/src/esp32/entry.rs
```diff
@@ -337,55 +369,105 @@ async fn serve_async_with_external_network<S>(
...
-                        "Externally managed network, not connected yet cause {:?}",
-                        error
-                    );
-                    Timer::after(Duration::from_secs(3)).await;
-                    continue;
+    let app_client = 'provisioned: loop {
+        if storage.has_robot_credentials() {
+            log::info!("Found cached robot credentials; attempting to serve");
+
+            if storage.has_robot_configuration() {
```
npmenard, line 394: nit: merge both if with &&

micro-rdk/src/esp32/entry.rs
```diff
@@ -337,55 +369,105 @@ async fn serve_async_with_external_network<S>(
...
+                if let RobotRepresentation::WithRegistry(ref registry) = repr {
+                    log::info!("Found cached robot configuration; speculatively building robot from config");
+                    match LocalRobot::from_cloud_config(
```
npmenard, line 397: feel's like this might be better placed outside the loop? Also i am not sure i am reading this correctly but there is a case where when the robot config is cached, the esp32 starts the robot would be created based on the cached config so if a newer config exists it won't be applied? And the time would not be properly set?

micro-rdk/src/esp32/entry.rs
```diff
@@ -337,55 +369,105 @@ async fn serve_async_with_external_network<S>(
...
+                                // Record the last error so that we can serve it once we reach provisioning.
+                                let _ = last_error.insert(e);
+                                break;
+                            }
+                        }
+
+                        // For all other errors, we assume we could not communicate with app due
+                        // to network issues, and just restart the inner loop until we are able
+                        // to communicate with app.
+                        log::info!("Unable to validate robot credentials {:?}; will retry", e);
```
npmenard, line 464: this means that as long as we don't have network connectivity at start the robot won't run?

npmenard (changes requested):
Great!
Done with first pass
I think the immediate next step is to rewrite the entry path. We have all the components now just a matter of putting them together

micro-rdk-server/esp32/main.rs
```diff
@@ -57,21 +58,44 @@ mod esp32 {
...
+            use micro_rdk::common::provisioning::storage::RobotCredentials;
+            use micro_rdk::common::provisioning::storage::WifiCredentialStorage;
+            use micro_rdk::common::provisioning::storage::WifiCredentials;
+
+            log::warn!("Unconditionally using build-time WiFi and robot configuration");
+
+            let storage = NVSStorage::new("nvs").unwrap();
+            log::info!("Storing static values from build time wifi configuration to NVS");
+            storage
+                .store_wifi_credentials(WifiCredentials::new(
```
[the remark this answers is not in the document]
acmorrow: True, but it is only for compile time wifi config, so mostly only for devs? Perhaps we could make the storage `store_x` methods eliminate redundant stores transparently? The RAM storage doesn't care, but NVS does, so I feel like it is better to push that logic into the Storage type and not burden all callers with doing a read-compare-update block every time they want to interact with storage.

micro-rdk/src/common/conn/server.rs
```diff
@@ -356,9 +354,18 @@ where
             network,
         }
     }
-    pub async fn serve(&mut self, robot: Arc<Mutex<LocalRobot>>) {
+    pub async fn serve(&mut self, robot: Arc<Mutex<LocalRobot>>, app_client: Option<AppClient>) {
```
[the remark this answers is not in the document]
acmorrow: Very likely so. It was on my list of small improvements to try but I didn't pursue it in the interest of getting this out for review. I'll try it and if it works I'll make the change. I expect it will.

micro-rdk/src/common/app_client.rs
```diff
@@ -293,9 +300,14 @@ impl AppClient {
             None
         };
 
-        let r = r.split_off(5);
+        let cfg_response = ConfigResponse::decode(r.split_off(5))?;
+        if let Some(robot_config) = cfg_response.config.as_ref() {
+            if let Some(cloud_config) = robot_config.cloud.as_ref() {
+                self.config.set_rpc_host(cloud_config.fqdn.clone());
```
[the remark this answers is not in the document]
acmorrow: Yes, this was among the things I wasn't too pleased with. The old path for this happening was in `ViamServer::build`, but that was too late in one particular circumstance that I'm having a hard time remembering the details. When I have a little more time, I'll revert it and write up the exact situation further. But yeah, I don't love it either.

micro-rdk/src/esp32/entry.rs
```diff
@@ -193,46 +146,27 @@ pub async fn serve_web_inner(
...
+    .with_periodic_app_client_task(Box::new(RestartMonitor::new(|| unsafe {
+        crate::esp32::esp_idf_svc::sys::esp_restart()
+    })))
+    .build(&cfg_response)
+    .unwrap();
 
-    futures_lite::future::zip(Box::pin(srv.serve(robot)), data_future).await;
-}
+    // Attempt to cache the config for the machine we are about to `serve`.
+    if let Err(e) = storage.store_robot_configuration(cfg_response.config.unwrap()) {
```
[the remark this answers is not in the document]
acmorrow: As above, would you be OK with having `store_robot_configuration` drop redundant writes by doing a read first?

micro-rdk/src/common/robot.rs
acmorrow: I wasn't planning to, but we could. I had to move the data manager ones because they need to exist on a robot created before we have network access. The restart monitor and config changes monitors do not have that constraint.

micro-rdk/src/common/robot.rs
acmorrow: Good catch, I will fix that.

micro-rdk/src/common/conn/server.rs
acmorrow: Nothing so organized was intended. I was just doing what I had to do to get a robot running data collection before the network was able to access app.viam.

micro-rdk/src/esp32/entry.rs
```diff
@@ -61,127 +66,75 @@ pub async fn serve_web_inner(
...
-        &cfg_response,
-        &app_config,
-        robot.clone(),
-    ) {
-        Ok(svc) => svc,
-        Err(err) => {
-            log::error!("error configuring data management: {:?}", err);
-            None
+                    //TODO shouldn't panic here, when we support offline mode and reloading configuration this should be removed
+                    panic!("couldn't build robot");
```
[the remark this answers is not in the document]
acmorrow: I will check.

micro-rdk/src/esp32/entry.rs
acmorrow: There is logic further down that executes in the has_robot_credentials case independently of has_robot_configuration, so I don't think they can be combined. See the code starting at `let mut Duration = ...`.

micro-rdk/src/esp32/entry.rs
acmorrow: Sorry: hit the button too soon. If you want, I can try to swap the loop around to do a connectivity check first and then only build from cache on failure. Now that I look at the loop it may not be that hard. Let me know if you feel strongly enough about it that you want me to fold it in for this review. Alternatively, we could lock in the current gains and work on that in a subsequent iteration.

micro-rdk/src/esp32/entry.rs
acmorrow: No, if there are both cached credentials and configuration present the robot will be created at line 397 (new). This loop will then spin attempting to communicate with app to validate. While the loop is spinning the robot will be collecting data. If/when app becomes reachable, if the robot validates then we are provisioned and we move on to `serve_web_inner`. Otherwise, if validation fails due to an explicit rejection of the credentials, the robot gets dropped and we move on to provisioning. Or, it is considered some sort of transient networking error and the loop keeps looping. Or, at least that is the intention!

micro-rdk/src/common/app_client.rs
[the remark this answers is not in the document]
npmenard: Yeah let's make a ticket to revert that change later on, the only concern I have is how are we going to handle config checks. But in the worst case we can create an extra function for that purpose in the meantime

micro-rdk/src/common/registry.rs
npmenard: I agree it's not ideal. We should keep track of this with a ticket if we want to leave this as is when merging

micro-rdk/src/esp32/entry.rs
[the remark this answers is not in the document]
npmenard: I think this is the way. In this particular case we should consider storing an hash (FNV32) of the config for faster comparison instead of doing a byte by byte.

micro-rdk/src/esp32/entry.rs
npmenard: I just wanted to confirm I was reading this correctly, i don't think we should the current behavior. And it makes sense to restart & rebuild once we have config checks

acmorrow: Just checking, but do you mean "don't think we should **change** the current behavior"?

micro-rdk/src/esp32/entry.rs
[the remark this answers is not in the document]
acmorrow: https://viam.atlassian.net/browse/RSDK-8120

micro-rdk/src/common/app_client.rs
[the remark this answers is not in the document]
acmorrow: https://viam.atlassian.net/browse/RSDK-8121

micro-rdk-server/esp32/main.rs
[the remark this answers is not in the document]
acmorrow: https://viam.atlassian.net/browse/RSDK-8120

micro-rdk/src/common/conn/server.rs
acmorrow: @mattjperez - OK to resolve?

micro-rdk/src/common/provisioning/storage.rs
```diff
@@ -3,22 +3,33 @@ use std::{convert::Infallible, error::Error, fmt::Debug, rc::Rc, sync::Mutex};
...
+        app::v1::RobotConfig,
+        provisioning::v1::{CloudConfig, SetNetworkCredentialsRequest},
+    },
 };
 
 #[derive(Clone, Default, Debug)]
 pub struct RobotCredentials {
-    pub(crate) robot_secret: String,
     pub(crate) robot_id: String,
+    pub(crate) robot_secret: String,
```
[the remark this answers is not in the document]
acmorrow: @gvaradarajan - OK to resolve?

micro-rdk/src/common/registry.rs
acmorrow: https://viam.atlassian.net/browse/RSDK-8122

micro-rdk/src/common/robot.rs
acmorrow: https://viam.atlassian.net/browse/RSDK-8123

micro-rdk/src/common/robot.rs
acmorrow: @mattjperez - OK to resolve?

micro-rdk/src/common/robot.rs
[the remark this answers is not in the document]
acmorrow: @gvaradarajan - What should be the next steps here? Do you want a ticket?

micro-rdk/src/esp32/entry.rs
```diff
@@ -9,48 +9,53 @@ use std::{
...
+        grpc::ServerError,
+        provisioning::server::ProvisioningInfo,
+        provisioning::storage::{RobotConfigurationStorage, WifiCredentialStorage},
+    },
+    proto::app::v1::ConfigResponse,
 };
 
-pub async fn serve_web_inner(
-    robot_creds: RobotCredentials,
+pub async fn serve_web_inner<S>(
```
acmorrow, line 45: Note that `serve_web_inner` between `esp32` and `native` is now almost identical mod some types. I've filed https://viam.atlassian.net/browse/RSDK-8124 to track that work.

micro-rdk/src/common/robot.rs
```diff
@@ -242,6 +275,31 @@ impl LocalRobot {
...
+        // TODO: When cfg's on expressions are valid, remove the outer scope.
+        #[cfg(feature = "data")]
+        {
+            match DataManager::<StaticMemoryDataStore>::from_robot_and_config(&robot, config_resp) {
+                Ok(Some(mut data_manager)) => {
+                    let _ = robot
+                        .data_manager_sync_task
+                        .insert(Box::new(data_manager.get_sync_task()));
+
+                    let _ = robot
```
acmorrow, line 288: There is currently a clippy error here that I don't fully understand:
```
error: non-binding `let` on a future
   --> micro-rdk/src/common/robot.rs:289:21
    |
289 | / ...   let _ = robot
290 | | ...       .data_manager_collection_task
291 | | ...       .insert(robot.executor.spawn(async move {
292 | | ...           if let Err(err) = data_manager.data_collection_task().await {
293 | | ...               log::error!("error running data manager: {:?}", err)
294 | | ...           }
295 | | ...       }));
    | |______________^
    |
    = help: consider awaiting the future or dropping explicitly with `std::mem::drop`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_future
    = note: `-D clippy::let-underscore-future` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::let_underscore_future)]`

error: could not compile `micro-rdk` (lib) due to previous error
```

acmorrow:
Updated, PTAL. I believe I've addressed all comments with either code changes or deferred work in tickets. There is still one clippy error I need to make sense of before this can merge.

micro-rdk/src/common/robot.rs
acmorrow: @mattjperez and I figured this out. Using `insert` here was not quite right, using `replace` placated clippy.

micro-rdk/src/lib.rs
```diff
@@ -43,36 +43,34 @@ pub mod google {
 
 /// gRPC prototypes from definitions in [api repository](https://github.com/viamrobotics/api/tree/main/proto/viam), auto-generated
 pub mod proto {
+
```
acmorrow, line 46: @npmenard - I had a newish nightly rust installed, and clippy started throwing some new complaints about the generated proto code. I didn't see much utility in needing to individually suppress in each instance as we were doing before, so I just took a big hammer to clippy w.r.t. generated code. Please let me know if you dislike that choice and I can go the other way.

micro-rdk/src/esp32/exec.rs
```diff
@@ -9,7 +9,7 @@ use futures_lite::{
 pub struct Esp32Executor {}
 
 std::thread_local! {
-    static EX: LocalExecutor<'static> = LocalExecutor::new();
+    static EX: LocalExecutor<'static> = const { LocalExecutor::new() };
```
acmorrow, line 12: Native clippy with new nightly rust made this suggestion for the native executor, so I made it here too just for consistency.

micro-rdk/src/native/exec.rs
```diff
@@ -12,7 +12,7 @@ use crate::common::{provisioning::server::ProvisioningExecutor, webrtc::exec::We
 pub struct NativeExecutor {}
 
 std::thread_local! {
-    static EX: LocalExecutor<'static> = LocalExecutor::new();
+    static EX: LocalExecutor<'static> = const { LocalExecutor::new() };
```
acmorrow, line 15: Native clippy with nightly rust suggested this.

micro-rdk/src/native/tls.rs
```diff
@@ -60,23 +60,11 @@ impl TlsClientConnector for NativeTls {
...
 }
 
-use rustls::KeyLog;
-
 use crate::common::conn::errors::ServerError;
 use crate::common::conn::server::TlsClientConnector;
 
 use super::tcp::NativeStream;
 
-struct Key {}
```
acmorrow, line 70: Native clippy with nightly rust noticed this was dead code.

micro-rdk/src/common/webrtc/ice.rs
```diff
@@ -370,7 +370,7 @@ impl ICEAgent {
     /// to do so it parses the pair list in the following manner
     /// 1) If a pair has no pending STUN request it generates an TransactionId and attach to the pair
     /// 2) If a pair has a pending STUN request and its timeout is elapsed it will resend
-    /// the generated TransactionId
+    ///    the generated TransactionId
```
acmorrow, line 373: Native clippy with nightly rust wanted this indented.

acmorrow (commented):
Clippy fixes

micro-rdk/src/common/robot.rs
```diff
@@ -87,12 +101,21 @@ impl ResourceType {
...
 }
 
+#[cfg(feature = "native")]
+type Executor = NativeExecutor;
+#[cfg(feature = "esp32")]
+type Executor = Esp32Executor;
+
 #[derive(Default)]
 pub struct LocalRobot {
+    pub part_id: String,
```
npmenard, line 111: nit: pub(crate)

npmenard (approved):
LGTM just some nits and comments

micro-rdk/src/common/robot.rs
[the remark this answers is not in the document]
gvaradarajan: yeah I never wanted it to be part of this PR. We should maybe hold off on the ticket; after we ship to DY I want us to do a little bit of a review of the architecture and what changes we want to make as a team

micro-rdk/src/common/provisioning/storage.rs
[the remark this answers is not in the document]
gvaradarajan: yes

micro-rdk/src/common/robot.rs
```diff
@@ -242,6 +275,31 @@ impl LocalRobot {
             components.map_err(RobotError::RobotParseConfigError)?,
             registry,
         )?;
+
+        // TODO: When cfg's on expressions are valid, remove the outer scope.
+        #[cfg(feature = "data")]
+        {
+            match DataManager::<StaticMemoryDataStore>::from_robot_and_config(&robot, config_resp) {
```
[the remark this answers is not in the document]
gvaradarajan: yeah that all makes sense, thanks for making all these tickets

gvaradarajan (approved):
LGTM, thank you for making all of those follow-up tickets

micro-rdk/src/common/robot.rs
acmorrow: Done

micro-rdk/src/lib.rs
acmorrow: That's my take, that there isn't much point to clippy for generated code, and we will just end up bouncing back and forth between needing to add and remove specific suppressions as clippy and prost each advance versions. If you still want a ticket let me know and I'll file one, but I think I'm going to merge this as is.

PR #252, opened 2024-07-03
Rsdk 7388 add idf component

npmenard:
Add the idf component + an example


examples/esp-idf-component/custom_partition.csv
```diff
@@ -0,0 +1,9 @@
+#
+# Custom partition format for two OTA partions
```
mattjperez, line 2: This partition table currently has only one OTA partition, correct?

examples/esp-idf-component/sdkconfig
```diff
@@ -0,0 +1,1551 @@
+#
```
gvaradarajan, line 1: should we use sdkconfig.defaults instead?

examples/esp-idf-component/main/main.c
```diff
@@ -0,0 +1,248 @@
+#include <stdio.h>
```
gvaradarajan, line 1: I think this (and micrordklib-idf-component) could maybe use a README? Some instructions for building and running?

gvaradarajan (changes requested):
first pass

micro-rdk-ffi/micrordklib-idf-component/CMakeLists.txt
```diff
@@ -0,0 +1,39 @@
+idf_component_register(INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}")
+idf_build_set_property(INCLUDE_DIRECTORIES "${CMAKE_CURRENT_BINARY_DIR}/assets" APPEND)
+
+set(LIBMICRORDK_VERSION v0.2.1-rc1)
```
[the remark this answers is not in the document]
npmenard: I don't like this either. But I want the idf_component version to be strictly tied to a libmicrordk version otherwise it will be hard to understand which versions user are running. Also I don't know if cmake is smart enough to automatically pull the latest version without erasing the build directory & rebuilding.

examples/esp-idf-component/custom_partition.csv
npmenard: no the comment in the file is out of place :P

PR #253, opened 2024-07-03
RSDK-8073 - weak C hook for callback to run on successful data upload

gvaradarajan:


micro-rdk-ffi/build.rs
```diff
@@ -113,6 +113,6 @@ fn main() {
             .with_config(cfg)
             .generate()
             .expect("Unable to generate bindings")
-            .write_to_file("include/micrordk.h");
```
npmenard, line 116: why change this?

micro-rdk/src/common/data_manager.rs
```diff
@@ -383,6 +398,10 @@ where
                                 current_chunk_size = next_message.len();
                                 current_chunk = vec![next_message];
                             }
+                            #[cfg(feature = "esp32")]
```
npmenard, line 401: doesn't have to be feature gated

micro-rdk/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 // TODO this check is disabled but shouldn't
 #![allow(clippy::arc_with_non_send_sync)]
+#![cfg_attr(feature = "esp32", feature(linkage))]
```
npmenard, line 3: is it a nightly only feature? if so then we have to feature-gate it but let's leave a comment so we know why

micro-rdk/src/common/data_manager.rs
```diff
@@ -27,6 +27,21 @@ use thiserror::Error;
 // the smaller amount of available RAM, we've halved it
 static MAX_SENSOR_CONTENTS_SIZE: usize = 32000;
 
+/// This technically allows for a C project using micro-RDK as a library
+/// to implement a callback to be run whenever data has successfully been uploaded.
```
npmenard, line 31: nit: Allow for a program using Micro-RDK as a library to implement...

micro-rdk/src/common/data_manager.rs
```diff
@@ -27,6 +27,21 @@ use thiserror::Error;
...
+/// The callback should be identical in signature to the weak symbol declared here
+/// Ex. for ESP-IDF project
+/// void data_upload_callback(void) {
+///     ESP_LOGI("proj_main", "hit data upload callback");
+/// }
+/// # Safety
+/// implemented callback must manage its own memory, micro-RDK will
```
npmenard, line 38: this is a bit implied :P But I would just say : the callback should be kept simple and not do any blocking or long operation

npmenard (commented):
Done with first pass

micro-rdk-ffi/build.rs
gvaradarajan: oh I forgot to undo this change, sorry. I needed it temporarily for my local idf-component example setup to work. But I assume that problem will be fixed when your other PR moving the component to this repo goes in.

micro-rdk/src/lib.rs
gvaradarajan: yes that's why I'll leave a comment

micro-rdk/src/common/data_manager.rs
gvaradarajan: It does because I had to feature-gate the weak symbol in the first place, but I suppose I could make an alternate plain empty implementation of `data_upload_callback` when the esp32 feature is not active. Is this preferred?

npmenard: make sense

npmenard:
LGTM. If possible I would have the symbols be absent during binary build assuming there is a way to detect build type during compilation

micro-rdk/src/common/data_manager.rs
```diff
@@ -27,6 +27,23 @@ use thiserror::Error;
...
+/// to the weak symbol declared here
+///
+/// Ex. for ESP-IDF project
+/// void data_upload_callback(void) {
+///     ESP_LOGI("proj_main", "hit data upload callback");
+/// }
+///
+/// # Safety
+/// The implemented callback should be simple. Long-running operations
+/// would endanger the stability of the data management feature
```
npmenard, line 41: nit: remove `would endanger the stability of the data management feature` and just put `Long-running or blocking operations should be avoided`

npmenard (approved):

micro-rdk/src/common/data_manager.rs
acmorrow: Or even `must be avoided`. Also should probably be illegal to re-enter micro-rdk or destroy resources in use by micro-rdk. Care should be taken regarding taking any locks to prevent potential deadlocks.

micro-rdk/src/common/data_manager.rs
```diff
@@ -27,6 +27,23 @@ use thiserror::Error;
...
+#[cfg(feature = "esp32")]
+#[linkage = "weak"]
+#[no_mangle]
+pub unsafe extern "C" fn data_upload_callback() {}
```
acmorrow, line 45: All quibbles but:

- I'd call this a "hook" not a "callback".
- The name right now doesn't tell me exactly when the call is made: before data is uploaded or after?
- I also think the function should include `micro-rdk` in the name to keep it namespaced, and `data_manager` to connect it to the struct that uses it.

So, maybe `micro_rdk_data_manager_post_upload_hook`?

micro-rdk/src/common/data_manager.rs
```diff
@@ -383,6 +400,10 @@ where
                                 current_chunk_size = next_message.len();
                                 current_chunk = vec![next_message];
                             }
+                            #[cfg(feature = "esp32")]
+                            unsafe {
+                                data_upload_callback();
```
acmorrow, line 405: Is there any metadata we want to provide to the callback? Size of the upload? Time upload started? Duration of upload? Is there any information we want to be able to have the callback provide back to us? Once we release this API we won't be able to change it easily without breaking ABI.

acmorrow (changes requested):

micro-rdk/src/common/data_manager.rs
npmenard: I don't think so. It's here mainly for debugging data upload from a C-FFI side and not intended to be used in normal ops . We should consider whether a more general API make senses longer term but that's a scope doc. But to your point we can feature gate it with something like "unstable/data-upload-hook" and not enable it by default so its easier to yank it in the future

gvaradarajan:
> LGTM. If possible I would have the symbols be absent during binary build assuming there is a way to detect build type during compilation

@npmenard I'm not sure I understand this request. Are you saying you would like to have the symbol for the callback hook not be generated if we are using micro-RDK "normally"?

npmenard:
> > LGTM. If possible I would have the symbols be absent during binary build assuming there is a way to detect build type during compilation
> 
> @npmenard I'm not sure I understand this request. Are you saying you would like to have the symbol for the callback hook not be generated if we are using micro-RDK "normally"?

yes basically for binary builds of micro rdk i don't want the symbol to be emitted (basically feature gate it on build type) But I don't think we can do that in rust right now so feel free to ignore

micro-rdk/src/common/data_manager.rs
```diff
@@ -27,6 +27,23 @@ use thiserror::Error;
 // the smaller amount of available RAM, we've halved it
 static MAX_SENSOR_CONTENTS_SIZE: usize = 32000;
 
+/// Allow for a C project using micro-RDK as a library to implement a callback to be run
+/// whenever data has successfully been uploaded. The callback should be identical in signature
+/// to the weak symbol declared here
+///
+/// Ex. for ESP-IDF project
+/// void data_upload_callback(void) {
```
acmorrow, line 35: Name in this comment needs to be updated to `micro_rdk_data_manager_post_upload_hook`

micro-rdk/src/common/data_manager.rs
```diff
@@ -27,6 +27,23 @@ use thiserror::Error;
...
+///     ESP_LOGI("proj_main", "hit data upload callback");
+/// }
+///
+/// # Safety
+/// The implemented callback should be simple and avoid using resources managed by
+/// micro-RDK. Long-running operations should be avoided.
```
acmorrow, line 41: I still think there should be language here about disallowing reentrant calls and being careful with taking locks.

acmorrow (approved):
LGTM mod some suggestions for changes to the comments

PR #255, opened 2024-07-08
[RSDK-7043] Config restart

allisonschiang:
Decided to compare configrequest instead of comparing hashed configrequest because:
1. Guaranteeing order of configrequest struct (after calling get_config) involves a lot of work and I tried a lot of different approaches : (
2. this work will be replaced by adding version into the config request and just comparing that anyways very soon

Draft PR because debugging a core dump error???


micro-rdk/src/esp32/entry.rs
```diff
@@ -205,7 +206,11 @@ pub async fn serve_web_inner(
         .with_webrtc(webrtc)
         .with_periodic_app_client_task(Box::new(RestartMonitor::new(|| unsafe {
             crate::esp32::esp_idf_svc::sys::esp_restart()
-        })));
+        })))
+        .with_periodic_app_client_task(Box::new(ConfigMonitor::new(||unsafe {
```
acmorrow, line 210: I expect this review to land after #251 which will add config caching. If we detect a config change, consider whether we should write the new config to NVS before restarting. That way, if we happen to come back up and find we can't talk to WiFi anymore, the robot will still start up with the newest configuration.

micro-rdk/src/esp32/entry.rs
```diff
@@ -205,7 +206,11 @@ pub async fn serve_web_inner(
...
+            crate::esp32::esp_idf_svc::sys::esp_restart()
+        },  *(cfg_response.clone()))))
+        ;
```
acmorrow, line 213: Formatting looks a little off, could you please run `cargo fmt`? I also recommend configuring your editor to do it for you automatically.

micro-rdk/src/esp32/entry.rs
```diff
@@ -21,6 +21,7 @@ use crate::common::{
     provisioning::storage::RobotCredentials,
     restart_monitor::RestartMonitor,
     robot::LocalRobot,
+    config_monitor::ConfigMonitor,
```
acmorrow, line 24: Looks out of alpha order

micro-rdk/src/common/config_monitor.rs
```diff
@@ -0,0 +1,56 @@
...
+        })
+    }
+
+}
+
+async fn get_app_config(client : &AppClient) -> Result<ConfigResponse, AppClientError>{
+    let (new_config, _cfg_received_datetime) = 
+            client.get_config(None).await.unwrap();
+    return Ok(*new_config);
+}
```
acmorrow, line 55: I'm not sure this function is pulling its weight. Is there a reason the body can't just be inlined into `invoke` above?

micro-rdk/src/common/app_client.rs
```diff
@@ -255,19 +255,18 @@ impl AppClient {
...
             ips: vec![ip.to_string()],
             version: env!("CARGO_PKG_VERSION").to_string(),
             git_revision: "".to_string(),
             platform: Some("esp32".to_string()),
-        };
-
+        });
         let req = ConfigRequest {
-            agent_info: Some(agent),
+            agent_info: agent,
```
acmorrow, line 269: What are the practical effects of the `agent_info` being `None` here when invoking the `/viam.app.v1.RobotService/Config` endpoint, if any?

micro-rdk/src/common/mod.rs
```diff
@@ -59,6 +59,7 @@ pub mod mpu6050;
 pub mod power_sensor;
 pub mod registry;
 pub mod restart_monitor;
+pub mod config_monitor;
```
acmorrow, line 62: alphabetize

micro-rdk/src/esp32/entry.rs
```diff
@@ -237,7 +242,7 @@ async fn validate_robot_credentials(
 
 // Four cases:
 // 1) No Robot Credentials + WiFi without external network
-// 2) No Robot Credentials with external network\
+// 2) No Robot Credentials with external network
```
acmorrow, line 245: Nice catch

micro-rdk/src/native/entry.rs
```diff
@@ -76,7 +76,7 @@ pub async fn serve_web_inner(
         );
 
         let (cfg_response, cfg_received_datetime) =
-            client.get_config(network.get_ip()).await.unwrap();
+            client.get_config(Some(network.get_ip())).await.unwrap();
```
acmorrow, line 79: Commenting here because I can't elsewhere, but I'd expect to see a ConfigMonitor added for native as well. Is there a reason one was not?

micro-rdk/src/common/config_monitor.rs
```diff
@@ -0,0 +1,56 @@
...
+                false => Ok(Some(self.get_default_period())),
+            }
+        })
+    }
+
+}
+
+async fn get_app_config(client : &AppClient) -> Result<ConfigResponse, AppClientError>{
+    let (new_config, _cfg_received_datetime) = 
+            client.get_config(None).await.unwrap();
```
acmorrow, line 53: An interesting case to consider is what currently does (and what should) happen if the configuration becomes invalid (e.g. the robot was deleted), such that you can no longer obtain it with `get_config`.

acmorrow (changes requested):
Looks like a great start! Please run `cargo fmt`, and I suggest applying whatever configuration is necessary in your editor of choice to automate running it.

micro-rdk/src/common/app_client.rs
allisonschiang: image
It's used to show app info, but if we don't supply it, it should still work fine

micro-rdk/src/native/entry.rs
allisonschiang: I wanted to get one working before I added it to another : ) will add in my next commit!

micro-rdk/src/common/config_monitor.rs
acmorrow: I don't think you need to make `invoke` an async function: it already "is" one in the sense that it returns a Future, but it can't be declared as one due to limitations in rust. However, note that on line 40 you are inside an `async` block, and as far as I know, inside there you can definitely use `await` (it already does).

micro-rdk/src/esp32/entry.rs
allisonschiang: Sounds like a plan, should this be in a new ticket? Not 100% sure about ticket splitting rules but would feel nice to have this PR closed with a working config restart before looking into NVS update

acmorrow: Yes, I'd suggest filing a new ticket, and then making it depend on the ticket for this PR and the ticket for #251 .

micro-rdk/src/common/config_monitor.rs
allisonschiang: Right now, I believe it core dumps and has a error within the monitor, but if the config is invalid we could just grab it from the config cache (in the next ticket as discussed above)?

allisonschiang: do we store a default empty config in config cache in case the config is invalid on the very first config initial setup?

micro-rdk/src/common/app_client.rs
```diff
@@ -255,19 +255,18 @@ impl AppClient {
     // `last_reconfigured` values for resource statuses.
     pub async fn get_config(
         &self,
-        ip: Ipv4Addr,
+        ip: Option<Ipv4Addr>,
     ) -> Result<(Box<ConfigResponse>, Option<DateTime<FixedOffset>>), AppClientError> {
-        let agent = AgentInfo {
+        let agent = ip.map(|ip| AgentInfo {
```
npmenard, line 260: before this is merged we should make sure the absence of AgentInfo as no effect on what is displayed on app.viam.com

micro-rdk/src/common/config_monitor.rs
```diff
@@ -0,0 +1,48 @@
...
+impl<'a> ConfigMonitor<'a> {
+    pub fn new(restart_hook: impl FnOnce() + 'a, curr_config: ConfigResponse) -> Self {
+        Self {
+            restart_hook: Some(Box::new(restart_hook)),
+            curr_config,
+        }
+    }
+
+    fn restart(&mut self) -> ! {
+        log::warn!("Robot configuration change detected");
```
npmenard, line 21: nit: `Robot configuration change detected restarting micro-rdk`

micro-rdk/src/native/entry.rs
```diff
@@ -168,7 +169,11 @@ pub async fn serve_web_inner(
                 .with_webrtc(webrtc)
                 .with_periodic_app_client_task(Box::new(RestartMonitor::new(|| {
                     std::process::exit(0)
-                })));
+                })))
+                .with_periodic_app_client_task(Box::new(ConfigMonitor::new(
+                    || std::process::exit(0),
```
npmenard, line 174: not a true restart :P (but ok to leave as it)

micro-rdk/src/common/config_monitor.rs
acmorrow: No. I filed https://viam.atlassian.net/browse/RSDK-8127 to track this general area.

micro-rdk/src/common/config_monitor.rs
```diff
@@ -0,0 +1,48 @@
...
+        Duration::from_secs(10)
+    }
+
+    fn invoke<'c, 'b: 'c>(
+        &'b mut self,
+        app_client: &'c AppClient,
+    ) -> Pin<Box<dyn Future<Output = Result<Option<Duration>, AppClientError>> + 'c>> {
+        Box::pin(async move {
+            let (app_config, _cfg_received_datetime) = app_client.get_config(None).await.unwrap();
+            match self.curr_config == *app_config {
```
npmenard, line 42: you may have already done it but add a comment +ticket saying this should be changed when config version exists.

micro-rdk/src/esp32/entry.rs
acmorrow: FYI #251 merged, so you will likely have some merge conflicts to deal with before you can merge this.

acmorrow: Also, please post the new ticket here after you file it.

allisonschiang: https://viam.atlassian.net/browse/RSDK-8159

Done!

micro-rdk/src/common/config_monitor.rs
allisonschiang: [RSDK-8160](https://viam.atlassian.net/browse/RSDK-8160)

micro-rdk/src/common/app_client.rs
allisonschiang: From my testing, it doesn't change what is displayed.

acmorrow (approved):

PR #259, opened 2024-07-09
log ssid to output

mattjperez:
while going through provisioning with @EshaMaharishi, we couldn't find device mac address in monitor output; tries to print MAC in monitor before MAC is set with FUSE at L33 here. This just makes things clear. should we also log the default provisioning password while we're at it?


EshaMaharishi (approved):

acmorrow (approved):
Seems fine to me but I'd like @npmenard's review too just in case I'm overlooking something.

npmenard (approved):
LGTM, let's add the AP password too

mattjperez:
confirmed output:
`I (32) micro_rdk::esp32::provisioning::wifi_provisioning: Provisioning SSID: esp32-micrordk-5FE8 - Password: viamsetup`
merging now

PR #260, opened 2024-07-11
[RSDK-7820] Camera on esp32

mattjperez:
with config for esp32 m5 camera timer x, ov3600 camera
```json
"components": [
    {
      "name": "esp32camera",
      "namespace": "rdk",
      "type": "camera",
      "model": "esp32-camera",
      "attributes": {
        "pin_pclk": 21,
        "pin_href": 26,
        "pin_d6": 36,
        "sccb_i2c_port": -1,
        "pin_d4": 39,
        "pin_d0": 32,
        "pin_sccb_sda": 25,
        "jpeg_quality": 32,
        "pin_vsync": 22,
        "pin_d7": 19,
        "pin_d2": 34,
        "pin_xclk": 27,
        "frame_size": 5,
        "ledc_channel": 1,
        "pin_reset": 15,
        "pin_d5": 18,
        "pin_d1": 35,
        "xclk_freq_hz": 20000000,
        "ledc_timer": 1,
        "pin_d3": 5,
        "pin_sccb_scl": 23
      }
    }
  ]

```


micro-rdk/src/esp32/camera.rs
```diff
@@ -1,22 +1,56 @@
...
-use crate::esp32::esp_idf_svc::systime::EspSystemTime;
-use crate::proto::component::camera;
+use crate::esp32::esp_idf_svc::{
+    sys::camera::{
+        camera_config_t, camera_config_t__bindgen_ty_1, camera_config_t__bindgen_ty_2, camera_fb_t,
+        esp_camera_fb_get, esp_camera_fb_return, esp_camera_init,
+    },
+    systime::EspSystemTime,
+};
+use crate::{
```
acmorrow, line 15: I think you can fold this in to the above `use crate`

micro-rdk/Cargo.toml
```diff
@@ -19,7 +19,7 @@ default = ["builtin-components", "data"]
 binstart = ["esp-idf-svc/binstart"]
 libstart = ["esp-idf-svc/libstart"]
 builtin-components = []
-camera = []
+camera = ["builtin-components"]
```
acmorrow, line 22: I'm not sure I see why. Also wondering: if !builtin-components should you even get the camera component? Maybe you need to ask for both `camera` (to get support) and `builtin-components` (to get the actual camera component)? After all, you might want to build your own camera component in a module but not use ours?

micro-rdk/src/common/camera/mod.rs
```diff
@@ -1,13 +1,20 @@
...
-// Enables FakeCamera for native server
-#[cfg(all(feature = "camera", feature = "native", feature = "builtin-components"))]
+#[cfg(feature = "camera")]
 mod fake_camera;
-#[cfg(all(feature = "camera", feature = "native", feature = "builtin-components"))]
-pub(crate) use fake_camera::register_models;
+
+#[allow(unused)]
+pub(crate) fn register_models(registry: &mut ComponentRegistry) {
+    #[cfg(feature = "camera")]
```
acmorrow, line 11: Yeah I feel like this should still be gated on builtin-components too

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,22 +1,56 @@
...
+        status::{Status, StatusError},
+    },
+    google::{self, api::HttpBody},
+    proto::component::camera,
+};
 use bytes::{Bytes, BytesMut};
 use log::*;
 use prost::Message;
 
+#[derive(DoCommand)]
```
acmorrow, line 29: `Duration` doesn't feel like the right type for a point in time.

micro-rdk/src/esp32/camera.rs
```diff
@@ -51,32 +85,28 @@ impl Esp32Camera {
             last_grab: t.now(),
         }
     }
-    pub fn setup(&self) -> Result<(), CameraError> {
-        let ret = (unsafe { crate::esp32::esp_idf_svc::sys::esp_camera_init(&self.config) })
-            as crate::esp32::esp_idf_svc::sys::esp_err_t;
+    fn setup(&self) -> Result<(), CameraError> {
```
acmorrow, line 88: The `from_config` function above seems to be the only caller, and this `fn` isn't pub. Maybe just pull the logic into `from_config`?

micro-rdk/src/esp32/camera.rs
```diff
@@ -51,32 +85,28 @@ impl Esp32Camera {
...
+        let ret =
+            (unsafe { esp_camera_init(&self.config) }) as crate::esp32::esp_idf_svc::sys::esp_err_t;
         let ret = crate::esp32::esp_idf_svc::sys::EspError::convert(ret);
-        ret.map_err(|e| CameraError::CameraInitError(e.into()))
+        ret.map_err(|e| CameraError::InitError(e.into()))
```
acmorrow, line 92: I suspect this block can be simplified.

micro-rdk/src/esp32/camera.rs
```diff
@@ -51,32 +85,28 @@ impl Esp32Camera {
...
             None
         } else {
             Some(ptr)
         }
     }
-    pub fn return_cam_frame(
-        &self,
-        frame: Option<*mut crate::esp32::esp_idf_svc::sys::camera_fb_t>,
-    ) {
+    pub fn return_cam_frame(&self, frame: Option<*mut camera_fb_t>) {
```
acmorrow, line 102: Is there an obligation to call `return_cam_frame` for each call to `get_cam_frame`? Could you build a resource management type to ensure that automatically?

micro-rdk/src/esp32/camera.rs
```diff
@@ -112,6 +142,38 @@ impl Camera for Esp32Camera {
             self.return_cam_frame(Some(ptr));
             return Ok(buffer);
         }
-        Err(CameraError::CameraCouldntGetFrame)
+        Err(CameraError::FailedToGetImage)
+    }
+
+    fn render_frame(&mut self, mut buffer: BytesMut) -> Result<BytesMut, CameraError> {
+        if let Some(ptr) = self.get_cam_frame() {
```
acmorrow, line 149: If you don't want to create an owner type for the frame, you could at least use `defer!` here to ensure cleanup. Above in `get_image` too

micro-rdk/src/esp32/camera.rs
```diff
@@ -51,32 +85,28 @@ impl Esp32Camera {
...
+        let ret =
+            (unsafe { esp_camera_init(&self.config) }) as crate::esp32::esp_idf_svc::sys::esp_err_t;
         let ret = crate::esp32::esp_idf_svc::sys::EspError::convert(ret);
-        ret.map_err(|e| CameraError::CameraInitError(e.into()))
+        ret.map_err(|e| CameraError::InitError(e.into()))
     }
-    pub fn get_cam_frame(&self) -> Option<*mut crate::esp32::esp_idf_svc::sys::camera_fb_t> {
-        let ptr = (unsafe { crate::esp32::esp_idf_svc::sys::esp_camera_fb_get() })
-            as *mut crate::esp32::esp_idf_svc::sys::camera_fb_t;
+    pub fn get_cam_frame(&self) -> Option<*mut camera_fb_t> {
```
acmorrow, line 94: Does `get_cam_frame` and `return_cam_frame` need to be `pub`?

acmorrow (changes requested):

micro-rdk-server/Cargo.toml
```diff
@@ -55,3 +55,8 @@ esp_idf_sdkconfig_defaults = [
   "micro-rdk-server/sdkconfig.defaults",
   "micro-rdk-server/sdkconfig.defaults.esp32"
 ]
+
+[[package.metadata.esp-idf-sys.extra_components]]
```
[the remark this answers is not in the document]
mattjperez: I'm not sure. According to the [`esp-idf-sys` build options](https://github.com/esp-rs/esp-idf-sys/blob/master/BUILD-OPTIONS.md), this feature needs to be in the Cargo.toml of the workspace or `ESP_IDF_SYS_ROOT_CRATE`, [link](https://github.com/esp-rs/esp-idf-sys/blob/master/BUILD-OPTIONS.md#esp-idf-configuration)

Also, we should note that this modifies the build of `esp-idf-sys` that gets linked to by everything in our workspace.
I don't know how this would affect a user who is using `cargo-generate` to build a project.

micro-rdk/src/common/camera/mod.rs
mattjperez: With the change to the `camera` feature in Cargo.toml in this PR, builtins is active anytime camera is active. Can update this when conclusions are made in the feature dependencies/enablings

micro-rdk/src/esp32/camera.rs
mattjperez: Yup! that's next step when I make the pin settings come dynamic

micro-rdk/src/esp32/camera.rs
mattjperez: it does not. will change

micro-rdk/src/common/camera/mod.rs
mattjperez: following offline conversation
- camera: enable traits
- camera && builtin: enable traits, fake_camera, esp32camera (on esp32)

micro-rdk-server/Cargo.toml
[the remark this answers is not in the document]
mattjperez: We have two options
- user adds block via camera option on cargo generate (recently added)
- we add this block into `micro-rdk` crate but will be included in `esp-idf-sys` even when the `camera` feature is not enabled. Whether this is optimized out by the compiler or not is unclear.

micro-rdk-server/Cargo.toml
```diff
@@ -17,15 +17,15 @@ qemu = []
...
+micro-rdk = { workspace = true, features = ["native", "provisioning", "camera"], default-features = true }
 
 
 [target.'cfg(target_os="espidf")'.dependencies]
 async-channel.workspace = true
 embedded-hal.workspace = true
 embedded-svc.workspace = true
 futures-lite.workspace = true
-micro-rdk = { workspace = true, features = ["esp32", "binstart", "provisioning"], default-features = true }
+micro-rdk = { workspace = true, features = ["esp32", "binstart", "provisioning", "camera"], default-features = true }
```
mattjperez, line 28: remove before merge

micro-rdk/Cargo.toml
mattjperez: following offline conversation
- camera: enable traits
- camera && builtin: enable traits, fake_camera, esp32camera (on esp32)

Cargo.lock
```diff
@@ -328,7 +328,7 @@ dependencies = [
  "cfg-if",
```
npmenard, line 328: any chance we can keep these changes separated from the PR?

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,265 @@
...
-    ) {
-        if let Some(ptr) = frame {
-            unsafe { crate::esp32::esp_idf_svc::sys::esp_camera_fb_return(ptr) }
-        }
+            last_grab: Instant::now(),
+        };
+
+        let ret =
+            (unsafe { esp_camera_init(&cam.config) }) as crate::esp32::esp_idf_svc::sys::esp_err_t;
+        EspError::convert(ret).map_err(|e| CameraError::InitError(e.into()))?;
```
npmenard, line 158: use esp!(unsafe {...})

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,265 @@
...
-                    if i > 0 && i % 80 == 0 {
-                        println!();
-                    }
-                }
-            }
-        }
+
+    fn get_frame(&mut self) -> Result<Esp32CameraFrameBuffer, CameraError> {
+        self.last_grab = Instant::now();
+        let frame = Esp32CameraFrameBuffer::get().ok_or_else(|| CameraError::FailedToGetImage)?;
```
npmenard, line 164: would `.ok_or(CameraError::FailedToGetImage)?` work too?

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,265 @@
...
+    fn height(&self) -> usize {
+        unsafe { (*(self.0)).height }
+    }
+    fn format(&self) -> u32 {
+        unsafe { (*(self.0)).format }
+    }
+    unsafe fn buf(&self) -> *const u8 {
+        (*(self.0)).buf
+    }
+    unsafe fn as_bytes(self) -> Bytes {
```
npmenard, line 255: ideally not unsafe

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,265 @@
...
-            let buf = unsafe {
-                let buf = (*ptr).buf;
-                let len = (*ptr).len as usize;
-                core::slice::from_raw_parts(buf, len)
-            };
-            if buf.len() > buffer.capacity() {
-                self.return_cam_frame(Some(ptr));
-                return Err(CameraError::CameraFrameTooBig);
+    fn get_image(&mut self, mut buffer: BytesMut) -> Result<BytesMut, CameraError> {
+        if let Ok(frame) = self.get_frame() {
```
npmenard, line 171: either remove the Ok(frame) or change get_frame to only return Option<> otherwise you are masking  the error returned by get_frame

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,265 @@
...
+            if frame.len() > buffer.capacity() {
+                return Err(CameraError::ImageTooBig);
             }
-            let bytes = Bytes::from(buf);
+            self.last_grab = Instant::now();
```
npmenard, line 175: what is the purpose of last grab?

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,265 @@
...
+            .get_attribute::<u32>("pixel_format")
+            .unwrap_or(PixelFormat::JPEG as u32);
+        let frame_size = cfg
+            .get_attribute::<u32>("frame_size")
+            .unwrap_or(FrameSize::W240XH240 as u32);
+        // Quality of JPEG output: 0-63 lower means higher quality
+        let jpeg_quality = cfg.get_attribute::<i32>("jpeg_quality").unwrap_or(32);
+        // Number of frame buffers to be allocated. If more than one, then each frame will be acquired (double speed)
+        let fb_count = cfg.get_attribute::<u32>("fb_count").unwrap_or(1) as usize;
+        let grab_mode = cfg.get_attribute::<u32>("grab_mode").unwrap_or(0);
```
npmenard, line 120: no sure this is make sense to have `fb_count` and `grab_mode` configurable considering the way the driver is implemented

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,265 @@
...
+        let fb_location = cfg.get_attribute::<u32>("fb_location").unwrap_or(0);
```
npmenard, line 121: same comment for `fb_location`

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,265 @@
...
+        Err(CameraError::FailedToGetImage)
+    }
+
+    fn render_frame(&mut self, mut buffer: BytesMut) -> Result<BytesMut, CameraError> {
+        if let Ok(frame) = self.get_frame() {
+            if frame.len() > buffer.capacity() {
+                return Err(CameraError::ImageTooBig);
+            }
+            let msg = HttpBody {
+                content_type: "image/jpeg".to_string(),
```
npmenard, line 194: does the format work even if the camera is configured to capture something different than JPEG? If not we might want to commit to JPEG for this version of the driver

npmenard (changes requested):
done with first pass

micro-rdk/src/esp32/camera.rs
mattjperez: it was part of the original code, I was just updating it to use Instant and assigning it when frames are taken. 
want me to remove it?

micro-rdk/src/common/camera/mod.rs
```diff
@@ -1,13 +1,23 @@
-use super::{generic::DoCommand, status::Status};
+use super::{generic::DoCommand, registry::ComponentRegistry, status::Status};
 use bytes::BytesMut;
 use std::sync::{Arc, Mutex};
 use thiserror::Error;
 
-// Enables FakeCamera for native server
-#[cfg(all(feature = "camera", feature = "native", feature = "builtin-components"))]
+#[cfg(feature = "builtin-components")]
 mod fake_camera;
```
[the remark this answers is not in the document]
mattjperez: With the current setup, `FAKE_JPEG` will only be available when both `camera` and `builtin-components` are enabled. otherwise, they're excluded from the module tree during compilation. I made the assumption that as it's not in the tree, it won't be included in the binary.

do you know how to verify this?

micro-rdk/src/esp32/camera.rs
mattjperez: done

micro-rdk/src/common/camera/mod.rs
mattjperez: done

micro-rdk-server/Cargo.toml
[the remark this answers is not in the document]
mattjperez: to clarify, should I remove this or comment it out and add a note to enable when developing builtin-camera?

I'm not sure there's a way to control this block from here with a feature-gate

micro-rdk/Cargo.toml
mattjperez: @acmorrow ok to resolve?

micro-rdk/src/esp32/camera.rs
mattjperez: committing to jpeg for this version

micro-rdk-server/Cargo.toml
```diff
@@ -55,3 +55,8 @@ esp_idf_sdkconfig_defaults = [
   "micro-rdk-server/sdkconfig.defaults",
   "micro-rdk-server/sdkconfig.defaults.esp32"
 ]
+
+[[package.metadata.esp-idf-sys.extra_components]]
+remote_component = { name = "espressif/esp32-camera", version = "2.0.9"}
+bindings_module = "camera"
+bindings_header = "bindings.h"
```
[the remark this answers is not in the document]
mattjperez: from offline conversation, was able to build this in docker as-is. will leave as part of default development (micro-rdk-server) build.

micro-rdk-server/Cargo.toml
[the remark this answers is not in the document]
mattjperez: leaving in as per offline conversation. builds in docker

micro-rdk/src/common/camera/mod.rs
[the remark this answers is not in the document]
npmenard: just look is the symbol gets pulled into the final linked binary. If so we would want to exclude it because it would waste space on flash

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,265 @@
...
-                sccb_i2c_port: 0,
+                sccb_i2c_port,
             },
-            last_grab: t.now(),
+        };
+
+        /*
+        unsafe {
+            let sensor = esp_camera_sensor_get();
+            if (*sensor).id.PID == 0x3660 {
```
npmenard, line 156: remove?

npmenard (approved):
LGTM just a concern about FAKE_JPEG being in the final binary

mattjperez:
Merging after #261 to make sure model naming is consistent and work properly.

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,254 @@
...
-        let ptr = (unsafe { crate::esp32::esp_idf_svc::sys::esp_camera_fb_get() })
-            as *mut crate::esp32::esp_idf_svc::sys::camera_fb_t;
-        if ptr.is_null() {
-            None
-        } else {
-            Some(ptr)
+}
+
+impl Camera for Esp32Camera {
+    fn get_image(&mut self, mut buffer: BytesMut) -> Result<BytesMut, CameraError> {
```
acmorrow, line 164: How does the caller know how big a buffer to provide to avoid the error or be able to succeed on the next attempt?

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,254 @@
...
+        let frame = self.get_frame()?;
+        if frame.len() > buffer.capacity() {
+            return Err(CameraError::ImageTooBig);
         }
+        let msg = camera::v1::GetImageResponse {
+            mime_type: "image/jpeg".to_string(),
+            image: frame.as_bytes(),
+        };
+        // safety: message must be encoded before the frame is dropped from scope
+        msg.encode(&mut buffer).unwrap();
```
acmorrow, line 174: Can we avoid the `unwrap`?

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,254 @@
...
                 grab_mode: 0,
                 fb_location: 0,
-                sccb_i2c_port: 0,
+                sccb_i2c_port,
             },
-            last_grab: t.now(),
-        }
+        };
+
+        esp!(unsafe { esp_camera_init(&cam.config) })
```
acmorrow, line 153: - This reads to me like there can only ever be one camera. If that is true, there probably should be some guardrails around trying to instantiate more than one. Or, have you tested this and esp32_camera does a good job handling that situation?
- The docs for `esp32_camera_init` state 'note call camera_probe before calling this function'. Is that being done somewhere? It also seems a little weird, because it looks like `esp32_camera_init` also calls that?

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,254 @@
...
-    pub fn return_cam_frame(
-        &self,
-        frame: Option<*mut crate::esp32::esp_idf_svc::sys::camera_fb_t>,
-    ) {
-        if let Some(ptr) = frame {
-            unsafe { crate::esp32::esp_idf_svc::sys::esp_camera_fb_return(ptr) }
+
+    fn render_frame(&mut self, mut buffer: BytesMut) -> Result<BytesMut, CameraError> {
+        let frame = self.get_frame()?;
+        if frame.len() > buffer.capacity() {
```
acmorrow, line 181: Ditto re knowing how big to make the buffer.

templates/project/Cargo.toml
```diff
@@ -20,8 +20,15 @@ embedded-hal = { version = "0.2.7", features = ["unproven"]}
 log = "0.4"
 async-channel = "2"
 futures-lite = "1"
-micro-rdk = {version = "0.2.1", git = "https://github.com/viamrobotics/micro-rdk.git", features = ["esp32", "binstart","provisioning"], rev = "deb0687" }
+micro-rdk = {version = "0.2.1", git = "https://github.com/viamrobotics/micro-rdk.git", features = ["esp32", "binstart", "provisioning"{% if camera == "true" %}, "camera" {%endif%}], rev = "deb0687" }
```
acmorrow, line 23: This is getting a little unwieldy as a one-liner. Are we permitted to make it multi-line?:
```
micro-rdk = {
    version = ...,
    git = ...,
    features = [
        "esp32",
        ...
        {% if camera ... }.
    ],
}

File: micro-rdk/src/esp32/camera.rs [hunk 29]
```diff
@@ -1,117 +1,254 @@
...
+            return Err(CameraError::ImageTooBig);
         }
+        let msg = HttpBody {
+            content_type: "image/jpeg".to_string(),
+            data: frame.as_bytes().to_vec(),
+            ..Default::default()
+        };
+        // safety: message must be encoded before the frame is dropped from scope
+        msg.encode(&mut buffer).unwrap();
```
acmorrow, line 190: It isn't a problem with your code, but it does feel like the camera data gets copied a lot here trying to get it into the proto message. I'm not sure you can do better though. Same for the `get_image` call above.

templates/project/cargo-generate.toml
```diff
@@ -8,3 +8,11 @@ type = "string"
 prompt = "MCU"
 choices = ["esp32"]
 default = "esp32" 
+
+[placeholders.camera]
+type = "bool"
+prompt = "include camera?"
```
acmorrow, line 14: Can we make this just prompt just slightly less terse?

micro-rdk/src/common/camera/mod.rs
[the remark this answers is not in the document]
mattjperez: Lazily and cache or generating on each call?
Would it be alright to file this under a future enhancement ticket?

micro-rdk/src/esp32/camera.rs
mattjperez: I have not tested having multiple cameras.
There are a couple ways you could handle this, but I think deserves its own ticket.

1. since there's a limited number of pins and a single camera needs almost half of them, tracking which pins are currently allocated on the device and throwing an error when trying to assign them
2. setting a global to track if a camera has already been initialized, throwing an error if so.

On your second point, I see the comment in `esp32-camera` that refers to this, but as you said, the probe is called in the function. Within the examples, the probe function isn't called either.

micro-rdk/src/esp32/camera.rs
mattjperez: This is a good point. The actual underlying buffer has its size hard-coded. The user is indirectly "requesting" a specific buffer size depending on the resolution they choose. So, a camera that has been set up with `FrameSize:: W240XH240` will work with the current hardcoding, but one initialized with `FrameSize::VGA` would not fit within the buffer, and we would know this before the first `get_image` request.

I think the best option for now is to error on the camera creation if the `FrameSize` exceeds the currently configured buffer. This will let them choose another `FrameSize` in app that will work.

micro-rdk/src/esp32/camera.rs
mattjperez: I know what you mean. I need to run more tests later to determine how many times data is being (re)written and everything. Even the `to_vec()` in L186 feels suspect.  

Ideally, the message can be encoded in-place, where it has already been written to by the camera so we can send it directly.

templates/project/cargo-generate.toml
mattjperez: `include camera module`? fine with any suggestions

micro-rdk/src/esp32/camera.rs
mattjperez: This would probably make the capacity check unnecessary as well

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,254 @@
...
+            .unwrap_or(20_000_000);
+        let ledc_timer = cfg.get_attribute::<u32>("ledc_timer").unwrap_or(1);
+        let ledc_channel = cfg.get_attribute::<u32>("ledc_channel").unwrap_or(1);
+        let frame_size = cfg
+            .get_attribute::<u32>("frame_size")
+            .unwrap_or(FrameSize::QVGA as u32);
+        // Quality of JPEG output: 0-63 lower means higher quality
+        let jpeg_quality = cfg.get_attribute::<i32>("jpeg_quality").unwrap_or(32);
+        //  If pin_sccb_sda is -1, use the already configured I2C bus by number
+        let sccb_i2c_port = cfg.get_attribute::<i32>("sccb_i2c_port").unwrap_or(0);
```
gvaradarajan, line 116: does this require that you've configured the i2c bus on board? If so, I think I'd actually like to see some validation in this regard (maybe by requiring an i2c bus name in the config and failing if we can't fetch the handle by name).

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,254 @@
...
 
 impl Esp32Camera {
-    pub fn new() -> Self {
-        let t = EspSystemTime;
-        Esp32Camera {
+    pub(crate) fn from_config(
+        cfg: ConfigType,
+        _: Vec<Dependency>,
+    ) -> Result<CameraType, CameraError> {
+        let pin_pwdn = cfg.get_attribute::<i32>("pin_pwdn").unwrap_or(-1);
```
gvaradarajan, line 89: could you include a couple comments on the reasoning behind these defaults?

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,254 @@
...
+            .map_err(|e| CameraError::InitError(Box::new(e)))?;
+        Ok(Arc::new(Mutex::new(cam)))
     }
-    pub fn setup(&self) -> Result<(), CameraError> {
-        let ret = (unsafe { crate::esp32::esp_idf_svc::sys::esp_camera_init(&self.config) })
-            as crate::esp32::esp_idf_svc::sys::esp_err_t;
-        let ret = crate::esp32::esp_idf_svc::sys::EspError::convert(ret);
-        ret.map_err(|e| CameraError::CameraInitError(e.into()))
+
+    fn get_frame(&mut self) -> Result<Esp32CameraFrameBuffer, CameraError> {
```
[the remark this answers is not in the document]
gvaradarajan: yeah, this doesn't feel like it needs to be its own function

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,254 @@
...
-            unsafe { crate::esp32::esp_idf_svc::sys::esp_camera_fb_return(ptr) }
+
+    fn render_frame(&mut self, mut buffer: BytesMut) -> Result<BytesMut, CameraError> {
+        let frame = self.get_frame()?;
+        if frame.len() > buffer.capacity() {
+            return Err(CameraError::ImageTooBig);
         }
+        let msg = HttpBody {
+            content_type: "image/jpeg".to_string(),
+            data: frame.as_bytes().to_vec(),
```
gvaradarajan, line 186: does this `to_vec` call make another copy? Is it worth having a separate `as_vec` function on `Esp32CameraFrameBuffer` that calls `Vec::from_raw_parts` directly instead of going through `Bytes`?

micro-rdk/src/esp32/camera.rs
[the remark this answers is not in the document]
mattjperez: I had this to
1. have self so can be used like `cam.get_frame()` instead of `Esp32Cam::get_frame`
2. to just enforce mutability semantics. While there's nothing in the `Esp32Camera` struct that's being mutated, in reality the underlying frame buffer used by the camera is. 

want me to just make it a `self` method instead?

micro-rdk/src/esp32/camera.rs
gvaradarajan: Other than the `to_vec` call, I don't understand where the additional copying is coming from. The `as_bytes` function consumes the frame and the `Bytes` struct is really just an iterator. Is it in the call to `encode`?

gvaradarajan (approved):
LGTM, pending the other code review. Only some minor concerns

micro-rdk/src/esp32/camera.rs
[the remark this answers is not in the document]
mattjperez: @gvaradarajan you want me to add this to the `get_image` and `render_frame` methods instead?

```rs
    impl Camera for Esp32Camera {
        fn get_image(&mut self, mut buffer: BytesMut) {
        let frame =  Esp32CameraFrameBuffer::get().ok_or(CameraError::FailedToGetImage)?;
        ...
    }
}
```

Feels a bit weird if there's nothing to tie this frame buffer to `self` and not use `self` anywhere in the trait function.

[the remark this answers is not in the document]
mattjperez: @acmorrow turns out it needs to be `&mut self` to use `self.get_frame()` properly otherwise I'm getting deref errors.

you prefer the `&mut self` with `self.get_frame()` or just embedding the `Esp32FrameBuffer` creation directly in the `get_image` and `render_frame` calls?

```rs
cannot move out of `*self` which is behind a mutable reference
   --> micro-rdk/src/esp32/camera.rs:181:21
    |
181 |         let frame = self.get_frame()?;
    |                     ^^^^ ----------- `*self` moved due to this method call                                         
    |                     |
    |                     move occurs because `*self` has type `Esp32Camera`, which does not implement the `Copy` trait
```

micro-rdk/src/esp32/camera.rs
mattjperez: Discussed offline with @gvaradarajan 
it appears that `esp_camera_init` accepts two options for i2c, either:
a) setting `pin_sccb_scl` and `pin_sccb_sda`, which it then initializes
b) choosing an already configured `sccb_i2c_port`, which it then attaches to.

when using the [m5 Camera Timer X](https://docs.m5stack.com/en/unit/timercam_x)'s ov3660 camera and values in `pin_sccb_sda`/`pin_sccb_scl`, I needed to set `sccb_i2c_port` to `-1` to prevent a low-level panic during `esp_cam_init`. 

I haven't confirmed what the behavior is if we use an i2c bus that we set in app and use those values here. When there are `i2c` pins set on the board, we unsafely initialize the pins with `unsafe AnyIOPin::new(pin)`. 

Follow up ticket: [RSDK-8327 ](https://viam.atlassian.net/browse/RSDK-8327)

micro-rdk/src/esp32/camera.rs
acmorrow: Let's talk this one over in person when you have a minute. I may be misunderstanding something about how this API is supposed to work, and I don't want to waste your time on it.

micro-rdk/src/esp32/camera.rs
acmorrow: Yes, the call to encode.

micro-rdk/src/esp32/camera.rs
mattjperez: I added the `Vec::from_raw_parts` but I'm concerned about the safety of it.

As it's now a `Vec`, rust will try to deallocate the data on drop, [but there doesn't appear to be deallocation, just setting bitflags](https://github.com/espressif/esp32-camera/blob/7aa37d4f22503fdac9ccd449e4678c4894c40055/driver/cam_hal.c#L522).

micro-rdk/src/esp32/camera.rs
mattjperez: I'm not sure how we can prevent this though and form the message buffer in-place as the actual buffer containing the bytes is managed by `esp-cam` via DMA. When we encode, the bytes are then copied over from the camera's frame buffer into our webrtc/grpc buffer.

micro-rdk/src/common/camera/mod.rs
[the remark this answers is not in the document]
npmenard: if it's only get linked when camera feature is enabled then i am ok with this.

micro-rdk/src/esp32/camera.rs
acmorrow: OK, thanks for looking into it.

micro-rdk/src/esp32/camera.rs
mattjperez: discussed offline with @acmorrow. The current grpc/webrtc buffer is statically sized, limiting the usable `frame_size`s and requiring these checks.

Created a follow up ticket: [RSDK-8830](https://viam.atlassian.net/browse/RSDK-8330?atlOrigin=eyJpIjoiOTU1NzM5N2ZhZDdjNDg3NTlhM2JmZTVlNWZiMTRlOGQiLCJwIjoiaiJ9)

acmorrow (approved):
LGTM mod some small fixes around the locking.

mattjperez:
confirmed now only one esp32camera can be instantiated at any given moment. The following is from having a duplicate component entry in app.

```sh
E (11158) micro_rdk::common::robot: Failed to build resource `esp32camera-copy` of type `camera`: RobotResourceBuildError(InitError("only one camera allowed per machine"))
E (11164) micro_rdk::common::robot: Failed to build resource `esp32camera-copy` of type `camera`: RobotResourceBuildError(InitError("only one camera allowed per machine"))
E (11180) micro_rdk::common::robot: These components couldn't be built ["esp32camera-copy"]. Check for errors, missing or circular dependencies in the config.
```

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,269 @@
...
-                }
-            }
-        }
+}
+
+impl Drop for Esp32Camera {
+    fn drop(&mut self) {
+        unsafe { esp_camera_deinit() };
+        let mut is_registered = CAMERA_ALREADY_REGISTERED.lock().unwrap();
+        *is_registered = true;
```
acmorrow, line 211: I think this should be `false`

mattjperez: yup, just found it. doing testing before merging stuff. 
Also realized that the failed initialization of camera actually does a drop, which calls `esp_camera_deinit`, so fixing that now. will confirm proper behavior before merging.

I'm doing things like adding, removing, testing multiple cameras on app.

acmorrow: OK. If you want me to review that new logic just re-request review and I'll take a look.

mattjperez:
@acmorrow confirmed behavior on app having multiple cameras configured. only one will be successfully built and the `test`s on each config RC card have expected responses/values.

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,278 @@
...
+        let ledc_channel = cfg.get_attribute::<u32>("ledc_channel").unwrap_or(1);
+        let frame_size = cfg
+            .get_attribute::<u32>("frame_size")
+            .unwrap_or(FrameSize::QVGA as u32);
+        // Quality of JPEG output: 0-63 lower means higher quality
+        let jpeg_quality = cfg.get_attribute::<i32>("jpeg_quality").unwrap_or(32);
+        //  If pin_sccb_sda is -1, use the already configured I2C bus by number
+        let sccb_i2c_port = cfg.get_attribute::<i32>("sccb_i2c_port").unwrap_or(-1);
+
+        let mut cam = Self {
```
acmorrow, line 121: I think if you move creating the actual Camera object to _after_ you have a successful call `esp_camera_init`, then you won't need the `active` field because there won't be a `Camera` object to `Drop`.

micro-rdk/src/esp32/camera.rs
```diff
@@ -1,117 +1,278 @@
...
+    CIF = 6,
+    /// 480x320, ~31.36KB JPEG
+    HVGA = 7,
+    /// 640x480, ~62.73KB JPEG
+    VGA = 8,
+}
+
+#[derive(DoCommand)]
 pub struct Esp32Camera {
+    active: bool,
```
acmorrow, line 83: I don't actually see that `config` is ever used after `from_config` returns. Maybe this `struct` can just be empty?

micro-rdk/src/esp32/camera.rs
mattjperez: tested your suggestion, works!

micro-rdk/src/esp32/camera.rs
mattjperez: we will probably want to use it later for setting up the other camera methods, like getting the configured resolution and stuff. I haven't decided how I want to deal with it yet, will decide on it when implementing the other trait methods.

PR #261, opened 2024-07-11
[RSDK-7822] follow triplet naming convention for model

gvaradarajan:


mattjperez (commented):
Could we use crate-level constants for the namespaces? One for micro rdk another for examples 

Something like 
```rs
// Namespace "rdk:builtin:"
crate::NAMESPACE + "fake"
```

micro-rdk/src/common/registry.rs
```diff
@@ -133,6 +137,27 @@ type GenericComponentConstructor =
...
+///     Ok(())
+/// }
+/// ```
+#[macro_export]
+macro_rules! model {
+    ($org:tt, $mname:tt) => {
+        concat!($org, ":", env!("CARGO_PKG_NAME"), ":", $mname)
+    };
+}
+
```
mattjperez, line 160: can we move this to the `micro-rdk-macros` crate?

mattjperez (approved):
LGTM, just a couple nits

micro-rdk/src/common/analog.rs
```diff
@@ -47,7 +47,7 @@ pub struct AnalogResolution {
     pub step_size: f32,
 }
 
-pub trait AnalogReader<Word> {
```
npmenard, line 50: why did we end up needing to add Send bound?

micro-rdk/src/common/board.rs
```diff
@@ -45,15 +45,15 @@ pub static COMPONENT_NAME: &str = "board";
...
+        .register_board("rdk:builtin:fake", &FakeBoard::from_config)
         .is_err()
     {
         log::error!("model fake is already registered")
     }
 }
 
 /// Represents the functionality of a general purpose compute board that contains various components such as analog readers and digital interrupts.
-pub trait Board: Status + DoCommand {
+pub trait Board: Status + DoCommand + Send {
```
npmenard, line 56: same question

micro-rdk/src/common/registry.rs
```diff
@@ -24,6 +24,8 @@ use crate::proto::common::v1::ResourceName;
 pub enum RegistryError {
     #[error("RegistryError : Model '{0}' not found")]
     ModelNotFound(String),
```
npmenard, line 26: do we still use this error then?

micro-rdk/src/common/registry.rs
```diff
@@ -24,6 +24,8 @@ use crate::proto::common::v1::ResourceName;
...
+    #[error("RegistryError : Component Type '{0}' not found")]
+    ComponentTypeNotFound(String),
```
npmenard, line 28: what would be the difference between ModelNotFound and ComponentTypeNotFound

npmenard (approved):
done with pass I left a couple of comments but otherwise LGTM.

If I understand correctly we now compare the FQDN of a component to lookup whether they exist or not in the registry?
Lastly for the config of the modular component what is the FQDN, rdk::viam::wfi_sensor? What does there JSON config look like then? do we need to update the config example in the README.md? Is it going  to brick current system using them if they update?

micro-rdk/src/common/registry.rs
gvaradarajan: I'd rather not The only reason that crate exists is because proc-macros currently need to be in their own crate

micro-rdk/src/common/analog.rs
gvaradarajan: The modules fail clippy without it. The way we use components (with `Arc<Mutex<Component>>`) is supposed to require that `MoistureSensor` implement `Send`, which stores an `Arc<Mutex<Box<dyn AnalogReader>>`, which in turn requires that anything that implements `AnalogReader` also implement `Send`

micro-rdk/src/common/board.rs
gvaradarajan: similar answer to above

micro-rdk/src/common/registry.rs
gvaradarajan: Component type => "board", "motor", "sensor", etc.
Model type => "rdk:builtin:gpio_motor", "rdk:builtin:esp32", etc.

micro-rdk/src/common/registry.rs
gvaradarajan: I believe so, in `get_board_constructor` etc.

gvaradarajan:
> done with pass I left a couple of comments but otherwise LGTM.
> 
> If I understand correctly we now compare the FQDN of a component to lookup whether they exist or not in the registry? Lastly for the config of the modular component what is the FQDN, rdk::viam::wfi_sensor? What does there JSON config look like then? do we need to update the config example in the README.md? Is it going to brick current system using them if they update?

@npmenard I've edited the PR decsiption to clarify some of this. This will be a breaking change for anyone who updates (in that the model value in their component configs will need to match whatever name they passed to `register_<component>`). I will update the config example in README.md

npmenard:
how is it going to affect module constructed via the C-FFI?

gvaradarajan:
> how is it going to affect module constructed via the C-FFI?

@npmenard at the moment, modules that don't use the macro as well as the C-FFI sensor should be unaffected

PR #262, opened 2024-07-15
[RSDK-8187] config cache changes

allisonschiang:
Added storage into the config_monitor in order to write newly detected configs to NVS before rebooting, preventing a reboot cycle where the config is out of date.

However, this introduced a new issue where some invalid configs would cause a reboot loop, as it would crash the robot before entry.rs could detect a robot error and reset_robot_configuration(). 

Some fixes could be:
- Keep this approach and try to catch configs that would cause a reboot loop (don't know if config is valid until bot crashes)
- Keep this approach and individually return errors to catch crashing invalid configs as we discover them
- Add another check that pulls config from app on boot-up and replaces the cache with any new one so if the user changes to a valid config it will overwrite the invalid config in cache


micro-rdk/src/common/config_monitor.rs
```diff
@@ -1,19 +1,39 @@
...
-impl<'a> ConfigMonitor<'a> {
-    pub fn new(restart_hook: impl FnOnce() + 'a, curr_config: ConfigResponse) -> Self {
+impl<'a, S> ConfigMonitor<'a, S>
+where
+    S: RobotConfigurationStorage + WifiCredentialStorage + Clone + 'static,
+    <S as RobotConfigurationStorage>::Error: Debug,
+    ServerError: From<<S as RobotConfigurationStorage>::Error>,
+    <S as WifiCredentialStorage>::Error: Sync + Send + 'static,
+{
+    pub fn new(restart_hook: impl FnOnce() + 'a, curr_config: ConfigResponse, storage: S) -> Self {
```
acmorrow, line 32: Let's make the lambda come last.

micro-rdk/src/common/config_monitor.rs
```diff
@@ -43,7 +69,15 @@ impl<'a> PeriodicAppClientTask for ConfigMonitor<'a> {
                 app_client.clone().get_config(None).await.unwrap();
             match self.curr_config == *new_config {
                 true => Ok(Some(self.get_default_period())),
-                false => self.restart(),
+                false => {
+                    if let Err(e) = self
+                        .storage
+                        .store_robot_configuration(new_config.config.unwrap())
```
acmorrow, line 75: I'm suspicious of `unwrap`. What should happen here if `new_config.config` is None?

acmorrow (changes requested):

micro-rdk/src/common/config_monitor.rs
allisonschiang: not sure, what happens in entry.rs when get_config is called and new_config is None or returns an AppClientError?

https://github.com/viamrobotics/micro-rdk/blob/f95d10b20e2970c1f32b203ff4f7e1cb4932521e/micro-rdk/src/esp32/entry.rs#L83

allisonschiang: I'm guessing it panics?

allisonschiang: oh I see, I already new_config is already unwrapped

micro-rdk/src/common/config_monitor.rs
```diff
@@ -1,19 +1,39 @@
...
 use std::time::Duration;
 
-pub struct ConfigMonitor<'a> {
-    restart_hook: Option<Box<dyn FnOnce() + 'a>>,
+pub struct ConfigMonitor<'a, S>
+where
+    S: RobotConfigurationStorage + WifiCredentialStorage + Clone + 'static,
+    <S as RobotConfigurationStorage>::Error: Debug,
+    ServerError: From<<S as RobotConfigurationStorage>::Error>,
+    <S as WifiCredentialStorage>::Error: Sync + Send + 'static,
```
acmorrow, line 18: This won't work unless `feature = "provisioning"` is enabled, but see #267 which offers a way out. Perhaps this change should wait on that one.

micro-rdk/src/common/config_monitor.rs
```diff
@@ -1,19 +1,39 @@
 use super::app_client::{AppClient, AppClientError, PeriodicAppClientTask};
+#[cfg(feature = "provisioning")]
+use crate::common::{
+    grpc::ServerError,
+    provisioning::storage::{RobotConfigurationStorage, WifiCredentialStorage},
+};
 use crate::proto::app::v1::ConfigResponse;
 use futures_lite::Future;
+use std::fmt::Debug;
```
acmorrow, line 9: Where is this getting used?

acmorrow (changes requested):

acmorrow:
> Added storage into the config_monitor in order to write newly detected configs to NVS before rebooting, preventing a reboot cycle where the config is out of date.
> 
> However, this introduced a new issue where some invalid configs would cause a reboot loop, as it would crash the robot before entry.rs could detect a robot error and reset_robot_configuration().
> 
> Some fixes could be:
> 
> * Keep this approach and try to catch configs that would cause a reboot loop (don't know if config is valid until bot crashes)
> * Keep this approach and individually return errors to catch crashing invalid configs as we discover them
> * Add another check that pulls config from app on boot-up and replaces the cache with any new one so if the user changes to a valid config it will overwrite the invalid config in cache

What about the following idea: when building from a cached config, we first fetch the config, and then we immediately erase it. Then we build the robot. Only if the robot builds OK do we write it back to the cache again. I don't love it because of the extra flash wear it implies, but that's already a problem we have.

acmorrow:
> > Added storage into the config_monitor in order to write newly detected configs to NVS before rebooting, preventing a reboot cycle where the config is out of date.
> > However, this introduced a new issue where some invalid configs would cause a reboot loop, as it would crash the robot before entry.rs could detect a robot error and reset_robot_configuration().
> > Some fixes could be:
> > 
> > * Keep this approach and try to catch configs that would cause a reboot loop (don't know if config is valid until bot crashes)
> > * Keep this approach and individually return errors to catch crashing invalid configs as we discover them
> > * Add another check that pulls config from app on boot-up and replaces the cache with any new one so if the user changes to a valid config it will overwrite the invalid config in cache
> 
> What about the following idea: when building from a cached config, we first fetch the config, and then we immediately erase it. Then we build the robot. Only if the robot builds OK do we write it back to the cache again. I don't love it because of the extra flash wear it implies, but that's already a problem we have.

I discussed this idea with @npmenard and he feels like the flash wear is too much of a problem. Please move forward with committing this (once you have addressed the failure to build without provisioning), and file a new bug ticket about the situation that arises when a panic inducing config is cached. Then we can tackle that in a new PR.

micro-rdk/src/common/config_monitor.rs
allisonschiang: I'm using it to copy the storage type when passing it in from entry.rs:

 S: RobotConfigurationStorage + WifiCredentialStorage + Clone + 'static,
    <S as RobotConfigurationStorage>::Error: Debug,
    ServerError: From<<S as RobotConfigurationStorage>::Error>,
    <S as WifiCredentialStorage>::Error: Sync + Send + 'static,
    
    Should I remove it from the type?

acmorrow: No, its fine I just didn't see it first time around.

micro-rdk/src/common/config_monitor.rs
```diff
@@ -43,7 +69,12 @@ impl<'a> PeriodicAppClientTask for ConfigMonitor<'a> {
                 app_client.clone().get_config(None).await
             {
                 if self.curr_config != *new_config {
-                    self.restart()
+                    if let Some(config) = new_config.config {
+                        if let Err(e) = self.storage.store_robot_configuration(config) {
+                            log::warn!("Failed to store new robot configuration from app: {}", e);
+                        }
+                    }
+                    self.restart();
```
mattjperez, line 77: please add the crash loop bug as a TODO comment here once it's made.

mattjperez (commented):
This looks good for me (nice work on the generics in signatures), but definitely 
1. wait for new #267 to merge and retest compilation with and without provisioning
2. make and reference a follow up ticket in TODO comment

micro-rdk/src/common/config_monitor.rs
```diff
@@ -43,7 +69,12 @@ impl<'a> PeriodicAppClientTask for ConfigMonitor<'a> {
                 app_client.clone().get_config(None).await
             {
                 if self.curr_config != *new_config {
-                    self.restart()
+                    if let Some(config) = new_config.config {
+                        if let Err(e) = self.storage.store_robot_configuration(config) {
```
npmenard, line 73: did we consider erasing  the config then restarting?

acmorrow: We could. That would leave open the possibility that if you can't connect when you come back online, you won't be able to build a robot until you talk to app again. That might be a risk worth taking.

allisonschiang: I think this could work, its either a empty config or an invalid config if you can't come back online which both wouldn't work. Should I implement this solution?

npmenard: let's make the assumption that network will be available until the next time we poll config after a reboot. The big risk would be an invalid config but that's on the user

acmorrow: OK. @allisonschiang please take @npmenard's suggestion to just erase the config instead of storing it here.

micro-rdk/src/common/config_monitor.rs
allisonschiang: @gvaradarajan Do I need to do anything to make provisioning work or will it just work after your PR?

micro-rdk/src/common/config_monitor.rs
```diff
@@ -43,7 +69,13 @@ impl<'a> PeriodicAppClientTask for ConfigMonitor<'a> {
...
             {
                 if self.curr_config != *new_config {
-                    self.restart()
+                    if let Err(e) = self.storage.reset_robot_configuration() {
+                        log::warn!(
+                            "Failed to reset robot config after new config detected: {}",
+                            e
+                        );
+                    }
+                    self.restart();
```
allisonschiang, line 78: Does it make more sense to restart if the config cache wasn't reset or to not reset? I'm leaning towards reset because it could fix the reason why the cache failed to erase.

acmorrow: I think I agree. If you fail to clear the cache, if you restart you will run the stale config, and if you don't restart you will continue to run the stale config.

acmorrow: But if NVS writes aren't working, your best hope is that a reboot will help.

acmorrow: Meanwhile, we spent a lot of time talking about why this clears the cache rather than writing the configuration to it. I think a fairly detailed comment is warranted explaining why resetting the cache and rebooting is our current best strategy.

micro-rdk/src/common/config_monitor.rs
acmorrow: @allisonschiang - Please follow up on this comment once you have made adjustments to address the issue.

micro-rdk/src/common/config_monitor.rs
allisonschiang: So we prefer a (possible reboot loop on an NVS write fail) over a (basically permanently stale config cache assuming NVS doesn't fix itself)?

allisonschiang: I just realized when writing my comment the alternative, which is nvs fail -> reboot -> nvs fail -> reboot ... and the config cache never changing. While rebooting may fix the issue, maybe a stale config is preferred over a reboot loop?

micro-rdk/src/common/config_monitor.rs
```diff
@@ -1,19 +1,39 @@
 use super::app_client::{AppClient, AppClientError, PeriodicAppClientTask};
+#[cfg(feature = "provisioning")]
```
acmorrow, line 2: I think you should remove this now, since the types are no longer gated on provisioning.

micro-rdk/src/common/config_monitor.rs
acmorrow: That's a good point. I think I agree then: if you can't clear NVS you are better off continuing with your existing config than entering a boot loop driven by continually detecting a new config. I also expect that being unable to write to NVS probably represents a badly degraded device or runtime, and our options are limited there. It is a fussy edge case though, and I'd be interested to hear if @npmenard has thoughts on the best outcome here.

acmorrow (approved):
LGTM!

PR #263, opened 2024-07-15
[RSDK-7692] Shutdown command

allisonschiang:
exit() and esp_restart() both return nothing, so I assume they cannot fail and did not return an error

### DIFF
--- a/micro-rdk/src/common/grpc.rs
+++ b/micro-rdk/src/common/grpc.rs
@@ -243,6 +243,7 @@ where
             "/viam.robot.v1.RobotService/ResourceNames" => self.resource_names(payload),
             "/viam.robot.v1.RobotService/GetStatus" => self.robot_status(payload),
             "/viam.robot.v1.RobotService/GetOperations" => self.robot_get_oprations(payload),
+            "/viam.robot.v1.RobotService/Shutdown" => self.robot_shutdown(payload),
             "/proto.rpc.v1.AuthService/Authenticate" => self.auth_service_authentificate(payload),
             "/viam.component.sensor.v1.SensorService/GetReadings" => {
                 self.sensor_get_readings(payload)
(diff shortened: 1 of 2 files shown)


micro-rdk/src/common/grpc.rs
```diff
@@ -1212,6 +1213,16 @@ where
         self.encode_message(operation)
     }
 
+    // robot_shutdown will not return anything because will restart
+    fn robot_shutdown(&mut self, _: &[u8]) -> ! {
```
acmorrow, line 1217: You could just not take an argument here too. Nothing requires you to pass `payload` above, that I see.

micro-rdk/src/common/grpc.rs
```diff
@@ -1212,6 +1213,16 @@ where
...
+        #[cfg(feature = "native")]
+        std::process::exit(0);
+        #[cfg(feature = "esp32")]
+        unsafe {
+            crate::esp32::esp_idf_svc::sys::esp_restart();
```
acmorrow, line 1222: Is it OK that `shutdown` for an esp32 is just going to reboot and restart the robot? What are the intended semantics of `shutdown`? Should this require some more active measures to get the robot running again, like powercycling the esp32? We could, for instance, enter deep sleep with no wake source instead.

acmorrow (approved):
The code LGTM with one small optional improvement. I'd like some clarification though on whether just restarting is an OK outcome on esp32 as opposed to taking some more drastic step. @stevebriskin and @npmenard I'd appreciate your thoughts. If `shutdown` is supposed to be estop-ish, what should happen here?

micro-rdk/src/common/grpc.rs
allisonschiang: I see some other requests that don't take any arguments still take a payload and don't use the value, like get_operations(), Resource_names(), etc... Is shutdown a different situation?

acmorrow: I wonder if you get an unused warning above or something if you do that. OK, nevermind, just follow the pattern that is in use and go with what you have here now.

mattjperez (approved):
LGTM

PR #266, opened 2024-07-16
Simpler esp idf setup

acmorrow:
To land this change (assuming that the decision is in favor of it), we will need to merge the first commit first, then make a new canon image release, and only then can the second commit merge.

The practical effect of this change is that within the canon environment, the shell injected ESP-IDF environment (i.e. the effect of sourcing $IDF_PATH/export.sh`) is used, because `ESP_IDF_TOOLS_INSTALL_DIR=fromenv` is set in the canon shell environment.

In other environments where `ESP_IDF_TOOLS_INSTALL_DIR=fromenv` is *not* set, then the default value for the ESP-IDF version comes from the value set in `Cargo.toml` via the `[package.metadata.esp-idf-sys].esp_idf_version` key, and `cargo` and `embuild` will collaborate to fetch, cache, and enable the required ESP-IDF version. Developers will no longer need to manually inject the ESP-IDF environment by running `export.sh` in the shell, nor will they need to ensure that those values are propagated into or established within their editor or IDE in order to build for ESP-IDF.

This change also removes the `MCU` specification, since this value is automatically derived from the target triple in effect.

I'll leave a few notes in the files.


.cargo/config.toml
```diff
@@ -1,7 +1,7 @@
 [target.xtensa-esp32-espidf]
 linker = "ldproxy"
 runner = "espflash --monitor"
-rustflags = ["-C", "default-linker-libraries", "-Clink-args=-Wl,-Map=./esp-build.map"] 
+rustflags = ["-C", "default-linker-libraries", "-Clink-args=-Wl,-Map=./esp-build.map"]
```
acmorrow, line 4: Just whitespace my editor trimmed

etc/docker/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM rust:1.75-bookworm AS builder
 RUN apt update && apt-get install -y --no-install-recommends git libudev-dev make bzip2 gcc libgcrypt20 libgcrypt20-dev  ninja-build libudev-dev sudo wget git flex bison gperf ccache libffi-dev libssl-dev dfu-util libusb-1.0-0 cmake pkg-config libglib2.0-dev libpixman-1-dev software-properties-common openssh-client libslirp-dev python3-venv
 
 RUN export TARGET=$(rustc -vV | sed -En 's/^host: //p') && \
-    curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash && \ 
+    curl -L --proto '=https' --tlsv1.2 -sSf https://raw.githubusercontent.com/cargo-bins/cargo-binstall/main/install-from-binstall-release.sh | bash && \
```
acmorrow, line 5: Ditto whitespace

etc/docker/Dockerfile
```diff
@@ -99,6 +99,7 @@ RUN rm -rf $CARGO_HOME/registry
 RUN echo "PATH=$PATH" >> ~/.bash_profile && \
     echo ". $ESP_ROOT/export-esp.sh" >> ~/.bash_profile && \
     echo ". $IDF_PATH/export.sh"  >> ~/.bash_profile && \
+    echo "ESP_IDF_TOOLS_INSTALL_DIR=fromenv"  >> ~/.bash_profile && \
```
acmorrow, line 102: This is the change we need to get into the canon image before we can release the rest of the change. The effect of this is to override
```
[package.metadata.esp-idf-sys]
esp_idf_version = "v4.4.4"
```
In the relevant `Cargo.toml`

micro-rdk-ffi/build.rs
```diff
@@ -20,9 +20,6 @@ fn main() {
     #[cfg(not(feature = "test"))]
     {
         if env::var("TARGET").unwrap() == "xtensa-esp32-espidf" {
-            if std::env::var_os("IDF_PATH").is_none() {
```
acmorrow, line 23: A lack of `IDF_PATH` in the environment is no longer indicative of a misconfigured environment.

micro-rdk-server/.cargo/config.toml
```diff
@@ -1,7 +1,7 @@
 [target.xtensa-esp32-espidf]
 linker = "ldproxy"
 runner = "espflash --monitor"
-rustflags = ["-C", "default-linker-libraries", "-Clink-args=-Wl,-Map=./esp-build.map"] 
+rustflags = ["-C", "default-linker-libraries", "-Clink-args=-Wl,-Map=./esp-build.map"]
```
acmorrow, line 4: Whitespace

micro-rdk-server/.cargo/config.toml
```diff
@@ -18,9 +18,3 @@ rustflags = ["-C", "default-linker-libraries"]
 linker = "ldproxy"
 runner = "espflash --monitor"
 rustflags = ["-C", "default-linker-libraries"]
-
-
-[env]
```
acmorrow, line 23: These were the controlling values since `micro-rdk-server` is set as the default workspace member. However, I'm not convinced that `ESP_IDF_VERSION` had any effect since `ESP_IDF_TOOLS_INSTALL_DIR` was `fromenv`, and it would have just used whatever version was injected with `IDF_PATH/export.sh`.

The rule from https://github.com/esp-rs/esp-idf-sys/blob/master/BUILD-OPTIONS.md#esp-idf-configuration is that environment variable values (either in the shell or via an `[env]` section in `config.toml`) dominate. So the way we were doing it meant that they could not be overridden.

By moving to using the `Cargo.toml` section to set the defaults, we both let embuild/cargo/esp-idf worry about getting the right ESP-IDF version in place and we allow users to override the values on the command line. The canon shell takes advantage of that by using `ESP_IDF_TOOLS_INSTALL_DIR=fromenv` to honor the local shell environment, and developers can take advantage of it to opt into newer ESP-IDF versions trivially: `ESP_IDF_VERSION=4.4.8 make build-esp32-bin` will now use that version.

micro-rdk-server/.cargo/config.toml
```diff
@@ -18,9 +18,3 @@ rustflags = ["-C", "default-linker-libraries"]
...
-ESP_IDF_TOOLS_INSTALL_DIR = {value = "fromenv", force = true}
-ESP_IDF_VERSION= { value = "v4.4.1", force = true}
-MCU = { value= "esp32", force = true}
```
acmorrow, line 26: I've stopped setting `MCU` explicitly because the docs say it is derived from the target triple.

micro-rdk-server/.cargo/config.toml
```diff
@@ -18,9 +18,3 @@ rustflags = ["-C", "default-linker-libraries"]
 linker = "ldproxy"
 runner = "espflash --monitor"
 rustflags = ["-C", "default-linker-libraries"]
-
-
-[env]
-ESP_IDF_TOOLS_INSTALL_DIR = {value = "fromenv", force = true}
-ESP_IDF_VERSION= { value = "v4.4.1", force = true}
```
acmorrow, line 25: Note that `4.4.1` was inconsistent with what we have in the canon shell, which is 4.4.4

micro-rdk-server/Cargo.toml
```diff
@@ -50,8 +50,8 @@ path = "esp32/main.rs"
 
 
 [package.metadata.esp-idf-sys]
+esp_idf_version = "v4.4.4"
```
acmorrow, line 53: By setting the ESP-IDF version this way, we get the default of v4.4.4 (matches canon image) but it can be overridden via an environment variable.

micro-rdk-server/Cargo.toml
```diff
@@ -50,8 +50,8 @@ path = "esp32/main.rs"
...
 esp_idf_tools_install_dir = "global"
 esp_idf_sdkconfig_defaults = [
   "micro-rdk-server/sdkconfig.defaults",
-  "micro-rdk-server/sdkconfig.defaults.esp32"
```
acmorrow, line 56: It should not be necessary to have this file since the version associated with the current MCU should be auto selected based on the basename above.

micro-rdk/.cargo/config.toml
```diff
@@ -17,10 +17,3 @@ rustflags = ["-C", "default-linker-libraries"]
 linker = "ldproxy"
 runner = "espflash --monitor"
 rustflags = ["-C", "default-linker-libraries"]
-
-
-
-[env]
```
acmorrow, line 23: It was not clear to me that these had / have any effect. I think the ones in `micro-rdk-server` were the only ones that matter.

templates/project/.cargo/config.toml
```diff
@@ -30,8 +30,3 @@ rustflags = ["-C", "default-linker-libraries"]
 
 build-std = ["std", "panic_abort"]
 #build-std-features = ["panic_immediate_abort"] # Required for older ESP-IDF versions without a realpath implementation
-
-[env]
```
acmorrow, line 34: The template now uses the same mechanism.

templates/project/cargo-generate.toml
```diff
@@ -1,10 +1,10 @@
 [template]
 cargo_generate_version = ">=0.10.0"
 ignore = [".git","README.md"]
-description = "Micro-Rdk template"
+description = "Micro-RDK template"
```
acmorrow, line 4: Whitespace: at one point I had some other changes to this file but they got reverted, but I kept the ws cleanups

templates/project/.cargo/config.toml
```diff
@@ -30,8 +30,3 @@ rustflags = ["-C", "default-linker-libraries"]
...
-ESP_IDF_TOOLS_INSTALL_DIR = {value = "fromenv", force = true}
-ESP_IDF_VERSION= { value = "v4.4.1", force = true}
-MCU = { value= "esp32", force = true}
```
acmorrow, line 37: I think this was actually wrong: above we set the triple based on the MCU that was selected, but then we overrode it here. As in the regular micro-rdk, we should let the triple selection drive the MCU.

acmorrow:
Note that micro-rdk developers wishing to work against a specific ESP-IDF installation can still do so (after running the necessary installation script and sourcing the setup script as usual), but they will need to add `ESP_IDF_TOOLS_INSTALL_DIR=fromenv` to their shell environment in order to tell the micro-rdk to use it.

acmorrow:
> LGTM, this is going to relieve a lot of headache in a lot of situations. I think we should make a docs ticket for the resulting changes to the development setup instructions

Good call. I'll make an actual RSDK ticket for this, then we can mark it as docs changes needed when we merge it.

acmorrow:
https://viam.atlassian.net/browse/RSDK-8248

mattjperez (approved):
LGTM! It's great we can simplify the extra configuration around the different `config.toml`s

npmenard (commented):
Could be my machine but i noticed the following pattern:
For each built target either in release or debug (see table) will trigger a full git clone of esp-idf making each build much longer, i assume this extend to different feature combination. (i would expect esp-idf to be cloned exactly once) Also a side change of sdkconfig will also trigger a full git clone.  Cargo clean also resets the state. 
| command                                                                                              | try | simpler-esp-idf-setup                             | main                                              |
|------------------------------------------------------------------------------------------------------|-----|---------------------------------------------------|---------------------------------------------------|
| cargo +esp build -p micro-rdk-ffi --target=xtensa-esp32-espidf  --locked                             |   1 | 644.66s user 77.68s system 295% cpu 4:04.55 total | 671.52s user 76.13s system 779% cpu 1:35.92 total |
|                                                                                                      |   2 | 0.14s user 0.07s system 99% cpu 0.205 total       | 0.13s user 0.07s system 99% cpu 0.207 total       |
| cargo +esp build -p micro-rdk-server --bin micro-rdk-server-esp32                                    |   1 | 813.88s user 83.91s system 304% cpu 4:54.93 total | 833.33s user 81.95s system 639% cpu 2:23.04 total |
|                                                                                                      |   2 | 32.59s user 1.04s system 100% cpu 33.570 total    | 0.12s user 0.07s system 99% cpu 0.190 total       |
| cargo +esp build -p micro-rdk-server --bin micro-rdk-server-esp32 --release                          |   1 | 276.52s user 50.27s system 134% cpu 4:02.48 total | 271.54s user 45.69s system 401% cpu 1:19.09 total |
|                                                                                                      |   2 | 45.49s user 1.03s system 238% cpu 19.506 total    | 0.13s user 0.06s system 99% cpu 0.189 total       |
| make clippy-esp32                                                                                    |   1 | 232.80s user 56.25s system 126% cpu 3:48.87 total | 229.27s user 52.14s system 462% cpu 1:00.83 total |
| cargo +esp build -p micro-rdk-server --bin micro-rdk-server-esp32 --release (after sdkconfig change) |   1 | 224.36s user 46.13s system 122% cpu 3:40.03 total | 99.98s user 4.62s system 268% cpu 38.962 total    |

Overall I am not against such change (the issue I am running into could be just on my machine)  but from my quick experimentation there are lot of side effects that will make dev slower in some cases.
Do you have a similar experience on your machine?

mattjperez:
> Overall I am not against such change (the issue I am running into could be just on my machine) but from my quick experimentation there are lot of side effects that will make dev slower in some cases. Do you have a similar experience on your machine?

Testing on my machine work machine (ubuntu)

| command                                                                                              | try | simpler-esp-idf-setup                             | main                                              |
|------------------------------------------------------------------------------------------------------|-----|---------------------------------------------------|---------------------------------------------------|
| cargo +esp build -p micro-rdk-ffi --target=xtensa-esp32-espidf  --locked                             |   1 | todo | todo |
|                                                                                                      |   2 | todo       | todo      |
| cargo +esp build -p micro-rdk-server --bin micro-rdk-server-esp32                                    |   1 | todo | todo |
|                                                                                                      |   2 | todol    | todo  |
| cargo +esp build -p micro-rdk-server --bin micro-rdk-server-esp32 --release                          |   1 | todo | todo |
|                                                                                                      |   2 | todo    | todo       |
| make clippy-esp32                                                                                    |   1 | 226.19s user 16.65s system 97% cpu 4:08.07 total | todo|
| cargo +esp build -p micro-rdk-server --bin micro-rdk-server-esp32 --release (after sdkconfig change) |   1 | todo | todo    |

acmorrow:
> Overall I am not against such change (the issue I am running into could be just on my machine) but from my quick experimentation there are lot of side effects that will make dev slower in some cases. Do you have a similar experience on your machine?

Anecdotally, no, I don't see that slowdown, on at least the base case (just an ordinary rebuild). It looks like @mattjperez is working on reproducing your results; let's see what his numbers look like.

micro-rdk-server/Cargo.toml
```diff
@@ -50,8 +50,8 @@ path = "esp32/main.rs"
 
 
 [package.metadata.esp-idf-sys]
-esp_idf_tools_install_dir = "global"
+esp_idf_version = "v4.4.4"
+esp_idf_tools_install_dir = "custom:target/.embuild/espressif"
```
acmorrow, line 54: @viamrobotics/micro-rdk-team - Small update here: this will now make the esp-idf environment get installed under the `target` directory by default. You can override this on the command line, if you want, by setting `ESP_IDF_TOOLS_INSTALL_DIR=global`, or to any of the other valid values. I like this better because it now means that `cargo clean` will also destroy any cached esp-idf installations, making it easier to really start "clean". If esp-idf installations are going to `~/.espressif` by default it is too easy to be unaware that you are relying on local state.

acmorrow:
I also ran the build comparisons. I created a new google cloud machine and tested inside a micro-rdk canon docker image to isolate my test from my local setup. For the test results from `main` I used a `bash -l` environment so the ESP-IDF setup was performed, and for the `simpler-esp-idf-setup` results I used a non-login bash shell. My results seem to show no real difference in build performance across the identified tasks compared to main:

| command                                                                                              | try | simpler-esp-idf-setup                             | main                                              |
|------------------------------------------------------------------------------------------------------|-----|---------------------------------------------------|---------------------------------------------------|
| CARGO_HOME=target/cargo /usr/bin/time cargo +esp build  -p micro-rdk-ffi --target=xtensa-esp32-espidf --locked -Zbuild-std=std,panic_abort | 1 | 1028.21user 82.72system 2:12.22elapsed 840%CPU | 1032.36user 82.45system 2:06.21elapsed 883%CPU  |
|| 2 | 0.19user 0.06system 0:00.26elapsed 99%CPU | 0.20user 0.05system 0:00.25elapsed 99%CPU |
| CARGO_HOME=target/cargo /usr/bin/time cargo +esp build -p micro-rdk-server --bin micro-rdk-server-esp32 --release --target=xtensa-esp32-espidf --locked -Zbuild-std=std,panic_abort | 1 |  1036.86user 80.12system 3:20.74elapsed 556%CPU | 1049.29user 79.66system 3:14.31elapsed 581%CPU |
|| 2 | 0.19user 0.06system 0:00.25elapsed 100%CPU | 0.19user 0.06system 0:00.25elapsed 100%CPU ||
| CARGO_HOME=target/cargo /usr/bin/time cargo +esp build -p micro-rdk-server --bin micro-rdk-server-esp32 --release --target=xtensa-esp32-espidf --locked -Zbuild-std=std,panic_abort | 1 | 787.99user 72.94system 2:20.95elapsed 610%CPU | 802.12user 71.53system 2:16.75elapsed 638%CPU |
|| 2 | 0.19user 0.06system 0:00.26elapsed 100%CPU | 0.16user 0.08system 0:00.25elapsed 100%CPU |
| CARGO_HOME=target/cargo /usr/bin/time make clippy-esp32 | 1 | 609.07user 72.62system 1:52.62elapsed 605%CPU | 599.18user 73.42system 1:40.19elapsed 671%CPU |
|| 2 | 0.23user 0.10system 0:00.32elapsed 101%CPU | 0.23user 0.10system 0:00.36elapsed 95%CPU |
| CARGO_HOME=target/cargo /usr/bin/time cargo +esp build -p micro-rdk-server --bin micro-rdk-server-esp32 --release --target=xtensa-esp32-espidf --locked -Zbuild-std=std,panic_abort | 1 | 788.03user 73.34system 2:20.53elapsed 612%CPU | 799.44user 72.22system 2:15.00elapsed 645%CPU |
|| 2 | 0.19user 0.06system 0:00.26elapsed 100%CPU | 0.18user 0.06system 0:00.25elapsed 100%CPU |
| *sdkconfig.defaults | 3 | 304.77user 41.72system 1:31.86elapsed 377%CPU | 301.77user 43.33system 1:31.93elapsed 375%CPU |
|| 4 | 0.18user 0.06system 0:00.25elapsed 99%CPU | 0.19user 0.05system 0:00.25elapsed 99%CPU |

I think these results are pretty solid, and the results look very close. My suspicion is that the variation that was originally observed is local to @npmenard's setup. Since my expectation is that development on that system is likely to proceed via `ESP_IDF_TOOLS_INSTALL_DIR=fromenv` anyway, I think we are fairly safe moving forward with this and making the common case easier.

micro-rdk-ffi/Cargo.toml
```diff
@@ -40,10 +40,3 @@ cbindgen.workspace = true
 embuild.workspace = true
 serde.workspace = true
 serde_json.workspace = true
-
```
acmorrow, line 43: @npmenard - I removed this, because I don't think it is required: this crate should obtain these settings automatically from the root crate, which is `micro-rdk-server`.

acmorrow:
@npmenard - Any further thoughts on this?

acmorrow (commented):
@mattjperez and @gvaradarajan - Could you please take another look at this? I had to make some changes to get CI working. You can see the relevant new diff state here:

https://github.com/viamrobotics/micro-rdk/pull/266/files/3deb43069601db4767b6699ddf951cb9c548007c..db52cb3f69d955ced4e58225d533d71d7ffd02ef

Essentially, CI jobs were failing because my changes to our canon environment were applied via the `testbot` user's `.bash_profile`: https://github.com/viamrobotics/micro-rdk/commit/5b3535745ee3783465a5446199ee94c4ac750ca8#diff-10ee1de477ba6f448dabce8b9718de78161125911ed38a784c956e2ee1c40ae1R102

But these changes aren't in effect for GH workflow runs in the canon shell, since those (must) run as root per GH docs.

The existing CI infrastructure worked around this by simply manually running the setup tasks (e.g. `export-esp.sh`). I've opted here to instead have the CI tasks manually source `testbot`'s `.bash_profile`.

I don't love this solution, though it does work: the CI tasks are passing now. But if either of you have other suggestions on how to achieve this, I'm definitely interested.

.github/workflows/test.yml
```diff
@@ -135,8 +135,8 @@ jobs:
     - name: ESP32 Build & Clippy
       if : needs.changes.outputs.src-esp32 == 'true' || needs.changes.outputs.src-common == 'true' ||  needs.changes.outputs.src-servers == 'true'
       run: |
-        bash -c 'git config --global --add safe.directory /opt/esp/esp-idf && . "$IDF_PATH"/export.sh && . "$ESP_ROOT"/export-esp.sh && make clippy-esp32'
-        bash -c '. "$IDF_PATH"/export.sh && . "$ESP_ROOT"/export-esp.sh && cd micro-rdk-server && cd ../ && make build-esp32-bin'
+        bash -c '. /home/testbot/.bash_profile ; git config --global --add safe.directory "$ESP_ROOT"/esp-idf && make clippy-esp32'
+        bash -c '. /home/testbot/.bash_profile ; cd micro-rdk-server && cd ../ && make build-esp32-bin'
```
mattjperez, line 139: is the second and third command on this line necessary?

acmorrow: Yeah that's a little odd now that you point it out. I hadn't noticed it. As you said, it isn't changed here, but it does seem to amount to a no-op. I'll remove it if I revisit this file.

mattjperez:
@acmorrow LGTM (again)!

PR #268, opened 2024-07-17
connect to best AP

npmenard:
Change the connection behavior to do a full scan and selecting the AP with the strongest signal, instead of connecting to the first found AP which may not be the best AP.

### DIFF
--- a/micro-rdk/src/esp32/conn/network.rs
+++ b/micro-rdk/src/esp32/conn/network.rs
@@ -109,7 +109,7 @@ impl Esp32WifiNetwork {
                 if let Err(e) = esp_idf_svc::sys::esp!(unsafe {
                     esp_wifi_set_config(esp_interface_t_ESP_IF_WIFI_STA, &mut sta_config as *mut _)
                 }) {
-                    log::warn!("couldn't update wifi station config {:?}", e);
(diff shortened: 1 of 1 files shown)


micro-rdk/src/esp32/conn/network.rs
```diff
@@ -86,6 +94,29 @@ impl Esp32WifiNetwork {
...
+            esp_wifi_get_config(esp_interface_t_ESP_IF_WIFI_STA, &mut sta_config as *mut _)
+        }) {
+            Ok(_) => {
+                sta_config.sta.scan_method = wifi_scan_method_t_WIFI_ALL_CHANNEL_SCAN;
+                sta_config.sta.sort_method = wifi_sort_method_t_WIFI_CONNECT_AP_BY_SIGNAL;
+
+                if let Err(e) = esp_idf_svc::sys::esp!(unsafe {
+                    esp_wifi_set_config(esp_interface_t_ESP_IF_WIFI_STA, &mut sta_config as *mut _)
+                }) {
+                    log::warn!("couldn't update wifi station config {:?}", e);
```
acmorrow, line 112: Maybe "couldn't update wifi station scan/sort config"?

acmorrow (approved):
LGTM, makes total sense, and this would explain quite a bit about the flaky wifi behavior we have long observed on esp32.

PR #269, opened 2024-07-17
[RSDK-7676] Return error when configuring unsupported pin as an AnalogReader

allisonschiang:
I couldn't find a good spot to propagate an error with ?, so I just checked to see if analog was an error.


micro-rdk/src/esp32/board.rs
```diff
@@ -213,6 +215,11 @@ impl EspBoard {
             };
             (analogs, pins, i2c_confs)
         };
+        //return BoardError if non-analog pin is configured as an analog pin
```
gvaradarajan, line 218: Putting the question mark back in L. 187 and putting another one before the semi-colon in L. 190 should make these lines unnecessary. Let me know if it doesn't

gvaradarajan (changes requested):

micro-rdk/src/esp32/board.rs
allisonschiang: I see it does let me propagate from collect(), but it leads to some incompatibility type issues I'm trying to fix right now. Thanks! Def better to propagate here then check for an error later on

micro-rdk/src/esp32/board.rs
```diff
@@ -70,28 +70,25 @@ impl EspBoard {
...
+                            let chan: Result<AnalogReaderType<u16>, BoardError> = match v.pin {
                                 32 => {
                                     let p: AnalogReaderType<u16> =
                                         Arc::new(Mutex::new(Esp32AnalogReader::new(
                                             v.name.to_string(),
                                             AdcChannelDriver::<Atten11dB, _>::new(unsafe {
                                                 crate::esp32::esp_idf_svc::hal::gpio::Gpio32::new()
                                             })
-                                            .ok()?,
+                                            .map_err(BoardError::EspError)?,
```
allisonschiang, line 88: Originally, if a esp32analogreader failed to be created (esp32error) we just turned it into a none and ignored it. Do we want to return an error and fail the config now or ignore it like prev?

gvaradarajan: It should return the error and fail to create the board

micro-rdk/src/common/board.rs
```diff
@@ -39,6 +41,9 @@ pub enum BoardError {
     BoardMethodNotSupported(&'static str),
     #[error(transparent)]
     BoardI2CError(#[from] I2CErrors),
+    #[error("Unable to make Adc Channel Driver with Pin")]
```
gvaradarajan, line 44: Make this `#[error(transparent)]` instead

gvaradarajan (approved):
LGTM, one last change so that the error from esp-idf is properly surfaced

PR #271, opened 2024-07-23
[RSDK-8225] Fix panic in signaling path

npmenard:
In some scenario the streamed RPC message can be split and the previous code would fail with a panic in this case. 
This PR fixes this behavior and introduce a test to replicate the issue


acmorrow (approved):
LGTM

PR #272, opened 2024-07-23
[RSDK-8181] stop statically allocating data store memory

gvaradarajan:
This PR makes the maximum amount of storage bytes allocated to a data collector configurable in the `capture_method` configuration for a component using a field called `cache_size_kb`.  This field is purely additive and, if not supplied, will default to 1KB in micro-RDK (with no effect in regards to RDK). Below is an example of its use for a component restricting the total cache size to 10KB

```json
{
    "name": "movement_sensor-1",
    "namespace": "rdk",
    "type": "movement_sensor",
    "model": "fake",
    "attributes": {},
    "service_configs": [
        {
        "type": "data_manager",
        "attributes": {
            "capture_methods": [
            {
                "method": "Readings",
                "capture_frequency_hz": 10,
                "cache_size_kb": 10,
                "additional_params": {}
            }
            ]
        }
        }
    ]
}
```


micro-rdk/src/common/data_collector.rs
```diff
@@ -37,6 +38,10 @@ impl TryFrom<&Kind> for DataCollectorConfig {
                 "capture_frequency_hz".to_string(),
             ))?
             .try_into()?;
+        let capacity = value
+            .get("cache_size")?
+            .unwrap_or(&Kind::NumberValue(5120.0))
```
acmorrow, line 43: This value shows up many many places, so I think it deserves a constant.

micro-rdk/src/common/data_store.rs
```diff
@@ -152,31 +146,26 @@ impl DataStoreReader for StaticMemoryDataStoreReader {
...
 
-impl StaticMemoryDataStore {
-    pub fn new(collector_keys: Vec<ResourceMethodKey>) -> Result<Self, DataStoreError> {
+impl DefaultDataStore {
+    pub fn new(
+        collector_settings: Vec<(ResourceMethodKey, usize)>,
+    ) -> Result<Self, DataStoreError> {
         if !DATA_STORE_INITIALIZED.load(Ordering::Acquire) {
-            if collector_keys.is_empty() {
+            if collector_settings.is_empty() {
```
acmorrow, line 160: ⬆️ - I don't think you need `DATA_STORE_INITIALIZED` anymore.

micro-rdk/src/common/data_store.rs
```diff
@@ -308,7 +297,7 @@ mod tests {
     use crate::common::data_store::DataStoreError;
     use crate::common::data_store::DataStoreReader;
     use crate::common::data_store::WriteMode;
-    use crate::common::data_store::DATA_STORE;
+    // use crate::common::data_store::DATA_STORE;
```
acmorrow, line 300: Remove

acmorrow (changes requested):
I know you said you were pausing this; just sharing the comments I'd already made.

micro-rdk/src/common/data_store.rs
```diff
@@ -628,9 +605,7 @@ mod tests {
         let message_byte_size = data.encoded_len();
         let message_byte_size_total = length_delimiter_len(message_byte_size) + message_byte_size;
 
-        // store was initialized with two keys, so the byte capacity is half the length of DATA_STORE
-        let message_capacity_for_buffer: usize =
-            unsafe { DATA_STORE.len() } / 2 / message_byte_size_total;
+        let message_capacity_for_buffer: usize = 10240 / 2 / message_byte_size_total;
```
mattjperez, line 608: why is this a constant?

mattjperez (approved):
LGTM, just a question about a magic number

micro-rdk/src/common/data_collector.rs
```diff
@@ -13,6 +13,8 @@ use super::{
 
 use thiserror::Error;
 
+pub(crate) const DEFAULT_CACHE_SIZE: f64 = 30240.0;
```
npmenard, line 16: let's keep it as a usize

micro-rdk/src/common/data_collector.rs
```diff
@@ -37,6 +40,10 @@ impl TryFrom<&Kind> for DataCollectorConfig {
                 "capture_frequency_hz".to_string(),
             ))?
             .try_into()?;
+        let capacity = value
+            .get("cache_size")?
+            .unwrap_or(&Kind::NumberValue(DEFAULT_CACHE_SIZE))
```
npmenard, line 45: there should be more checks like min and max value.

npmenard (commented):
I think there are some changes unrelated to adding a capacity attribute to the data collector, would that be ok to move them to a separate PR?

gvaradarajan:
> I think there are some changes unrelated to adding a capacity attribute to the data collector, would that be ok to move them to a separate PR?

Because the cache is now variable size, not including the other changes introduces bugs, so I thought I could just fold them into the PR

micro-rdk/src/common/data_collector.rs
gvaradarajan: because the units here are moving to KB, I've left the it as f64

npmenard (approved):
LGTM

mattjperez (approved):

PR #273, opened 2024-07-23
[RSDK-8324] Remove provisioning from default

mattjperez:


acmorrow (changes requested):

acmorrow:
> TODO: create branch on top of #278 to check for meaningful overlap

Actually it doesn't look to me like there is much overlap at all? Unless you have changes you haven't pushed. The `entry` files are almost entirely unchanged in your review, and my review exclusively changes the entry files. I don't think you need to wait unless you have more changes that do affect the entry files.

acmorrow:
> > TODO: create branch on top of #278 to check for meaningful overlap
> 
> Actually it doesn't look to me like there is much overlap at all? Unless you have changes you haven't pushed. The `entry` files are almost entirely unchanged in your review, and my review exclusively changes the entry files. I don't think you need to wait unless you have more changes that do affect the entry files.

@mattjperez - I've merged #278.

micro-rdk-server/native/main.rs
```diff
@@ -5,33 +5,45 @@ mod native {
...
         env_logger::builder()
             .format_timestamp(Some(env_logger::TimestampPrecision::Millis))
             .init();
 
+        #[cfg(not(has_robot_config))]
+        #[cfg(not(feature = "provisioning"))]
+        {
+            log::error!("cannot create robot without using robot config or provisioning");
+            std::process::exit(1);
+        }
```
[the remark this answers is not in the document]
mattjperez: from our offline discussion, building and flashing are decoupled, so it's possible to have neither a robot_config nor provisioning enabled but still be in a valid state if the target device was has a valid previous configuration.

acmorrow (changes requested):

micro-rdk-server/build.rs
```diff
@@ -48,4 +48,5 @@ fn main() {
             }
         }
     }
+    println!("cargo:rustc-check-cfg=cfg(has_robot_config)");
```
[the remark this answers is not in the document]
mattjperez: this lets me use `has_robot_config` in an `if` statement, eg `if cfg!(has_robot_config)`. which I used once in native/main.rs

Without it, I get the warning 
```
  --> micro-rdk-server/native/main.rs:31:31
   |
31 | ...   let storage = if cfg!(has_robot_config) && ROBOT_ID.is_some() && ROBOT_SECRET.is_some() {
   |                             ^^^^^^^^^^^^^^^^
   |
   = help: expected names are: `clippy`, `debug_assertions`, `doc`, `docsrs`, `doctest`, `feature`, `miri`, `overflow_checks`, `panic`, `proc_macro`, `relocation_model`, `rustfmt`, `sanitize`, `sanitizer_cfi_generalize_pointers`, `sanitizer_cfi_normalize_integers`, `target_abi`, `target_arch`, `target_endian`, `target_env`, `target_family`, `target_feature`, `target_has_atomic`, `target_has_atomic_equal_alignment`, `target_has_atomic_load_store`, `target_os`, `target_pointer_width`, `target_thread_local`, `target_vendor`, `test`, `ub_checks`, `unix`, and `windows`
   = help: consider using a Cargo feature instead
   = help: or consider adding in `Cargo.toml` the `check-cfg` lint config for the lint:
            [lints.rust]
            unexpected_cfgs = { level = "warn", check-cfg = ['cfg(has_robot_config)'] }
   = help: or consider adding `println!("cargo::rustc-check-cfg=cfg(has_robot_config)");` to the top of the `build.r
``` 

It's only being used in one place so can change it, from 
```rs
let storage = if cfg!(has_robot_config) && ROBOT_ID.is_some() && ROBOT_SECRET.is_some() {
    RAMStorage::new("", "", ROBOT_ID.unwrap(), ROBOT_SECRET.unwrap())
} else {
    RAMStorage::default()
};
```

to 
```rs

#[cfg(feature = "provisioning")]
let storage = {...};

#[cfg(not(feature = "provisioningt')]
let storage = RAMStorage::default();
```

micro-rdk-server/esp32/main.rs
```diff
@@ -39,44 +36,46 @@ mod esp32 {
...
+            )
+            .unwrap();
+            eth_configure(&sys_loop, eth).unwrap()
+        };
 
         {
-            micro_rdk::esp32::esp_idf_svc::sys::esp!(unsafe {
-                micro_rdk::esp32::esp_idf_svc::sys::esp_vfs_eventfd_register(
-                    &micro_rdk::esp32::esp_idf_svc::sys::esp_vfs_eventfd_config_t { max_fds: 5 },
+            esp_idf_svc::sys::esp!(unsafe {
```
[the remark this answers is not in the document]
mattjperez: not particularly. It was more about moving this network stuff up rather than the eventfd stuff down.
Will move it higher. 

Any reason we have these braces around the `eventfd_register` call? it's already being conversted to a result and unwrapped, so not sure why it's scoped.

micro-rdk-server/esp32/main.rs
```diff
@@ -39,44 +36,46 @@ mod esp32 {
...
-            }
-        }
+        let mut r = Box::<ComponentRegistry>::default();
+        register_examples(&mut r);
+        let repr = RobotRepresentation::WithRegistry(r);
 
         // When building the server locally if a user gives a "config" (Robot credentials and Wifi Credentials)
         // then the entire provisioning step can be skipped
+        #[allow(unused_mut)]
+        let mut storage = NVSStorage::new("nvs").unwrap();
```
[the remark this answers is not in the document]
mattjperez: I guess not. The methods further down when `has_robot_config` (`store_robot_credentials`, `store_wifi_credentials`) *looked* like they would need a mutable reference, but confirmed in the method signature that it does not.

Makefile
```diff
@@ -97,6 +100,9 @@ size:
 build-esp32-bin:
 	cargo +esp espflash save-image --package micro-rdk-server --merge --chip esp32 target/xtensa-esp32-espidf/micro-rdk-server-esp32.bin -T micro-rdk-server/esp32/partitions.csv -s 4mb  --bin micro-rdk-server-esp32 --target=xtensa-esp32-espidf  -Zbuild-std=std,panic_abort --release
 
+build-esp32-bin-provisioning:
```
mattjperez, line 103: not sure how to make this less copy-paste

micro-rdk-server/build.rs
```diff
@@ -17,6 +17,7 @@ fn main() {
     println!("cargo:rerun-if-changed=viam.json");
     println!("cargo:rerun-if-env-changed=MICRO_RDK_WIFI_SSID");
     println!("cargo:rerun-if-env-changed=MICRO_RDK_WIFI_PASSWORD");
+    println!("cargo:rustc-check-cfg=cfg(has_robot_config)");
```
mattjperez, line 20: was getting a warning during build that suggested to add this.

micro-rdk-server/esp32/main.rs
```diff
@@ -39,44 +40,42 @@ mod esp32 {
...
-        micro_rdk::esp32::esp_idf_svc::log::EspLogger::initialize_default();
+        esp_idf_svc::sys::esp!(unsafe {
+            esp_idf_svc::sys::esp_vfs_eventfd_register(
+                &esp_idf_svc::sys::esp_vfs_eventfd_config_t { max_fds: 5 },
+            )
+        })
+        .unwrap();
+
+        #[cfg(feature = "qemu")]
+        let _network = {
```
mattjperez, line 54: I'm not really sure how this works exactly since we're not passing it into some form of `serve_web_with_external`

micro-rdk-server/esp32/main.rs
```diff
@@ -39,44 +36,46 @@ mod esp32 {
...
+
         #[cfg(has_robot_config)]
         {
-            use micro_rdk::common::credentials_storage::RobotConfigurationStorage;
-            use micro_rdk::common::credentials_storage::RobotCredentials;
-            use micro_rdk::common::credentials_storage::WifiCredentialStorage;
-            use micro_rdk::common::credentials_storage::WifiCredentials;
+            use micro_rdk::common::credentials_storage::{
+                RobotConfigurationStorage, RobotCredentials, WifiCredentialStorage, WifiCredentials,
+            };
```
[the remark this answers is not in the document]
mattjperez: implemented your suggestion

micro-rdk/src/common/provisioning/mod.rs
```diff
@@ -1 +1,23 @@
+#[cfg(feature = "provisioning")]
 pub mod server;
+
+#[derive(Default, Clone)]
+pub struct ProvisioningInfo(crate::proto::provisioning::v1::ProvisioningInfo);
```
gvaradarajan, line 5: nice, good place to move this

micro-rdk-server/esp32/main.rs
```diff
@@ -101,30 +100,35 @@ mod esp32 {
...
+            if !g_spiram_ok {
+                log::info!("spiram not initialized disabling cache feature of the wifi driver");
+                g_wifi_feature_caps &= !(CONFIG_FEATURE_CACHE_TX_BUF_BIT as u64);
+                1
+            } else {
+                3
+            }
+        };
+
+        let info = if cfg!(feature = "provisioning") {
```
gvaradarajan, line 115: should the main code in the project template do something similar? (as in pass None if the user says no to including provisioning capabilities)

micro-rdk-server/esp32/main.rs
```diff
@@ -101,30 +100,35 @@ mod esp32 {
...
             info.set_manufacturer("viam".to_owned());
             info.set_model("test-esp32".to_owned());
-            let storage = NVSStorage::new("nvs").unwrap();
-            serve_web(Some(info), repr, max_connection, storage);
+            Some(info)
+        } else {
+            None
+        };
+
+        if info.is_none() && !storage.has_wifi_credentials() {
```
gvaradarajan, line 124: just clarifying, when we hit this scenario, the only recourse is to flash new firmware right?

gvaradarajan (approved):
LGTM, couple questions

micro-rdk-server/esp32/main.rs
mattjperez: correct, if there's no way for the device to provision itself and there's no wifi credentials then it's isolated. adding another log line with info on what suggested action is.

mattjperez:
@gvaradarajan @acmorrow had to add more changes to the template so it's basically sync'd with what we have in `micro-rdk-server`, here's concise diff https://github.com/viamrobotics/micro-rdk/pull/273/files/6fef7bb90701e8054d239d2e2dcaf2d7be03f59b..12e6f7dc46c5c1fc675887fdefcda668b486e1be

micro-rdk-server/esp32/main.rs
mattjperez: you're right, I sync'd the project template and `micro-rdk-server/esp32/main.rs`

Makefile
mattjperez: like `make build-esp32-bin --features=provisioning`? I think we can using Make flags, but I'm not as familiar with it compared to a command runner like [`just`](https://github.com/casey/just).

Or did you mean just leave this off for now entirely?

micro-rdk-server/esp32/main.rs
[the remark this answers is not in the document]
mattjperez: removed

templates/project/Cargo.toml
```diff
@@ -25,6 +25,7 @@ embedded-hal = { version = "0.2.7", features = ["unproven"]}
 log = "0.4"
 async-channel = "2"
 futures-lite = "1"
+micro-rdk = {version = "0.2.2", git = "https://github.com/viamrobotics/micro-rdk.git", features = ["esp32", "binstart","provisioning"], rev = "a1863c9" }
```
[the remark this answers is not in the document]
mattjperez: removed

micro-rdk-server/native/main.rs
```diff
@@ -5,39 +5,46 @@ mod native {
...
-        {
-            if ROBOT_ID.is_some() && ROBOT_SECRET.is_some() {
-                let ram_storage = RAMStorage::new("", "", ROBOT_ID.unwrap(), ROBOT_SECRET.unwrap());
-                serve_web_with_external_network(None, repr, 3, ram_storage, network);
-            }
-        }
-        #[cfg(not(has_robot_config))]
+
+        let storage = if cfg!(has_robot_config) && ROBOT_ID.is_some() && ROBOT_SECRET.is_some() {
+            RAMStorage::new("", "", ROBOT_ID.unwrap(), ROBOT_SECRET.unwrap())
```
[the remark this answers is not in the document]
mattjperez: I can change build.rs so that `has_robot_config` is only dependent on `viam.json`, not wifi credentials. Then for `esp32`, I can set the wifi credentials only if the static options of SSID/PASSWORD are not none, in lieu of a `has_wifi_creds` cfg flag.

micro-rdk-server/esp32/main.rs
```diff
@@ -101,30 +100,32 @@ mod esp32 {
...
-            log::info!("creating eth object");
-            let eth = EspEth::wrap(
-                EthDriver::new_openeth(Peripherals::take().unwrap().mac, sys_loop_stack.clone())
-                    .unwrap(),
-            )
-            .unwrap();
-            eth_configure(&sys_loop_stack, eth).unwrap()
-        };
+        #[cfg(not(feature = "provisioning"))]
+        serve_web(repr, 3, storage);
```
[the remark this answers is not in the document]
mattjperez: now uses the same 1v3 thing. 

added your infinite deep sleep.

In principle, no. Espflash sends a reset when connecting to a device to get it into a flashable/writable state

micro-rdk-server/esp32/main.rs
```diff
@@ -101,30 +100,32 @@ mod esp32 {
...
+                    3
+                }
+            };
+
             let mut info = ProvisioningInfo::default();
             info.set_manufacturer("viam".to_owned());
             info.set_model("test-esp32".to_owned());
-            let storage = NVSStorage::new("nvs").unwrap();
-            serve_web(Some(info), repr, max_connection, storage);
+            serve_web(Some(info), repr, max_connections, storage);
```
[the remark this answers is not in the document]
mattjperez: implemented

micro-rdk-server/esp32/main.rs
acmorrow: Let's not worry too much about qemu right now.

micro-rdk-server/esp32/main.rs
```diff
@@ -101,30 +101,39 @@ mod esp32 {
...
             info.set_manufacturer("viam".to_owned());
             info.set_model("test-esp32".to_owned());
-            let storage = NVSStorage::new("nvs").unwrap();
-            serve_web(Some(info), repr, max_connection, storage);
+            Some(info)
+        } else {
+            None
+        };
+
+        if info.is_none() && !(storage.has_wifi_credentials() && storage.has_robot_credentials()) {
```
acmorrow, line 125: I wonder a bit if this bit of logic doesn't belong in the entry files rather than the mains. But, I'm just thinking aloud. If I feel strongly about it I'll move it later.

micro-rdk-server/build.rs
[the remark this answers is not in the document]
acmorrow: I think it must be because you are using a newer rust version. But, I don't mind future proofing things a bit. Fine to keep in.

micro-rdk-server/esp32/main.rs
mattjperez: I agree since `serve_web` will now have all the info it needs to determine this undesireable state.
Will add a ticket and TODO comment before merging

micro-rdk-server/native/main.rs
```diff
@@ -22,22 +22,24 @@ mod native {
...
             info.set_model("test-esp32".to_owned());
-            let storage = RAMStorage::default();
-            serve_web_with_external_network(Some(info), repr, 3, storage, network);
-        }
+            Some(info)
+        } else {
+            None
+        };
+
+        serve_web_with_external_network(info, repr, 3, storage, network);
```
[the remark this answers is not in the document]
mattjperez: updated `native` to exit when there is neither provisioning or robot credentials

micro-rdk-server/build.rs
```diff
@@ -36,16 +36,14 @@ fn main() {
...
-        if let Ok(content) = std::fs::read_to_string("viam.json") {
-            if let Ok(cfg) = serde_json::from_str::<Config>(content.as_str()) {
-                println!(
-                    "cargo:rustc-env=MICRO_RDK_ROBOT_SECRET={}",
-                    cfg.cloud.secret
-                );
-                println!("cargo:rustc-env=MICRO_RDK_ROBOT_ID={}", cfg.cloud.id);
-                println!("cargo:rustc-cfg=has_robot_config");
-            }
+    if let Ok(content) = std::fs::read_to_string("viam.json") {
```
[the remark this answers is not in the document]
mattjperez: I don't think it's necessary, since it will have no effect; it's not referenced in `native/main.rs`. 
It's also one less thing to keep track of when hopping between native and esp32 builds.

micro-rdk-server/esp32/main.rs
mattjperez: [RSDK-8445](https://viam.atlassian.net/browse/RSDK-8445)

templates/project/Cargo.toml
```diff
@@ -16,7 +16,7 @@ opt-level = "z"
 [dependencies.micro-rdk]
 version = "0.2.3"
 git = "https://github.com/viamrobotics/micro-rdk.git"
-features = ["esp32", "binstart", "provisioning", {% if camera == "true" %}"camera"{%endif%}]
+features = ["esp32", "binstart", {% if provisioning == "true" %} "provisioning",{% endif %}{% if camera == "true" %}"camera"{%endif%}]
```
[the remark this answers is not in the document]
mattjperez: updated

mattjperez:
@acmorrow think this is ready for a final pass

mattjperez:
@acmorrow I decoupled the presence of provisioning, wifi credentials, and robot configs except for the final check before `serve_web`, otherwise everything is passed down the stack into `common/entry.rs`. We can handle the different combinations there via a `match` statement or something. 

I can add those changes in this PR or another one

micro-rdk-server/build.rs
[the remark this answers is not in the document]
mattjperez: provisioning now included by default, though wifi and robot credentials required when building via template project

micro-rdk-server/esp32/main.rs
```diff
@@ -39,92 +44,80 @@ mod esp32 {
...
-            use micro_rdk::esp32::esp_idf_svc::hal::prelude::Peripherals;
-            log::info!("creating eth object");
-            let eth = EspEth::wrap(
-                EthDriver::new_openeth(Peripherals::take().unwrap().mac, sys_loop_stack.clone())
-                    .unwrap(),
-            )
-            .unwrap();
-            eth_configure(&sys_loop_stack, eth).unwrap()
-        };
+        serve_web(Some(info), repr, max_connections, storage);
```
mattjperez, line 120: @acmorrow I left `serve_web` take an Option as input because `micro-rdk-ffi` still takes `Option<ProvisioningInfo>` in the `serve_` functions it uses. If we want to change it later, we can do it in a dedicated PR after these merge.

mattjperez:
@acmorrow ready for another look. Confirmed building from both `esp32` and `native` from repo and from generated project.

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -35,6 +35,27 @@ use hyper::{
 };
 use prost::Message;
 
+#[derive(Default, Clone)]
```
acmorrow, line 38: Did this need to move up here? If not and it is just fallout, I'd recommend moving it back to its original location.

micro-rdk/src/esp32/entry.rs
```diff
@@ -35,7 +34,7 @@ use crate::{
 // If they are invalid or absent it will start the provisioning server. Once provision is done it invokes the main server.
 async fn serve_async<S>(
     exec: Executor,
-    #[cfg(feature = "provisioning")] info: Option<ProvisioningInfo>,
+    info: ProvisioningInfo,
```
acmorrow, line 37: Why did the `Option`-ality go away here?

templates/project/build.rs
```diff
@@ -1,16 +1,59 @@
...
     if Regex::new(r"\w+-esp3?2?s?\d?-espidf")
         .unwrap()
         .is_match(&env::var("TARGET").unwrap())
     {
-        embuild::build::CfgArgs::output_propagated("MICRO_RDK").unwrap();
-        embuild::build::LinkArgs::output_propagated("MICRO_RDK").unwrap();
+        println!("cargo:rerun-if-changed=viam.json");
+        println!("cargo:rerun-if-env-changed=MICRO_RDK_WIFI_SSID");
+        println!("cargo:rerun-if-env-changed=MICRO_RDK_WIFI_PASSWORD");
+        if env::var("TARGET").unwrap() == "xtensa-esp32-espidf" {
```
acmorrow, line 26: I think this isn't quite right to be filtering on this target. Doesn't this mean that if you built, for, say esp32s3 that you wouldn't get values set for wifi ssid / password even if they were in the environment?

And, actually, now that I think about it: I don't see any real reason that handing the SSID  and PASSWORD needs to be gated on `TARGET` at all. That is really just here to gate doing the `embuild` stuff. In which case, I'd argue for the same change to micro-rdk-server/build.rs

templates/project/build.rs
```diff
@@ -1,16 +1,59 @@
...
+}
+
 fn main() {
     if Regex::new(r"\w+-esp3?2?s?\d?-espidf")
         .unwrap()
         .is_match(&env::var("TARGET").unwrap())
     {
-        embuild::build::CfgArgs::output_propagated("MICRO_RDK").unwrap();
-        embuild::build::LinkArgs::output_propagated("MICRO_RDK").unwrap();
+        println!("cargo:rerun-if-changed=viam.json");
```
acmorrow, line 23: Maybe move this line down to the block below where we are dealing with `viam.json`?

micro-rdk-server/esp32/main.rs
```diff
@@ -39,92 +44,80 @@ mod esp32 {
...
-            micro_rdk::esp32::esp_idf_svc::sys::esp!(unsafe {
-                micro_rdk::esp32::esp_idf_svc::sys::esp_vfs_eventfd_register(
-                    &micro_rdk::esp32::esp_idf_svc::sys::esp_vfs_eventfd_config_t { max_fds: 5 },
-                )
-            })
-            .unwrap();
+        let storage = NVSStorage::new("nvs").unwrap();
+
+        // At runtime, if the program does not detect credentials or configs in storage,
+        // it will try to load statically compiled values.
```
acmorrow, line 76: OK, so this is a meaningful behavior change. It used to be that we would just always use the static ones if we had them and step on NVS. Now, it is the other way: if NVS has values, we prefer them. I think that is right: you might have a build with static configs, but then have adjusted information with the installer tool.

acmorrow (approved):
LGTM all minor things.

templates/project/build.rs
mattjperez: removed the conditional for TARGET and QEMU from project template build.rs

micro-rdk/src/esp32/entry.rs
mattjperez: reverted back to Option

micro-rdk/src/common/provisioning/server.rs
mattjperez: moved back

micro-rdk-server/esp32/main.rs
mattjperez: @npmenard @gvaradarajan @EshaMaharishi @stevebriskin , after talking more with @acmorrow:

If a device is flashed a) from our published binary or b) build from source repo without wifi credentials, it can access provisioning whenever wifi credentials are not present in NVS, as `PASS` and `SSID` will be `None`.

Devices using builds from the project template (which now requires the presence of wifi credentials), will not be able to access provisioning, even if the NVS cache is completely wiped, as `PASS` and `SSID` will always be `Some()`.

In both scenarios, wifi credentials added post-flashing using the installer will be prioritized.

EshaMaharishi: Sgtm. My understanding: If the user doesn't provide appropriate credentials, they may enter provisioning mode which will appear to them as being bricked. They will need to re-flash.

PR #274, opened 2024-07-23
[RSDK-8127] on invalid config create empty robot over panicking

gvaradarajan:


micro-rdk/src/esp32/entry.rs
```diff
@@ -121,16 +121,15 @@ pub async fn serve_web_inner<S>(
...
                         let logs = vec![config_log_entry(datetime, Some(err))];
                         app_client
                             .push_logs(logs)
                             .await
                             .expect("could not push logs to app");
                     }
-                    // TODO(RSDK-8172) shouldn't panic here, when we support offline mode and
-                    // reloading configuration this should be removed
-                    panic!("couldn't build robot");
+                    LocalRobot::new()
```
acmorrow, line 132: - I think the logged error should state that we couldn't build the robot from the *config* served by app, and that we are defaulting to creating an empty robot.
- Is there a way we can avoid writing the log push block twice?
- We probably shouldn't `expect` on the log push either since that will panic if it fails.

All these comments for the native entry as well.

micro-rdk/src/common/board.rs
```diff
@@ -130,6 +132,10 @@ impl FakeBoard {
     }
 
     pub(crate) fn from_config(cfg: ConfigType) -> Result<BoardType, BoardError> {
+        if cfg.get_attribute::<bool>("fail_new").unwrap_or(false) {
```
acmorrow, line 135: Is there a test that was supposed to leverage this? Or was it just for hand testing? Do you want to keep it in? Maybe we should reserve a name in the component namespace for an invalid component (`rdk:builtin:invalid`) which must never be registered?

acmorrow (approved):
LGTM with some optional suggestions

micro-rdk/src/common/board.rs
gvaradarajan: This was just for hand testing. The reason I opted for this instead of a new model was because this attribute and functionality already exists for the fake board in RDK and I thought we would benefit from the consistency

acmorrow: OK if you are doing as macro-RDK does then good by me. I agree we should be consistent.

mattjperez (approved):
LGTM

micro-rdk/src/esp32/entry.rs
```diff
@@ -103,36 +103,23 @@ pub async fn serve_web_inner<S>(
...
-                    // TODO(RSDK-8172) shouldn't panic here, when we support offline mode and
-                    // reloading configuration this should be removed
-                    panic!("couldn't build robot");
+                    log::error!("could not build robot from config due to {:?}, defaulting to empty robot until a valid config is accessible", err);
+                    (LocalRobot::new(), Some(err))
                 }
             };
+            if let Some(datetime) = cfg_received_datetime {
+                let logs = vec![config_log_entry(datetime, err)];
+                app_client.push_logs(logs).await;
```
acmorrow, line 121: You could still log an error here (and in native) if this failed maybe? I'm not sure how useful it will be in practice though. Up to you.

acmorrow (approved):
LGTM

PR #277, opened 2024-07-25
[RSDK-8201] add heap sensor as builtin component

mattjperez:
This PR is just to migrate the `free_heap_sensor` from examples to the main micro-rdk builtins with minimal modifications. 

A follow-up ticket [RSDK-8353](https://viam.atlassian.net/browse/RSDK-8353) has been made to migrate builtin-sensors to a separate crate in the workspace.

Currently, it's not feature-gated from builtins due to its utility, but it may be worthwhile to eventually feature-gate this and similar tools under a `debug` flag or something. I can gate it in this PR behind the builtin-components flag if that's the concensus.

[RSDK-8353]: https://viam.atlassian.net/browse/RSDK-8353?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/micro-rdk-server/Cargo.toml
+++ b/micro-rdk-server/Cargo.toml
@@ -18,6 +18,7 @@ qemu = []
 env_logger.workspace = true
 local-ip-address.workspace = true
 micro-rdk = { workspace = true, features = ["native", "provisioning"], default-features = true }
+micro-rdk-modular-driver-example.workspace = true
 
 
 [target.'cfg(target_os="espidf")'.dependencies]
@@ -26,7 +27,7 @@ embedded-hal.workspace = true
 embedded-svc.workspace = true
 futures-lite.workspace = true
 micro-rdk = { workspace = true, features = ["esp32", "binstart", "provisioning"], default-features = true }
-micro-rdk-modular-driver-example = {path = "../examples/modular-drivers", features = ["esp32"]}
+micro-rdk-modular-driver-example = {workspace = true, features = ["esp32"]}
 
 
 [dependencies]
--- a/micro-rdk-server/Cargo.toml
+++ b/micro-rdk-server/Cargo.toml
@@ -26,10 +26,10 @@ embedded-hal.workspace = true
 embedded-svc.workspace = true
 futures-lite.workspace = true
 micro-rdk = { workspace = true, features = ["esp32", "binstart", "provisioning"], default-features = true }
+micro-rdk-modular-driver-example = {path = "../examples/modular-drivers", features = ["esp32"]}
 
 
 [dependencies]
-micro-rdk-modular-driver-example = {path = "../examples/modular-drivers", features = ["esp32"]}
 log.workspace = true
 serde.workspace = true
 serde_json.workspace = true
--- a/micro-rdk-server/Cargo.toml
+++ b/micro-rdk-server/Cargo.toml
@@ -29,6 +29,7 @@ micro-rdk = { workspace = true, features = ["esp32", "binstart", "provisioning"]
 
 
 [dependencies]
+micro-rdk-modular-driver-example = {path = "../examples/modular-drivers", features = ["esp32"]}
 log.workspace = true
 serde.workspace = true
 serde_json.workspace = true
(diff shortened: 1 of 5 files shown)


acmorrow:
I thought the decision in the meeting was to go the other way: to leave this (an RSSI sensor) as example modules in their own crate, but then have `micro-rdk-server` import them.

mattjperez:
> I thought the decision in the meeting was to go the other way: to leave this (an RSSI sensor) as example modules in their own crate, but then have `micro-rdk-server` import them.

You're right, I think I got jumbled with the other statements right after about moving builtin sensors to a separate crate and this would be included.

Will revert it

mattjperez:
confirmed in meeting, going with just importing directly from examples into `micro-rdk-server` entry

micro-rdk-server/Cargo.toml
```diff
@@ -29,6 +29,7 @@ micro-rdk = { workspace = true, features = ["esp32", "binstart", "provisioning"]
 
 
 [dependencies]
+micro-rdk-modular-driver-example = {path = "../examples/modular-drivers", features = ["esp32"]}
```
mattjperez, line 32: using `micro-rdk-modular-driver-example.workspace = true` does not work, fails with an invalid workspace error.

micro-rdk-server/Cargo.toml
```diff
@@ -26,6 +26,7 @@ embedded-hal.workspace = true
 embedded-svc.workspace = true
 futures-lite.workspace = true
 micro-rdk = { workspace = true, features = ["esp32", "binstart", "provisioning"], default-features = true }
+micro-rdk-modular-driver-example = {path = "../examples/modular-drivers", features = ["esp32"]}
```
mattjperez, line 29: `micro-rdk-modular-driver-example = {workspace = true, features = ["esp32"]}` results in an invalid workspace error

mattjperez:
Confirmed working on app

micro-rdk-server/esp32/main.rs
```diff
@@ -34,7 +34,11 @@ mod esp32 {
 
         micro_rdk::esp32::esp_idf_svc::log::EspLogger::initialize_default();
 
-        let repr = RobotRepresentation::WithRegistry(Box::<ComponentRegistry>::default());
+        let mut r = Box::<ComponentRegistry>::default();
+        if let Err(e) = micro_rdk_modular_driver_example::register_models(&mut r) {
```
acmorrow, line 38: - This is going to register the water pump and moisture sensor too. Is that intended? f you just want to register just the heap and rssi sensors, then you I think will need to make their individual registration functions `pub fn` in `micro-rdk-modular-driver-example` so you can call them directly here.
- Should this be gated on `builtin-components` somehow? Though I suppose that is a `micro-rdk` feature not a `micro-rdk-server` feature.

acmorrow (changes requested):

micro-rdk-server/esp32/main.rs
mattjperez: - changed to just heap sensor
- I see what you mean. I don't feel strongly about it either way. I know the intention is to have this available out of the box for users so they can see immediately how app works.

micro-rdk-server/esp32/main.rs
```diff
@@ -34,7 +34,12 @@ mod esp32 {
 
         micro_rdk::esp32::esp_idf_svc::log::EspLogger::initialize_default();
 
-        let repr = RobotRepresentation::WithRegistry(Box::<ComponentRegistry>::default());
+        let mut r = Box::<ComponentRegistry>::default();
+        if let Err(e) = micro_rdk_modular_driver_example::free_heap_sensor::register_models(&mut r)
```
acmorrow, line 38: I think you should also register the RSSI sensor, or at least that is what the ticket seems to ask for. I think you will need to adjust the visibility of the register function.

mattjperez: @EshaMaharishi is this in line with what you wanted for features included in the binary hosted on github, at the same time users will reimplement/re-register the WIFI sensor when learning to add modules?

EshaMaharishi: I wanted to leave at least one sensor that doesn't require any hardware that works with the modular driver setup instructions.

Though, now I'm confused again... did we conclude that if we add a sensor this way, it will only affect the micro-rdk-installer flow, not the cargo flow?

acmorrow (approved):
LGTM

PR #278, opened 2024-07-25
[RSDK-8124] Extract serve_web_inner to common

acmorrow:
This combines the two (native and esp32) implementations of `serve_web_inner` under a common implementation, by way of using `cfg` blocks in the implementation.

This is (or at least is intended to be) a purely mechanical change, though the order of some blocks changed around a bit relative to the originals, which were also slightly different in some non-meaningful ways.

I don't love it, but I like it better than the duplicated code, and it unblocks doing this to the next duplicative node(s) up the graph: the `serve_async_with_external_network` variants, which are almost identical. It wasn't possible to easily eliminate the duplication since they called into different functions at the end, but now they call the same function.


acmorrow:
Updated: with `serve_web_inner` factored out, it became possible to do `serve_async_with_external_network` as well.

acmorrow:
@gvaradarajan and @mattjperez tagging you for this one, please take a look. It is going to be a pain to review because of all the code motion. Sorry in advance.

gvaradarajan (approved):
LGTM

mattjperez (approved):
LGTM

PR #279, opened 2024-07-25
hotfix: template project toml invalid newlines

mattjperez:
Cargo doesn't allow for newlines when we split it out. I had tested the generation but somehow hadn't checked if compilation was broken.


acmorrow (approved):
LGTM

PR #280, opened 2024-07-25
update revs to include camera changes

mattjperez:


templates/module/Cargo.toml
```diff
@@ -7,7 +7,7 @@ resolver = "2"
 rust-version = "1.75"
 
 [dependencies]
-micro-rdk = {git = "https://github.com/viamrobotics/micro-rdk.git", features = ["{{mcu}}"], version = "0.2.3", rev = "098d3f1"}
+micro-rdk = {git = "https://github.com/viamrobotics/micro-rdk.git", features = ["{{mcu}}"], version = "0.2.3", rev = "78d13fd"}
```
acmorrow, line 10: Looking at https://github.com/viamrobotics/micro-rdk/commit/a1ce24c98e5601ef94de5072512311d38ef8a9e0, I think this commit needs a `cargo.lock` update too.

acmorrow (changes requested):
What motivates this change?

mattjperez:
> What motivates this change?
That hash points to the latest camera PR. Without it, new projects will not have access to the new esp32camera changes and will try and fail to compile the outdated esp32 camera implementation.

acmorrow:
> > What motivates this change?
> 
> That hash points to the latest camera PR. Without it, new projects using the camera feature will not have access to the new esp32camera changes and will try and fail to compile the outdated esp32 camera implementation.

Sure, but only until 0.2.4 is released, right? At which point camera features would become available naturally as part of the new release when the revs get bumped for that?

Until then, someone working with a project can manually advance the hashes if they want camera access?

Doing this now sort of releases camera support into projects before those features have been released in micro-rdk.

mattjperez:
Ah OK, didn't realize it was only on a release basis. Closing

acmorrow:
> Ah OK, didn't realize it was only on a release basis. Closing

I'm not sure there is a hard rule about it, but it just makes more sense to me that the feature would do out in the templates at the same time as a release.

PR #282, opened 2024-07-29
[RSDK-8117] update installer wifi prompt

mattjperez:


micro-rdk-installer/src/main.rs
```diff
@@ -190,7 +190,7 @@ fn request_wifi(
         ssid
     } else {
         Input::with_theme(&ColorfulTheme::default())
-            .with_prompt("Please enter WiFi SSID")
+            .with_prompt("Please enter WiFi SSID (WiFi Network Name) for the network that micro-rdk will connect to")
```
acmorrow, line 193: If we are cleaning this up, maybe:
- "Please enter the"
-  put an `i.e.` inside the parens
- remove the capitals from `Network Name` since it isn't a proper noun
- Capitalize to `Micro-RDK` 

```
Please enter the WiFi SSID (i.e. WiFi network name) for the network that Micro-RDK will connect to
```

acmorrow (approved):
LGTM

PR #283, opened 2024-07-29
[RSDK-8123] Unify executors to remove unnecessary cfg blocks

gvaradarajan:


mattjperez (approved):
Nice!

nit- I think we should use [upper_case_acronyms](https://rust-lang.github.io/rust-clippy/master/#/upper_case_acronyms), `CpuBoundExecutor`

Otherwise, LGTM

micro-rdk/src/common/camera/fake_camera.rs
```diff
@@ -94,15 +94,13 @@ mod tests {
             app_client::encode_request,
             config::DynamicComponentConfig,
             conn::server::{AsyncableTcpListener, Http2Connector},
+            exec::CPUBoundExecutor,
```
acmorrow, line 97: I think I somewhat dislike this naming. It doesn't seem to me like any of this code actually depends on the CPU-bound-ness of the executor. At least I hope it doesn't. It should all work just as well if we had a executor that could move from core to core. We just don't happen to configure for that right now. Is there a reason that it can't just be `common::exec::Executor`?

micro-rdk/src/common/exec.rs
```diff
@@ -11,13 +11,13 @@ use crate::common::webrtc::exec::WebRtcExecutor;
 
 #[derive(Clone, Debug, Default)]
 /// This executor is local and bounded to the CPU that created it usually you would create it after spwaning a thread on a specific core
-pub struct NativeExecutor {}
+pub struct CPUBoundExecutor {}
```
acmorrow, line 14: Here I think I'd keep it as `CPUBoundExecutor`, and then re-export it as `Executor`. That way, later, if we have an environment where we had two cores available to micro-rdk (on, say, an ESP32-P4 which has two performance cores and one lesser core), we could swap in another implementation based on features tests or targeting.

But actually, it really isn't pinned to a CPU. It is pinned to a thread. It just so happens that we have pinned that thread to a core.

acmorrow (changes requested):

acmorrow (approved):
LGTM

PR #284, opened 2024-07-29
[RSDK-8632] improve monitor message

mattjperez:


micro-rdk-installer/src/main.rs
```diff
@@ -523,3 +524,7 @@ fn update_app_image(args: &AppImageArgs) -> Result<(), Error> {
     }
     Ok(())
 }
+
+fn monitor_message() {
+    log::info!("In order to begin monitoring (viewing the log output), you must run the command to reset the chip");
```
acmorrow, line 529: Maybe add (`e.g. Ctrl-R`)

acmorrow (approved):
LGTM with a small suggestion

gvaradarajan (approved):
LGTM, I agree with @acmorrow 's comment unless the CTRL-R hint appears after this message regardless

mattjperez:
> LGTM, I agree with @acmorrow 's comment unless the CTRL-R hint appears after this message regardless

the `CTRL-R` hint appears regardless in all cases. I was hesitant to add a specific key combination and just used the same wording to `reset the chip` that's used in the hint, but it looks like `espflash` is feature complete and the devs don't plan on more than maintenance, so the actual mapping probably won't change in the future. Since the suggestion doesn't hurt, I'll add it.

PR #286, opened 2024-07-31
[RSDK-8121] - separate RPC host handling from robot credentials

gvaradarajan:
addressing the request from [this PR](https://github.com/viamrobotics/micro-rdk/pull/251/files#r1664613484)

### DIFF
--- a/micro-rdk/src/common/app_client.rs
+++ b/micro-rdk/src/common/app_client.rs
@@ -30,6 +30,7 @@ use crate::proto::{
 };
 
 use super::conn::network::Network;
+use super::credentials_storage::RobotCredentials;
 use super::{
     grpc_client::{GrpcClient, GrpcClientError, GrpcMessageSender, GrpcMessageStream},
     webrtc::{
@@ -66,44 +67,9 @@ pub enum AppClientError {
     AppClientRequestTimeout,
 }
 
-#[derive(Debug, Clone)]
-pub struct AppClientConfig {
-    robot_id: String,
-    robot_secret: String,
-    rpc_host: String,
-}
-
-impl Default for AppClientConfig {
-    fn default() -> Self {
-        Self {
-            robot_id: "".to_owned(),
-            robot_secret: "".to_owned(),
-            rpc_host: "".to_owned(),
-        }
-    }
-}
-
-impl AppClientConfig {
-    pub fn new(robot_secret: String, robot_id: String, rpc_host: String) -> Self {
-        AppClientConfig {
-            robot_id,
-            robot_secret,
-            rpc_host,
-        }
-    }
-
-    pub fn get_robot_id(&self) -> String {
-        self.robot_id.clone()
-    }
-
-    pub fn set_rpc_host(&mut self, rpc_host: String) {
-        self.rpc_host = rpc_host
-    }
-}
-
 pub struct AppClientBuilder {
     grpc_client: Box<GrpcClient>,
-    config: AppClientConfig,
+    robot_credentials: RobotCredentials,
 }
 
 pub(crate) fn encode_request<T>(req: T) -> Result<Bytes, AppClientError>
@@ -124,10 +90,10 @@ where
 
 impl AppClientBuilder {
     /// Create a new AppClientBuilder
-    pub fn new(grpc_client: Box<GrpcClient>, config: AppClientConfig) -> Self {
+    pub fn new(grpc_client: Box<GrpcClient>, config: RobotCredentials) -> Self {
         Self {
             grpc_client,
-            config,
+            robot_credentials: config,
         }
     }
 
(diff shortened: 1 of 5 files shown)


micro-rdk/src/common/conn/server.rs
```diff
@@ -95,11 +95,12 @@ pub struct ViamServerBuilder<M, C, T, NetworkType, CC = WebRtcNoOp, D = WebRtcNo
     _marker: PhantomData<T>,
     exec: Executor,
     app_connector: C,
-    app_config: AppClientConfig,
+    robot_credentials: RobotCredentials,
     max_connections: usize,
     app_client_tasks: Vec<Box<dyn PeriodicAppClientTask>>,
     network: NetworkType,
     app_client: Option<AppClient>,
+    rpc_host: String,
```
acmorrow, line 103: Now that we have a clearer understanding of what this field is for, can we improve its name?

micro-rdk/src/common/entry.rs
```diff
@@ -105,8 +100,22 @@ pub async fn serve_web_inner<S>(
...
+
+    let rpc_host = if let Some(robot_config) = cfg_response.config.as_ref() {
+        if let Some(cloud_config) = robot_config.cloud.as_ref() {
+            cloud_config.fqdn.clone()
+        } else {
+            "".to_string()
+        }
+    } else {
+        "".to_string()
+    };
```
acmorrow, line 118: Is there some way to write this with more map/default sort of handling rather than the nested `if let`s?

micro-rdk/src/common/config_monitor.rs
```diff
@@ -64,8 +64,8 @@ where
         app_client: &'c AppClient,
     ) -> Pin<Box<dyn Future<Output = Result<Option<Duration>, AppClientError>> + 'c>> {
         Box::pin(async move {
-            if let Ok((_app_client_config, new_config, _cfg_received_datetime)) =
-                app_client.clone().get_config(None).await
+            if let Ok((new_config, _cfg_received_datetime)) =
+                app_client.clone().get_app_config(None).await
```
acmorrow, line 68: Do you still need the `.clone` here?

micro-rdk/src/common/app_client.rs
```diff
@@ -301,18 +245,13 @@ impl AppClient {
         };
 
         let cfg_response = ConfigResponse::decode(r.split_off(5))?;
-        if let Some(robot_config) = cfg_response.config.as_ref() {
```
acmorrow, line 304: I do sort of see why it made sense to have the `rpc_host` be part of the `AppClient`, because without it `AppClient` doesn't manage all the state it needs to perform all the various tasks against app. However, I also don't see a way to keep mutation out of `AppClient`  if it is supposed to manage that state, since it only learns of it via the `get_config` method. It makes me wonder a bit if `get_config` belongs on `AppClient` at all, or more the responsibility of `AppClientBuilder`.

acmorrow (changes requested):
This looks good and I'm happy to approve it if you want to stop here, but do wonder a bit if there is a slightly better way where `AppClient` can continue to manage the `rpc_host` (perhaps under a better name) by re-homing `get_config` onto `AppClientBuilder` in some way.

micro-rdk/src/common/conn/server.rs
gvaradarajan: I want to keep it this way because it matches the corresponding header name on the gRPC request, would you settle for a comment?

micro-rdk/src/common/config_monitor.rs
gvaradarajan: ah no, probably not

micro-rdk/src/common/app_client.rs
gvaradarajan: I disagree, the idea of a host isn't really "client-side" behavior, so it makes sense that a client should not be in charge of the state of the host. It's the server that is using an application client to act as a host making it a better candidate for maintaining the state of something like `rpc_host`

micro-rdk/src/common/entry.rs
```diff
@@ -105,8 +100,26 @@ pub async fn serve_web_inner<S>(
...
+        .config
+        .clone()
+        .unwrap_or(RobotConfig {
+            ..Default::default()
+        })
+        .cloud
+        .clone()
+        .unwrap_or(CloudConfig {
+            ..Default::default()
+        })
```
mattjperez, line 120: is it so we know what's being unwrapped at each level that you're not using `unwrap_or_default()`?

mattjperez (approved):
LGTM, couple non-blocking comments

micro-rdk/src/common/app_client.rs
acmorrow: OK, I'm reasonably convinced.

micro-rdk/src/common/conn/server.rs
acmorrow: No, that's fine then if it is following the protos. I don't think it needs a comment.

acmorrow (approved):
LGTM mod @mattjperez's question

micro-rdk/src/common/entry.rs
gvaradarajan: I tried, `Default` isn't implemented for those types

PR #287, opened 2024-07-31
Installer write flash prompt

mattjperez:
as `micro-rdk-installer write-flash` doesn't require a path or version to run, it will flash the device completely with the latest build immediately, overwriting stored configs.

During development, whether by poor naming of the options or me misremembering, I've had to quickly cancel so my configs don't get wiped. This prompt will give a user a chance to decline if this (or flashing with latest) is not the desired outcome.

For scripts where flashing latest is the desired behavior, `micro-rdk-installer write-flash --version latest` will work.

### DIFF
--- a/micro-rdk-installer/src/main.rs
+++ b/micro-rdk-installer/src/main.rs
@@ -349,7 +349,6 @@ fn main() -> Result<(), Error> {
             )?;
         }
         Some(Commands::WriteFlash(args)) => {
-            let config = Config::load().map_err(|err| Error::SerialConfigError(err.to_string()))?;
             let tmp_path = tempfile::NamedTempFile::new()
                 .map_err(Error::FileError)?
                 .path()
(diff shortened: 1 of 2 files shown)


micro-rdk-installer/src/main.rs
```diff
@@ -349,18 +351,29 @@ fn main() -> Result<(), Error> {
...
             let tmp_path = tempfile::NamedTempFile::new()
                 .map_err(Error::FileError)?
                 .path()
                 .to_path_buf();
             let app_path = match &args.binary_path {
                 Some(path) => PathBuf::from(path),
                 None => {
+                    if args.version.is_none() {
+                        let use_latest = dialoguer::Confirm::new()
+                            .with_prompt("Download and flash the latest Micro-RDK release?")
```
acmorrow, line 363: What happens here if stdin is not a tty?

micro-rdk-installer/src/nvs/request.rs
```diff
@@ -22,7 +22,16 @@ pub async fn download_micro_rdk_release(
                 RELEASES_BASE_URL, "latest/download", BINARY_NAME
             )
         },
-        |version| format!("{}/download/{}/{}", RELEASES_BASE_URL, version, BINARY_NAME),
+        |version| {
```
acmorrow, line 25: ⬆️ - If you defaulted in `latest` above could you avoid the need to re-iterate the format block for the latest case?

gvaradarajan (approved):
LGTM

micro-rdk-installer/src/nvs/request.rs
mattjperez: I can do this, was avoiding it for the clone, but as this is a cli, it doesn't really matter.
```rs
let release_url = if version.is_none() || version.clone().unwrap() == "latest" { ... } else { ... };
```

Better?

acmorrow: Yes, I think so.

micro-rdk-installer/src/main.rs
mattjperez: confirmed with `micro-rdk-installer write-flash < /dev/null` that it doesn't hang, but somehow still results in an interactive prompt.

acmorrow: Alright, not worth spending more time on that I can see.

acmorrow (approved):
LGTM

PR #288, opened 2024-08-02
RSDK-8444 Add support for xtensa-esp32s3-espidf

acmorrow:
Per https://viam.atlassian.net/browse/RSDK-8444 (which I just filed), the esp32s3 platform can make tflite micro run much much faster when using esp-nn. This change adds support for esp32s3. I didn't bother with esp32s2 but I suppose it wouldn't be much more work if folks think it is interesting, but I don't have that board to try to flash to.

I will leave a few comments on the change about things I already know I want to address.


.gitignore
```diff
@@ -4,5 +4,5 @@ Cargo.lock
 /micro-rdk-server/viam.json
 /canary/.venv/*
 .vscode
-components_esp32.lock
+components_esp32*.lock
```
acmorrow, line 7: I wonder if we are supposed to be checking these files in, much like we do with `Cargo.lock`?

micro-rdk/src/esp32/board.rs
```diff
@@ -68,6 +73,9 @@ impl EspBoard {
     /// The potential approach is described in esp32/motor.rs:383
     pub(crate) fn from_config(cfg: ConfigType) -> Result<BoardType, BoardError> {
         let (analogs, mut pins, i2c_confs) = {
+            #[cfg(not(esp32))]
```
acmorrow, line 76: This needs a ticket and a TODO added. There will be no analog pin support on esp32s3 until this is fixed.

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -34,13 +34,10 @@ use esp_idf_svc::{
...
-use crate::{
-    common::credentials_storage::WifiCredentials,
-    esp32::esp_idf_svc::{
-        eth::{BlockingEth, EspEth, OpenEth},
-        sys::EspError,
-    },
-};
+use crate::{common::credentials_storage::WifiCredentials, esp32::esp_idf_svc::sys::EspError};
+
+#[cfg(feature = "qemu")]
```
acmorrow, line 39: Probably also another ticket/TODO pair: I don't think qemu is going to work on esp32s3 anymore since the OpenEth thing got taken out, and I'm not sure what to do about it.

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -174,6 +171,7 @@ impl Network for Esp32WifiNetwork {
     }
 }
 
+#[cfg(feature = "qemu")]
```
acmorrow, line 174: This stuff probably always should have been feature gated on qemu because my understanding is that it was the only place the openeth stuff was used.

micro-rdk/src/esp32/pin.rs
```diff
@@ -38,7 +38,7 @@ fn install_gpio_isr_service() -> Result<(), BoardError> {
 // ESP-IDF, so we must replicate the logic from that function here. If we do not validate,
 // PinDriver::input_output will panic because esp-idf-hal does not perform the same check
 fn is_valid_gpio_pin(pin: i32) -> Result<(), BoardError> {
-    match (1_u32 << (pin as u32)) & SOC_GPIO_VALID_DIGITAL_IO_PAD_MASK {
+    match (1_u32 << (pin as u32)) & (SOC_GPIO_VALID_DIGITAL_IO_PAD_MASK as u32) {
```
acmorrow, line 41: This confused me. For esp32s3 it was acting as if this macro was a u64 but I don't see how that could be. @gvaradarajan definitely interested if you have any insights since you added this recently.

acmorrow:
If you have an esp32s3 and want to try this PR locally, you need to have the following in your build / flash commands:

```
env MCU=esp32s3 ... --chip esp32s3 ...  target/xtensa-esp32s3-espidf/... ... --target=xtensa-esp32s3-espidf ...
```

Yes, that's too many times saying the same thing in slightly different ways, but I don't yet have a solution at hand to make it easier.

micro-rdk/src/esp32/pin.rs
mattjperez: how did you observe it was acting as a u64?

acmorrow: Without the cast, I get the following on esp32s3 targets:
```
error: could not compile `micro-rdk` (lib) due to 3 previous errors
error[E0308]: mismatched types
  --> micro-rdk/src/esp32/pin.rs:41:37
   |
41 |     match (1_u32 << (pin as u32)) & SOC_GPIO_VALID_DIGITAL_IO_PAD_MASK {
   |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `u64`

error[E0277]: no implementation for `u32 & u64`
  --> micro-rdk/src/esp32/pin.rs:41:35
   |
41 |     match (1_u32 << (pin as u32)) & SOC_GPIO_VALID_DIGITAL_IO_PAD_MASK {
   |                                   ^ no implementation for `u32 & u64`
   |
   = help: the trait `BitAnd<u64>` is not implemented for `u32`
   = help: the following other types implement trait `BitAnd<Rhs>`:
             <u32 as BitAnd>
             <u32 as BitAnd<&u32>>
             <&'a u32 as BitAnd<u32>>
             <&u32 as BitAnd<&u32>>

error: aborting due to 2 previous errors

Some errors have detailed explanations: E0277, E0308.
For more information about an error, try `rustc --explain E0277`.
```

Unfortunately, *with* the cast I get clippy complaints on plain esp32, per CI. See https://github.com/viamrobotics/micro-rdk/actions/runs/10219337838/job/28277453866?pr=288#step:6:815.

Though probably easier to just silence clippy in that case.

mattjperez: Found it.

That constant comes from esp-idf, and it has different implementations depending on the target.
So you actually don't want to be casting here as you'd be missing out on the rest of the mask.

[esp32](https://github.com/espressif/esp-idf/blob/41dd1a351b4f8630a0b9dbdd87a6bf782094dd66/components/soc/esp32/include/soc/soc_caps.h#L190): `#define SOC_GPIO_VALID_DIGITAL_IO_PAD_MASK 0xEF0FEAULL`

[esp32s3](https://github.com/espressif/esp-idf/blob/41dd1a351b4f8630a0b9dbdd87a6bf782094dd66/components/soc/esp32s3/include/soc/soc_caps.h#L187) : `#define SOC_GPIO_VALID_DIGITAL_IO_PAD_MASK 0x0001FFFFFC000000ULL`

acmorrow: @mattjperez - Nice find, thanks for that. I'm going to try to figure out a way that works for both targets.

acmorrow:
@mattjperez - This is ready for another look.

mattjperez (approved):
LGTM

We might want to make a ticket tracking esp32s3 support; maybe a grid for feature parity across supported boards

acmorrow:
> LGTM
> 
> We might want to make a ticket tracking esp32s3 support; maybe a grid for feature parity across supported boards

I will mark RSDK-8444 as docs changes needed.

PR #290, opened 2024-08-07
[RSDK-8008] use Strings instead of static references for ComponentRegistry Keys

mattjperez:
might want to look at changing `DynamicComponentConfig` to use an enum like `enum ComponentType {Motor, Sensor, ...}` in the future instead of a string. Would take the place of all the static `crate::common::[motor,sensor,etc]::COMPONENT_NAME`s.

### DIFF
--- a/examples/modular-drivers/src/free_heap_sensor.rs
+++ b/examples/modular-drivers/src/free_heap_sensor.rs
@@ -21,7 +21,7 @@ use micro_rdk::{
 pub struct FreeHeapSensor;
 
 pub fn register_models(registry: &mut ComponentRegistry) -> Result<(), RegistryError> {
-    registry.register_sensor("free-heap".to_string(), &FreeHeapSensor::from_config)?;
+    registry.register_sensor("free-heap", &FreeHeapSensor::from_config)?;
     log::debug!("free-heap sensor registration ok");
     Ok(())
 }
--- a/examples/modular-drivers/src/free_heap_sensor.rs
+++ b/examples/modular-drivers/src/free_heap_sensor.rs
@@ -21,7 +21,7 @@ use micro_rdk::{
 pub struct FreeHeapSensor;
 
 pub fn register_models(registry: &mut ComponentRegistry) -> Result<(), RegistryError> {
-    registry.register_sensor("free-heap", &FreeHeapSensor::from_config)?;
+    registry.register_sensor("free-heap".to_string(), &FreeHeapSensor::from_config)?;
     log::debug!("free-heap sensor registration ok");
     Ok(())
 }
--- a/micro-rdk/src/common/mpu6050.rs
+++ b/micro-rdk/src/common/mpu6050.rs
@@ -37,7 +37,7 @@ use std::sync::{Arc, Mutex};
 
 pub(crate) fn register_models(registry: &mut ComponentRegistry) {
     if registry
-        .register_movement_sensor("gyro-mpu6050".to_string(), &MPU6050::from_config)
+        .register_movement_sensor("gyro-mpu6050", &MPU6050::from_config)
         .is_err()
     {
         log::error!("gyro-mpu6050 type is already registered");
--- a/micro-rdk/src/common/mpu6050.rs
+++ b/micro-rdk/src/common/mpu6050.rs
@@ -37,7 +37,7 @@ use std::sync::{Arc, Mutex};
 
 pub(crate) fn register_models(registry: &mut ComponentRegistry) {
     if registry
-        .register_movement_sensor("gyro-mpu6050", &MPU6050::from_config)
+        .register_movement_sensor("gyro-mpu6050".to_string(), &MPU6050::from_config)
         .is_err()
     {
         log::error!("gyro-mpu6050 type is already registered");
--- a/micro-rdk/src/common/registry.rs
+++ b/micro-rdk/src/common/registry.rs
@@ -65,7 +65,7 @@ impl ResourceKey {
             "power_sensor" => crate::common::power_sensor::COMPONENT_NAME,
             "generic" => crate::common::generic::COMPONENT_NAME,
             &_ => {
-                return Err(RegistryError::ModelNotFound(model.to_string()));
+                return Err(RegistryError::ModelNotFound(model.into()));
             }
         };
         Ok(Self(model_str.to_string(), name))
@@ -88,7 +88,7 @@ impl TryFrom<ResourceName> for ResourceKey {
             "power_sensor" => crate::common::power_sensor::COMPONENT_NAME,
             "generic" => crate::common::generic::COMPONENT_NAME,
             _ => {
-                return Err(RegistryError::ModelNotFound(comp_type.to_string()));
+                return Err(RegistryError::ModelNotFound(comp_type.into()));
             }
         };
         Ok(Self(comp_name.to_string(), value.name))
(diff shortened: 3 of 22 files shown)


examples/modular-drivers/src/free_heap_sensor.rs
```diff
@@ -21,7 +21,7 @@ use micro_rdk::{
 pub struct FreeHeapSensor;
 
 pub fn register_models(registry: &mut ComponentRegistry) -> Result<(), RegistryError> {
-    registry.register_sensor("free-heap", &FreeHeapSensor::from_config)?;
+    registry.register_sensor("free-heap".to_string(), &FreeHeapSensor::from_config)?;
```
acmorrow, line 24: Any existing modules will have their registration calls broken by this change. See below in `registry` for a suggestion on how to avoid that.

micro-rdk/src/common/registry.rs
```diff
@@ -214,21 +232,21 @@ impl ComponentRegistry {
     #[cfg(feature = "camera")]
     pub fn register_camera(
         &mut self,
-        model: &'static str,
+        model: String,
```
acmorrow, line 235: If this is written as:

```
pub fun register_camera<S: Into<String>>(&mut self, name: S)
```

Then you should be able to pass either `String` or `&str`, and existing callers will not need to change.

micro-rdk/src/common/registry.rs
```diff
@@ -352,20 +370,20 @@ impl ComponentRegistry {
 
     pub(crate) fn get_dependency_function(
         &self,
-        component_type: &'static str,
-        model_name: &str,
+        component_type: String,
```
acmorrow, line 373: If these are just used for lookups, could they remain `&str`? Here and elsewhere for lookups.

mattjperez: specifically `component_type` right? or did you mean `component_type` and `model_name`?

acmorrow: Both, I think, if it works.

examples/modular-drivers/src/free_heap_sensor.rs
mattjperez: reverted examples, though now compatible with both string slices and Strings

micro-rdk/src/common/registry.rs
mattjperez: implemented

micro-rdk/src/common/registry.rs
```diff
@@ -48,11 +48,11 @@ pub fn get_board_from_dependencies(deps: Vec<Dependency>) -> Option<BoardType> {
 // first element is a string representing the component type (arm, motor, etc.)
 // and the second element is its name.
 #[derive(Debug, Clone, Hash, Eq, PartialEq)]
-pub struct ResourceKey(pub &'static str, pub String);
+pub struct ResourceKey(pub String, pub String);
 
 impl ResourceKey {
-    pub fn new(model: &str, name: String) -> Result<Self, RegistryError> {
-        let model_str = match model {
+    pub fn new(model: String, name: String) -> Result<Self, RegistryError> {
```
acmorrow, line 54: Are there any callers of `ResourceKey::new`?

micro-rdk/src/common/mpu6050.rs
```diff
@@ -37,7 +37,7 @@ use std::sync::{Arc, Mutex};
 
 pub(crate) fn register_models(registry: &mut ComponentRegistry) {
     if registry
-        .register_movement_sensor("gyro-mpu6050", &MPU6050::from_config)
+        .register_movement_sensor("gyro-mpu6050".to_string(), &MPU6050::from_config)
```
acmorrow, line 40: Do you still need this `to_string`

micro-rdk/src/common/registry.rs
```diff
@@ -595,12 +615,12 @@ mod tests {
...
         // register fake board
         common::board::register_models(&mut registry);
-        let ctor = registry.get_board_constructor("fake".to_string());
+        let ctor = registry.get_board_constructor("fake");
         assert!(ctor.is_ok());
 
         // register test sensor
         assert!(registry
-            .register_sensor("test_sensor", &TestSensor::from_config)
+            .register_sensor("test_sensor".to_string(), &TestSensor::from_config)
```
acmorrow, line 623: It looks to me like the `to_string`s here and below on `register_x` calls are no longer needed.

micro-rdk/src/common/registry.rs
```diff
@@ -212,138 +230,150 @@ impl ComponentRegistry {
         }
     }
     #[cfg(feature = "camera")]
-    pub fn register_camera(
+    pub fn register_camera<S: Into<String>>(
         &mut self,
-        model: &'static str,
+        model: S,
```
acmorrow, line 235: BTW, I realized you can also write these in the equivalent form:

`pub fn register_x(&mut self, model: impl Into<String>)`, if you like that better.

micro-rdk/src/common/registry.rs
```diff
@@ -212,138 +230,150 @@ impl ComponentRegistry {
...
         Ok(())
     }
 
-    pub fn register_dependency_getter(
+    pub fn register_dependency_getter<S: Into<String>>(
         &mut self,
-        component_type: &'static str,
-        model: &'static str,
+        component_type: S,
+        model: S,
```
acmorrow, line 365: This is why I think using `impl Into<String>` on the args would be a little better, because as written, `component_type` and `model` must be the same type.

micro-rdk/src/common/registry.rs
```diff
@@ -48,11 +48,11 @@ pub fn get_board_from_dependencies(deps: Vec<Dependency>) -> Option<BoardType> {
 // first element is a string representing the component type (arm, motor, etc.)
 // and the second element is its name.
 #[derive(Debug, Clone, Hash, Eq, PartialEq)]
-pub struct ResourceKey(pub &'static str, pub String);
+pub struct ResourceKey(pub String, pub String);
```
acmorrow, line 51: If the below `ResourceKey::new` is truly unused, could we introduce a new `new` that admitted both argument types:

```
impl ResourceKey {
    pub fn new(model: impl Into<String>, name: impl Into<String>) -> Self {
        Self ( model.into(), name.into() )
    }
}
```

Then you could drop the `.to_string` calls on, e.g.:

```
let r_key = ResourceKey(EncoderCompName.to_string(), enc_name);
```

micro-rdk/src/common/registry.rs
```diff
@@ -186,16 +186,34 @@ impl Default for ComponentRegistry {
 impl ComponentRegistry {
     pub fn new() -> Self {
         let mut dependency_func_map = Map::new();
-        dependency_func_map.insert(crate::common::motor::COMPONENT_NAME, Map::new());
-        dependency_func_map.insert(crate::common::movement_sensor::COMPONENT_NAME, Map::new());
-        dependency_func_map.insert(crate::common::encoder::COMPONENT_NAME, Map::new());
-        dependency_func_map.insert(crate::common::sensor::COMPONENT_NAME, Map::new());
-        dependency_func_map.insert(crate::common::base::COMPONENT_NAME, Map::new());
+        dependency_func_map.insert(crate::common::motor::COMPONENT_NAME.to_string(), Map::new());
```
acmorrow, line 189: Does `.into()` rather than `.to_string()` work for these calls?

micro-rdk/src/common/registry.rs
```diff
@@ -366,119 +396,109 @@ impl ComponentRegistry {
...
-        model: String,
+        model: &str,
     ) -> Result<&'static BoardConstructor, RegistryError> {
-        let model_name: &str = &model;
-        if let Some(ctor) = self.board.get(model_name) {
+        if let Some(ctor) = self.board.get(model) {
             return Ok(*ctor);
         }
-        Err(RegistryError::ModelNotFound(model))
+        Err(RegistryError::ModelNotFound(model.to_string()))
```
acmorrow, line 410: Does `.into()` work for these rather than `.to_string`, here and below?

acmorrow (changes requested):

mattjperez:
@acmorrow implemented your suggestions on signatures

PR #292, opened 2024-08-15
update installer

npmenard:
Make app-config and wifi-ssid argument required when flashing using write-flash or when writing the nvs partition to the binary.


acmorrow (approved):
LGTM

PR #293, opened 2024-08-19
[RSDK-8578] Revise locking for periodic app client tasks

acmorrow:
Three effects/changes here:
- No upgradable read lock is taken in periodic app client task executions, since this conflicts with the upgradable read lock in signaling. A plain read lock is used instead.
- No lock is held now across the periodic task invocation, so long running tasks (e.g. data sync across a low bandwidth channel) do not block each other / signaling.
- Errors observed by periodic app client task invocations no longer drop the app client, reverting on-error behavior to what existed prior to #275.

This is intended as a stopgap solution until we have a better design in mind.


npmenard (approved):
LGTM

PR #295, opened 2024-08-19
[RSDK-8563] handle get_app_config error case

mattjperez:


micro-rdk/src/common/entry.rs
```diff
@@ -100,7 +100,10 @@ pub async fn serve_web_inner<S>(
     let (cfg_response, cfg_received_datetime) = app_client
         .get_app_config(Some(network.get_ip()))
         .await
-        .unwrap();
+        .unwrap_or_else(|e| {
+            log::error!("failed to get valid app config: {}", e.to_string());
+            (Box::new(ConfigResponse::default()), None)
```
mattjperez, line 105: invalid config response gets handled when matching on `RobotRepresentation`, using an empty robot.
verifying the behavior.

micro-rdk/src/common/entry.rs
```diff
@@ -100,7 +100,10 @@ pub async fn serve_web_inner<S>(
     let (cfg_response, cfg_received_datetime) = app_client
         .get_app_config(Some(network.get_ip()))
         .await
-        .unwrap();
+        .unwrap_or_else(|e| {
```
mattjperez, line 103: invalid config response gets handled further down when matching on RobotRepresentation, using an empty robot.

still verifying the behavior.

I like that this gets rid of an unwrap, but there is another option to handle this within get_app_config itself, returning a default ConfigResponse and logging the error.

gvaradarajan (commented):
I'd like the rest of the team to weigh in as well, but I think I would prefer that `serve_web_inner` return a `Result` with an error in this scenario. It's only caller `serve_async` is capable of returning a result, so this feels acceptable over a direct panic

gvaradarajan (approved):
LGTM, I'm convinced

acmorrow (approved):
LGTM

npmenard:
A bit of an Anti-pattern to return an empty config on an error. What are the side effects when :
1) Robot has a cached config does it it keep the cached config and runs on it (even through multiple reboots)? or is it going to create an empty robot?
2) What are the consequences on the periodic ConfigRequest task?
3) What about a robot with no cached config?

mattjperez:
> A bit of an Anti-pattern to return an empty config on an error. What are the side effects when :
> 
>     1. Robot has a cached config does it it keep the cached config and runs on it (even through multiple reboots)? or is it going to create an empty robot?
> 
>     2. What are the consequences on the periodic ConfigRequest task?
> 
>     3. What about a robot with no cached config?

Made some changes that result in the following for those cases:
1. it will use the cached config (testing now across multiple reboots)
3. it will reset when a config has been fixed in app, restarting the device, then getting and storing the valid config.
4. A new build (no cached config) that cannot `get_app_config` for the first time will 
  a. set an empty robot
  b. give the following output
```sh
I (6587) micro_rdk::esp32::entry: Attempting to validate stored robot credentials
I (6713) wifi:<ba-add>idx:1 (ifx:0, 78:45:58:6a:f3:c1), tid:0, ssn:0, winSize:64
I (9305) micro_rdk::esp32::entry: Robot credentials validated OK
E (9804) micro_rdk::common::entry: Failed to get_app_config CloudConfig, falling back to cached config if available: grpc error code 2, message "failed to retrieve config for robot part [robot part id]: config will not be returned and machine will use cached config due to the following errors in applying fragmentOverwrites: [Fragment Overwrites: received non 2xx status code: 400 in response with msg: Fragment with id [fragment id] not found; specify the correct fragment_id in the mods field.]"
I (9842) micro_rdk::common::entry: building robot from config
E (9848) micro_rdk::common::robot: Error configuring data management: ConfigError
E (10311) micro_rdk::common::conn::server: RobotCloudConfig not available, mdns not set
I (10618) micro_rdk::common::conn::server: waiting for connection
I (11074) micro_rdk::common::conn::server: waiting for connection
//  more waiting for connection logs
```

after fixing config in app, the restart monitor reset, using the new valid config, caching it.
*However*, as the cloud config has mdns information, a new device pulling a bad config will never show as `live` in app.viam.
Therefore, the device will be in a state (with an empty robot) where it is recoverable (by fixing the config on app), but there is nothing on app to tell the user it's recoverable.

Need to check after #299 merges if the logs appear despite a new machine saying `Awaiting setup` on app.

micro-rdk/src/common/conn/server.rs
```diff
@@ -241,36 +241,32 @@ where
...
-        self.mdns
-            .add_service(
-                &cfg.fqdn.replace('.', "-"),
-                "_rpc",
-                "_tcp",
-                self.port,
-                &[("grpc", "")],
-            )
-            .map_err(|e| ServerError::Other(e.into()))?;
+        if let Some(cloud_config) = config.config.clone().unwrap_or_default().cloud {
```
npmenard, line 244: there is no need to clone here refs should be enough

npmenard: also not sure if default is the right choice here we should just be able to look if cloud_config exists and if so use a referenceto it

micro-rdk/src/common/conn/server.rs
```diff
@@ -241,36 +241,32 @@ where
...
+                .add_service(
+                    &cfg.fqdn.replace('.', "-"),
+                    "_rpc",
+                    "_tcp",
+                    self.port,
+                    &[("grpc", "")],
+                )
+                .map_err(|e| ServerError::Other(e.into()))?;
+        } else {
+            log::error!("RobotCloudConfig not available, mdns not set");
```
npmenard, line 268: no need log here

micro-rdk/src/common/entry.rs
```diff
@@ -97,10 +97,14 @@ pub async fn serve_inner<S>(
...
-    let (cfg_response, cfg_received_datetime) = app_client
-        .get_app_config(Some(network.get_ip()))
-        .await
-        .unwrap();
+    let (cfg_response, cfg_received_datetime) = {
+        let config = app_client.get_app_config(Some(network.get_ip())).await;
+        config.unwrap_or_else(|e| {
+            let config = storage.get_robot_configuration().ok();
+            log::error!("Failed to get_app_config CloudConfig, falling back to cached config if available: {e}");
+            (Box::new(ConfigResponse { config }), None)
```
npmenard, line 105: Same comment as before feels very weird to re package a cached config into a ConfigResponse. Would make more sense to return something along the lines of `Option<Box<RobotConfig>>`

micro-rdk/src/common/entry.rs
```diff
@@ -228,7 +232,7 @@ pub async fn serve_inner<S>(
         .unwrap();
 
     // Attempt to cache the config for the machine we are about to `serve`.
-    if let Err(e) = storage.store_robot_configuration(cfg_response.config.unwrap()) {
+    if let Err(e) = storage.store_robot_configuration(cfg_response.config.unwrap_or_default()) {
```
npmenard, line 235: in the absence of config we should never cache an empty config as it would be an invalid state

micro-rdk/src/common/robot.rs
```diff
@@ -273,8 +273,8 @@ impl LocalRobot {
 
         let components: Result<Vec<Option<DynamicComponentConfig>>, AttributeError> = config_resp
             .config
-            .as_ref()
-            .unwrap()
+            .clone()
+            .unwrap_or_default()
```
npmenard, line 277: same comment lets not clone unless absolutely necessary

npmenard (changes requested):
done with first apss

micro-rdk/src/common/entry.rs
mattjperez: I need to pass *something* into all the code down the line. The robot config is referenced multiple times in this function as parameters for other calls and the function itself (`serve_inner`) doesn't return a result for me to propogate upwards.

By having an `Option<Box<RobotConfig>>` that means I can have `None` (when the device fails to `get_app_config` && fails to get from storage). 

These later calls all assume a `CloudConfig` is available:
- `from_cloud_config`
- initializing `ConfigMonitor` task
- building the server

Should I refactor all those (in new tickets) to consider Options? Or something else?

gvaradarajan: I agree with this refactor, working with a `ConfigResponse` makes less sense since it only wraps a `RobotConfig`

gvaradarajan (approved):
LGTM, I think using `Option<Box<RobotConfig>>` instead of `ConfigReponse` as a follow-up is worth doing

npmenard (approved):
ok let's merge with the current design.

PR #298, opened 2024-08-22
STOPGAP MEASURE - panic when too many PeriodicAppClientTasks are hanging

gvaradarajan:
There is currently an instability on ESP32s that we believe is being caused by a request to the Viam App properly resolving but hanging on dropping the TLS context. This causes signaling to hang and blocks all incoming requests rendering the machine inaccessible and unable to move forward with any of its other tasks. The purpose of this PR is two-fold: we want to collect more data around the frequency with which these tasks hang and when all the tasks are in such a state we want to prompt a restart that will start of all the tasks anew. 

This is intended to be temporary until the investigation resolves and a proper fix is scoped out.

### DIFF
--- a/micro-rdk/src/common/conn/server.rs
+++ b/micro-rdk/src/common/conn/server.rs
@@ -395,6 +395,10 @@ where
         self.app_client_tasks
             .append(&mut robot.lock().unwrap().get_periodic_app_client_tasks());
 
+        // At most, we could have every periodic app client task running concurrently. The plus one
+        // is to account for the grpc client held in self.app_client.
+        let max_anticipated_grpc_clients = self.app_client_tasks.len() + 1;
+
         // Convert each `PeriodicAppClientTask` implementer into an async task spawned on the
         // executor, and collect them all into `_app_client_tasks` so we don't lose track of them.
         let _app_client_tasks: Vec<_> = self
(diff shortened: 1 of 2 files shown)


micro-rdk/src/common/conn/server.rs
```diff
@@ -470,6 +474,16 @@ where
...
                         .await
                         .insert(app_client);
+                } else {
+                    let current_count = urguard.as_ref().unwrap().get_grpc_client_count();
+                    log::info!(
+                        "ViamServer signaling flow has created grpc client clone {} of {}",
+                        current_count,
+                        max_anticipated_grpc_clients
+                    );
+                    if current_count > (max_anticipated_grpc_clients - 1) {
```
mattjperez, line 484: nit: might be worth just repeating the `self.app_client_tasks.len()` here. don't have to think about what the -1 when looking at this line

mattjperez (approved):
LGTM

micro-rdk/src/common/conn/server.rs
acmorrow: Or just write it as `= max...`. Or `>=` if you want to be a little paranoid.

acmorrow (approved):
LGTM

PR #301, opened 2024-08-23
[RSDK-8626] Remove http2 builder customization in GrpcClient construction

acmorrow:


micro-rdk/src/common/grpc_client.rs
```diff
@@ -238,10 +238,6 @@ impl GrpcClient {
     {
         let (http2_connection, conn) = {
             let client = hyper::client::conn::http2::Builder::new(executor.clone())
-                .initial_stream_window_size(4096)
-                .initial_connection_window_size(4096)
-                .max_concurrent_reset_streams(2)
-                .max_send_buf_size(4096)
```
mattjperez, line 244: not sure if we should leave a commented link to the defaults in the library, though updating that on every upgrade might be cumbersome.

mattjperez (approved):
LGTM

micro-rdk/src/common/grpc_client.rs
acmorrow: We will never remember to do so and it will invariably become stale and therefore misleading

gvaradarajan (approved):
LGTM

npmenard (approved):
LGTM

PR #302, opened 2024-08-27
Return empty body error when no data are present after a request has completed

npmenard:
Most of the restarts over the weekend are cause by panicking on a 0 length split_off in app_client. Still unclear why this is happening since we didn't get an HTTP2 error or a gRPC error. Will make further investigation to see if we can with the root of the problem. In the meantime this fix should avoid a spurious restart.
Furthermore this issues affects older versions so it's not an side effects of recent changes.


acmorrow (approved):
LGTM

PR #303, opened 2024-08-27
[RSDK-8544] rerun build.rs when Cargo.toml changes

mattjperez:


templates/project/build.rs
```diff
@@ -15,6 +15,7 @@ pub struct Cloud {
 }
 
 fn main() -> Result<(), &'static str> {
+    println!("cargo:rerun-if-changed=Cargo.toml");
```
acmorrow, line 18: I'm a little surprised this isn't implicit, but a quick google search does find other instances of this. Do any of our other crates need similar treatment? If not, why not?

mattjperez: I don't think the `micro-rdk-server` benefits from it at the moment because despite edits to `Cargo.toml` not triggering a rebuild, changes to the underlying code that `Cargo.toml` points to will. So it will rebuild if we add more modular examples.

Additionally, we already include the `esp-idf-sys` metadata in `micro-rdk-server/Cargo.toml` by default, even though we don't have the `camera` feature enabled (as it will change the statically chosen buffer size for webrtc and grpc, not because we don't want to have it in by default).

TLDR: I don't think so, but we can always add it as-needed

acmorrow: I did some experiments, and changing `Cargo.toml` *does* usually cause a rebuild. What is needed here though is to teach it that because we are fishing data out of `Cargo.toml` ourselves as part of `build.rs`, that it *also* needs to rebuild/rerun build.rs as well. So, I think I agree, just here.

acmorrow (approved):
LGTM

PR #304, opened 2024-08-29
[RSDK-8259] Enable mDNS and http2 on esp32 builds

acmorrow:
This is prerequisite work towards support for local signaling. Basically, lighting up mDNS and HTTP2 with TLS on esp32.

- Switches esp32 builds from using `NoMdns` to `Esp32Mdns` in regular service flow.
- Binds the mDNS implementation into `ViamServer` in order to extend its lifetime.
- Adds `with_http2` to the esp32 startup flow in `serve_inner`, enabling direct gRPC/HTTP2 service
- Extends the lifetime of `Esp32TLSServerConfig` in order to prevent a UAF which otherwise prevented esp32 builds from offering a functional TLS server infrastructure.
- Fixes an issue in native TLS where `rusttls_pemfile::rsa_private_keys` was not being called, which prevented native builds from offering a functional TLS server infra.

I tested that with these changes in place I'm able to have a go SDK client make a direct gRPC/HTTP2 connection to both a native and an esp32 build of the micro-rdk.

The go client code looks like:

```
		esp32, _ := client.New(
			context.Background(),
			"acm-esp32-default-example-main.........",
			logger,
			client.WithDialOptions(rpc.WithEntityCredentials(
				"xxxxxx",
				rpc.Credentials{
					Type:    rpc.CredentialsTypeAPIKey,
					Payload: "xxxxx",
				}),
				rpc.WithWebRTCOptions(rpc.DialWebRTCOptions{Disable: true}),
			))
		if esp32 != nil {
			defer esp32.Close(context.Background())
			logger.Info("ESP32 Resources:", esp32.ResourceNames())
		}
```

Note that WebRTC has been explicitly disabled here, so the client has no choice but to go direct via gRPC/HTTP2.

This work is pretty rough, I'll leave some comments on things that need to be done.


micro-rdk/src/common/entry.rs
```diff
@@ -205,20 +204,36 @@ pub async fn serve_inner<S>(
         restart,
     )));
 
-    #[cfg(feature = "native")]
     let server_builder = {
+        let http2_port = 12346;
```
acmorrow, line 208: This is just to get rid of naming the port twice as we did before, by using a slightly different way of constructing a`SocketAddr `.

micro-rdk/src/common/entry.rs
```diff
@@ -205,20 +204,36 @@ pub async fn serve_inner<S>(
...
+                }
+                #[cfg(feature = "esp32")]
+                {
+                    let tls_server_config = Esp32TLSServerConfig::new(
+                        CString::new(certs.tls_certificate)
+                            .unwrap()
+                            .into_bytes_with_nul(),
+                        CString::new(certs.tls_private_key)
+                            .unwrap()
+                            .into_bytes_with_nul(),
```
acmorrow, line 230: There's probably a more graceful way to say this, but see below because the whole thing may need to be rethought: converting to null-terminated CString here may be a waste of time. It might be nice because then we could have the same form of args as for `NativeTlsServer` and simplify the blocks a bit.

micro-rdk/src/esp32/tls.rs
```diff
@@ -42,6 +42,8 @@ pub struct Esp32TLS {
     #[allow(dead_code)]
     alpn_ptr: Vec<*const c_char>,
     tls_cfg: Either<Box<esp_tls_cfg_server>, Box<esp_tls_cfg>>,
+    #[allow(dead_code)]
+    cfg: Option<Box<Esp32TLSServerConfig>>,
```
acmorrow, line 46: We need to keep the `Esp32TLSServerConfig` around, because it holds the memory with the cert/key material. Again, there is probably a better way to do this, but getting it onto the heap with `Box` was sufficient to prevent it getting dropped and/or moved as it was before.

micro-rdk/src/esp32/tls.rs
```diff
@@ -91,22 +93,18 @@ pub struct Esp32TLSStream {
     socket: Async<TcpStream>, // may store the raw socket
 }
 
+#[derive(Clone)]
 pub struct Esp32TLSServerConfig {
-    srv_cert: Vec<u8>,
-    srv_key: *const u8,
-    srv_key_len: u32,
+    pub srv_cert: Vec<u8>,
```
acmorrow, line 98: This is now identical to `NativeTLSServerConfig` so it should probably be commoned out.

micro-rdk/src/esp32/tls.rs
```diff
@@ -198,10 +198,10 @@ impl Esp32TLS {
                 servercert_bytes: cfg.srv_cert.len() as u32,
             },
             __bindgen_anon_5: crate::esp32::esp_idf_svc::sys::esp_tls_cfg_server__bindgen_ty_5 {
-                serverkey_buf: cfg.srv_key,
+                serverkey_buf: cfg.srv_key.as_ptr(),
```
acmorrow, line 201: Above on 195 that `as_ptr` created, I believe, a use-after-free. Nothing is holding alive the state in `Esp32TLSServerConfig` such that these pointers are valid once we get into `serve`.

micro-rdk/src/esp32/tls.rs
```diff
@@ -176,11 +174,13 @@ impl Esp32TLS {
...
             alpn_ptr,
             tls_cfg: Either::Right(tls_cfg_client),
+            cfg: None,
         }
     }
     /// Creates a TLS object ready to accept connection or connect to a server
-    pub fn new_server(cfg: &Esp32TLSServerConfig) -> Self {
+    pub fn new_server(cfg: Esp32TLSServerConfig) -> Self {
         let mut alpn_ptr: Vec<_> = vec![ALPN_PROTOCOLS.as_ptr() as *const i8, std::ptr::null()];
+        let cfg = Box::new(cfg);
```
acmorrow, line 183: See below. There is probably a better way to do this, maybe with `Pin`ing?

micro-rdk/src/native/tls.rs
```diff
@@ -78,14 +78,20 @@ impl NativeTlsStream {
                     .iter()
                     .map(|c| rustls::Certificate(c.clone()))
                     .collect();
+            let priv_keys: Vec<_> =
```
acmorrow, line 81: Native TLS didn't work server side because we were feeding the literal text of the private key from the pem stuff in, rather than the delimited bytes. Using `rsa_private_keys` does that for us.

micro-rdk/src/native/tls.rs
```diff
@@ -78,14 +78,20 @@ impl NativeTlsStream {
...
+                    .collect();
 
             let mut cfg = ServerConfig::builder()
                 .with_safe_default_cipher_suites()
                 .with_safe_default_kx_groups()
                 .with_protocol_versions(&[&rustls::version::TLS12])
                 .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?
                 .with_no_client_auth()
-                .with_single_cert(cert_chain, rustls::PrivateKey(tls_cfg.srv_key.clone()))
+                .with_single_cert(cert_chain, priv_keys[0].clone())
```
acmorrow, line 94: There should probably be a check that we have one, at least. What should we be doing if we have multiple. Is there a `with_multiple_cert`? I haven't seen such a thing.

micro-rdk/src/common/conn/server.rs
```diff
@@ -385,6 +389,7 @@ where
             app_client_tasks,
             network,
             rpc_host,
+            _mdns: mdns,
```
acmorrow, line 392: If we lose the network, we may need to re-create this since we may come up with a different IP. Maybe that happens organically as part of refactoring the relationship between networking/robot/provisioning/server? I'm not inclined to tackle this directly in this review because I suspect there are other places where we have this problem as well. Should I file a ticket so we don't lose track of the need to handle that better?

micro-rdk/src/common/entry.rs
```diff
@@ -205,20 +204,36 @@ pub async fn serve_inner<S>(
...
             {
                 let certs = app_client.get_certificates().await.unwrap();
-                let tls_server_config = NativeTlsServerConfig::new(
-                    certs.tls_certificate.as_bytes().to_vec(),
-                    certs.tls_private_key.as_bytes().to_vec(),
-                );
-                let address: SocketAddr = "0.0.0.0:12346".parse().unwrap();
-                let tls = Box::new(NativeTls::new_server(tls_server_config));
-                NativeListener::new(address.into(), Some(tls)).unwrap()
+                let address = SocketAddr::new("0.0.0.0".parse().unwrap(), http2_port);
```
acmorrow, line 212: ⬆️ - Presumably, we need to add support for caching the certs we get back from app just like we cache the robot config. Otherwise, there is no way to bring up the servers in the network partition case. I'll file a ticket for this.

micro-rdk/src/common/conn/server.rs
```diff
@@ -349,9 +351,11 @@ pub struct ViamServer<C, T, CC, D, L, NetworkType> {
     app_client_tasks: Vec<Box<dyn PeriodicAppClientTask>>,
     network: NetworkType,
     rpc_host: String,
+    _mdns: M,
```
mattjperez, line 354: why `_`?

micro-rdk/src/common/conn/server.rs
mattjperez: I'd like to track it. We can maybe put network, mdns, and whatever else might be coupled to it in a single struct and simplify these arguments.

micro-rdk/src/esp32/tls.rs
mattjperez: I think you're right about `Pin`ing. With Box there's no guarantee that the underlying data doesn't get manipulated/moved-out/invalidated/etc. As long as it's not a `PhantomPin` it should be cleaned up once dropped out of scope.

micro-rdk/src/native/tls.rs
mattjperez: [`with_single_cert`](https://docs.rs/rustls/latest/rustls/struct.ConfigBuilder.html#method.with_single_cert) actually takes a `Vec<CertificateDer<'static>>` (DER encoded certificates) as the first argument, `cert_chain`.

If you mean the ability to dynamically respond using specific cert-key combos, there's [`Acceptor`](https://docs.rs/rustls/latest/rustls/server/struct.Acceptor.html) for async and [`ResolvesServerCert`](https://docs.rs/rustls/latest/rustls/server/trait.ResolvesServerCert.html#) for blocking.

PR #306, opened 2024-09-04
[RSDK-8384] implement RobotService/GetVersion API

gvaradarajan:
Some not-so-relevant auto-generation. The relevant files are

- Makefile
- micro-rdk/src/common/grpc.rs
- micro-rdk/src/gen/api_version.rs
- micro-rdk/src/lib.rs

From this commit forward, the preceding for incrementing the version of the Viam Proto API used by micro-RDK is to change the hardcoded `VIAM_API_VERSION` variable in Makefile and then run `make buf`. This alters micro-rdk/src/gen/api_version.rs, which is then referenced by `GrpcServer::get_version`


mattjperez (approved):
this hackery LGTM

PR #307, opened 2024-09-04
[RSDK-8668] camera i2c board

mattjperez:
still testing to make sure behavior is right but wanted to start gathering feedback

### DIFF
--- a/micro-rdk/src/common/i2c.rs
+++ b/micro-rdk/src/common/i2c.rs
@@ -25,7 +25,7 @@ pub enum I2CErrors {
 // embedded_hal I2C trait when supporting boards beyond ESP32.
 pub trait I2CHandle {
     fn name(&self) -> String;
-
+    fn bus_no(&self) -> u8;
     fn read_i2c(&mut self, _address: u8, _buffer: &mut [u8]) -> Result<(), I2CErrors> {
         Err(I2CErrors::I2CUnimplemented("read_i2c"))
     }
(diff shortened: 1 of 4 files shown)


micro-rdk/src/esp32/mperez@chaos.44516:1725459714
gvaradarajan, line 1: lol wut?

micro-rdk/src/common/i2c.rs
```diff
@@ -25,7 +25,7 @@ pub enum I2CErrors {
 // embedded_hal I2C trait when supporting boards beyond ESP32.
 pub trait I2CHandle {
     fn name(&self) -> String;
-
+    fn bus_no(&self) -> u8;
```
gvaradarajan, line 28: I think I'd prefer either just `bus` or `bus_number`. I also think this should not return a number, but instead an `&str` or a `String`. The reason is that, while most processors will use a numbering scheme for buses, that scheme is not universally enforced. A `String` can most flexibly accommodate different conventions and I think it's ok for board-specific drivers to rely on numbering schemes defined by that board's implementation

gvaradarajan (changes requested):
Not sure about the universality of bus numbering

micro-rdk/src/esp32/mperez@chaos.44516:1725459714
mattjperez: wut wut

micro-rdk/src/common/i2c.rs
mattjperez: that works. @stevebriskin and I were talking about maybe removing the `name` portion from an i2c config as it's just an alias and is now an unused remnant, so this would replace that.

mattjperez:
closing until decision made on removing `name` from i2c attributes

PR #309, opened 2024-09-06
[RSDK-8708] add coredump reader

npmenard:


micro-rdk/src/esp32/coredump.rs
```diff
@@ -0,0 +1,196 @@
...
+                let offset = ((val.floor() as usize) * CORE_FRAGMENT_SIZE).min(self.len);
+                offset
+                    .lt(&self.len)
+                    .then_some(Null)
+                    .ok_or(GenericError::Other(
+                        "chunk requested outside of bounds".into(),
+                    ))?;
+
+                let to_read: usize = CORE_FRAGMENT_SIZE.min(self.len - offset);
+                let mut buf = BytesMut::with_capacity(CORE_FRAGMENT_SIZE);
```
mattjperez, line 160: I'm not sure if I'm reading this right. is this 0-indexed?

`offset`: the smaller of `the starting byte of the n'th chunk` or the `len` of coredump 

if `offset < self.len`, then turn to `Option<Null>`, converted to a Result<Null, Err>, mapping on the Err as out of bounds.

Wouldn't you want L152 (after getting offset) to compare `(offset + CORE_FRAGMENT_SIZE).lt(&self.len)` to ensure that the end of the nth chunk is within boundaries, not just its starting point?

mattjperez (changes requested):

micro-rdk/src/esp32/coredump.rs
```diff
@@ -0,0 +1,196 @@
...
+            fields: HashMap::new(),
+        }))
+    }
+}
+
+struct Null;
+
+const CORE_FRAGMENT_SIZE: usize = 3072;
+
+impl DoCommand for Coredump {
```
gvaradarajan, line 105: I think it's worth having a comment outlining the three commands that can be sent, their expected values, and the high-level expected behavior

micro-rdk/src/esp32/coredump.rs
```diff
@@ -0,0 +1,196 @@
...
+            ))?
+        };
+        // the first two fields of the coredump are length (uint32_t) and version (uint32_t)
+        let mut first_bytes = [0xFF_u8; 8];
+        esp_idf_svc::sys::esp!(unsafe {
+            esp_partition_read(partition_iterator, 0, first_bytes.as_mut_ptr() as *mut _, 8)
+        })?;
+
+        if first_bytes[7] != 0xFF {
+            len = usize::from_le_bytes(first_bytes[0..4].try_into().unwrap()); // safe because we would have read 4 bytes
```
gvaradarajan, line 64: just to make sure I understand: we only need to do this once because even if someone erases the coredump via DoCommand, a coredump can only be written in a situation that what prompt a restart and result in reconfiguration? Is this correct? Or is there a situation in which something can be written to the coredump and a restart does not occur (meaning this sensor's state won't be properly reset)?

micro-rdk/src/esp32/coredump.rs
npmenard: yes it's 0 indexed

offset is the lower bound, to_read+offset will give you the upper bound. Because we will read [offset, offset + min(CORE_FRAGMENT_SIZE, self.len -offset))

micro-rdk/src/esp32/coredump.rs
npmenard: make sense

micro-rdk/src/esp32/coredump.rs
npmenard: coredump is only written on panic/abort whether from rust or C. So during the lifetime of the program the coredump exists until we erase it

gvaradarajan (approved):
LGTM

mattjperez (approved):
LGTM,
the embedded `DoCommand` docs would be great

PR #310, opened 2024-09-11
[RSDK-8673] respect sync_disabled/capture_disabled flags in data config

gvaradarajan:
Fixes an issue where, when the user disables the "Capturing" and/or "Syncing" flag in the data management service config, the corresponding functionality is not turned off


npmenard (approved):
LGTM

PR #312, opened 2024-09-13
[RSDK-8757] allow disabling of individual capture configs

gvaradarajan:
In app, you can "turn off" individual data capture methods, but the micro-RDK infrastructure does not respect this setting. This fixes that.

### DIFF
--- a/micro-rdk/src/common/data_collector.rs
+++ b/micro-rdk/src/common/data_collector.rs
@@ -25,11 +25,19 @@ pub struct DataCollectorConfig {
     pub method: CollectionMethod,
     pub capture_frequency_hz: f32,
     pub capacity: usize,
+    pub disabled: bool,
 }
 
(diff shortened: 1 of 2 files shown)


micro-rdk/src/common/data_collector.rs
```diff
@@ -45,7 +53,7 @@ impl TryFrom<&Kind> for DataCollectorConfig {
             .unwrap_or(&Kind::NumberValue(DEFAULT_CACHE_SIZE_KB))
             .try_into()?;
         let capacity = (capacity_kb * 1000.0) as usize;
-        if capacity < 1000 {
+        if (capacity < 1000) && !disabled {
```
mattjperez, line 56: nit: not actually a change from this PR but wouldn't this (without the casting to usize) be the same as `if capacity_kb < 1`?

mattjperez (approved):
LGTM

PR #313, opened 2024-09-16
[RSDK-8786] avoid code duplication in rpc handling path

npmenard:
Having a template parameter in the path handling requests would trigger code duplication when both http2 & webrtc path are enable. By separating the handling path in a struct without generics we can save on code duplication
See details in jira ticket

### DIFF
--- a/micro-rdk/src/common/generic.rs
+++ b/micro-rdk/src/common/generic.rs
@@ -22,6 +22,8 @@ pub static COMPONENT_NAME: &str = "generic";
 pub enum GenericError {
     #[error("Generic: method {0} unimplemented")]
     MethodUnimplemented(&'static str),
+    #[error("Generic other error: {0}")]
+    Other(Box<dyn std::error::Error + Send + Sync>),
 }
 #[cfg(feature = "builtin-components")]
 pub(crate) fn register_models(registry: &mut ComponentRegistry) {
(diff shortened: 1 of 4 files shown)


micro-rdk/src/common/grpc.rs
```diff
@@ -1239,7 +1279,7 @@ where
     }
 
     #[cfg(feature = "camera")]
-    fn camera_get_image(&mut self, message: &[u8]) -> Result<(), ServerError> {
+    fn camera_get_image(&mut self, message: &[u8]) -> Result<Bytes, ServerError> {
```
mattjperez, line 1282: Does this compile? The implementations still return an `Ok(())`.
The camera rpc handlers don't call `encode` directly to reduce copying though that might be unavoidable.

micro-rdk/src/common/grpc.rs
```diff
@@ -1270,7 +1310,7 @@ where
     }
 
     #[cfg(feature = "camera")]
-    fn camera_render_frame(&mut self, message: &[u8]) -> Result<(), ServerError> {
+    fn camera_render_frame(&mut self, message: &[u8]) -> Result<Bytes, ServerError> {
```
mattjperez, line 1313: same here

mattjperez (changes requested):

micro-rdk/src/common/grpc.rs
npmenard: now yes :P i am returning the encoded Bytes does it looks good to you?

mattjperez (approved):
LGTM, confirmed FakeCam that things work as expected

PR #314, opened 2024-09-16
[RSDK-8787] remove assoc type in wifi manager trait

npmenard:
remove the associated type in wifimanager trait so that merging entrypoints will be easier.

### DIFF
--- a/micro-rdk/src/common/provisioning/server.rs
+++ b/micro-rdk/src/common/provisioning/server.rs
@@ -34,6 +34,7 @@ use hyper::{
     Response,
 };
 use prost::Message;
+use thiserror::Error;
 
 async fn dns_server(ap_ip: Ipv4Addr) {
     let socket = async_io::Async::<UdpSocket>::bind(([0, 0, 0, 0], 53)).unwrap();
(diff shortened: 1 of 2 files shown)


micro-rdk/src/common/provisioning/server.rs
```diff
@@ -445,20 +446,33 @@ where
     }
 }
 
+#[derive(Error, Debug)]
+pub enum WifiManagerError {
+    #[error("cannot assign to heapless string")]
+    HeaplessStringError,
```
mattjperez, line 452: are we sure it will always be due to a `heapless` error? should we not propagate the underlying message somehow like in `OtherError`

mattjperez (approved):
LGTM, just a non-blocking question

micro-rdk/src/common/provisioning/server.rs
gvaradarajan: This error is only returned when essentially using the `TryFrom<&str>` trait for `heapless::String`. The corresponding `Error` type for that trait implementation is `()`, so the lack of an underlying message feels all but guaranteed

gvaradarajan (approved):
LGTM

PR #316, opened 2024-09-17
[RSDK-8596] Revamp Entrypoints

npmenard:
### Disclaimer
This is a work in progress many things are missing but it's now at a stage where I think most of the architecture is there and we are mostly left with coding the logic of the different states. There is still a lot of work to do but now is a good time wonder if the approach is sensible.

### The PR

This PR attempt to address the tech debt we have been accumulating over the past months. When ViamServer was first written we didn't take into account adding extra tasks, dealing with provisioning, having support for offline & online mode etc ... overtime adding these features increased complexity and made interaction between the different components or states harder to track and debug. 

#### `ViamServerBuilder`
The new approach centers around a monolithic ViamServerBuilder that users will configure with appropriate options we are also reducing the overall number of generics and replacing them with `Box<dyn traits>`. There is a penalty to that since now we have fat pointers and we pay some indirection price. I think that it is okay, perhaps in the future we will be able to return to some generic parametrization. 
The burden of configuring ViamServer will be passed on the binary.
##### `RobotServer`
The `RobotServer` struct will be dealing with serving incoming http2 connections and webrtc requests. Since it now mostly take references to a `ViamServer` it will not spawn it's listeners futures onto an executor. I want to introduce a strong correlation between the `ViamServer` and its components. It doesn't require and `AppClient` to fucntion so we should be able to instantiate it  offline
##### `AppClient` and Periodic Tasks
The `AppClient` & it's periodic tasks will be disjoint for the `RobotServer` with the exception of a shared channel for passing signaling requests. 
Each tasks will be converted to an `AppClientTaskRunner` which holds references to the task and an `AppClient`. Because we use references we can't spawn each tasks on the executor as such they are now collected into a `FutureUnordered` collection which is awaited along with the `RobotServer` The following assumptions are made:

1. Because each task is periodic they never stop running
2. HTTP2 protocol or Network errors are unrecoverable and the AppClient should be considered invalid
3. Grpc Errors should be swallowed by the tasks with the exceptions of permission related errors

If any tasks returns an error the collection should be dropped and an new attempt to create an `AppClient` and spawn the relevant tasks should be made.

### Downsides

- Not spawning task on an executor might increase the stack usage on the esp32
- The new approach significantly differs for what we have done so far so things may break or not work as expected.
- Unclear how well we can deal with loosing and or gaining internet connection 
- May not solve the AppClient stuck on esp32

### ToDos (non exhaustive)

- [x] Support for esp32 (migrating traits & timestamp setting)
- [x] Add service removal functions in Mdns traits
- [x] Maybe define a set of default tasks( restart monitor, config etc...)
- [x] Define a state machine for transitioning between state (explicit or implicit)
- [x] Test provisioning
- [x] Add `IncomingConnectionManager`
- [x] Where does rpc_host lives?
- [x] Expand functionalities of the `AppServerInsecure` for testing most of the case scenario
- [x] Name stuff more sensibly
- [x] Remove duplicated code
- [x] Final cleanup


micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,756 @@
...
+            rpc_host: self.rpc_host.clone(),
+        }));
+        let mut tasks: FuturesUnordered<_> = FuturesUnordered::new();
+        tasks.push(Either::Right(self.run_app_client_tasks(Some(app_client))));
+        tasks.push(Either::Left(inner.run()));
+        while let Some(_) = tasks.next().await {}
+    }
+    async fn connect_to_app(&self) -> Result<AppClient, AppClientError> {
+        //ugly hack to remove last /
+        //needs to change that but would need to update GrpcClient
```
gvaradarajan, line 320: Tbh, the PR is pretty big anyway, might as well make this change as well?

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,756 @@
...
+        for task in &self.app_client_tasks {
+            app_client_tasks.push(AppClientTaskRunner {
+                app_client: &app_client,
+                invoker: task,
+                timer: None,
+                task: None,
+            });
+        }
+        while let Some(res) = app_client_tasks.next().await {
+            res.unwrap();
```
gvaradarajan, line 356: It's more verbose, but I think it would be clearer to do an `if res.is_err() { panic!(...); }`. I think unwrap obfuscates the fact that when an app client task fails we want to cause a restart.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,756 @@
...
+    timer: Option<async_io::Timer>,
+}
+
+impl<'a> Future for AppClientTaskRunner<'a> {
+    type Output = Result<(), AppClientError>;
+    fn poll(
+        mut self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Self::Output> {
+        if let Some(&mut ref mut timer) = self.timer.as_mut() {
```
gvaradarajan, line 530: Could we make use of an enum to simplify this? We can set task as an instance of the enum and switch whether the enum is the timer or the actual task based on what the task was previously and whether it returns `Poll::Ready` or not

micro-rdk/src/common/data_manager.rs
```diff
@@ -361,7 +415,7 @@ where
         Ok(msg)
     }
 
-    async fn run<'b>(&mut self, app_client: &'b AppClient) -> Result<(), AppClientError> {
+    async fn run<'b>(&self, app_client: &'b AppClient) -> Result<(), AppClientError> {
```
gvaradarajan, line 418: asking for curiosity: does `self` not need to be a mutable reference to get the store lock below? Was that using some sort of interior mutability this whole time?

gvaradarajan (commented):
first pass

micro-rdk/src/common/conn/viam.rs
npmenard: I updated the code hopefully is a bit easier to follow through

micro-rdk/src/common/data_manager.rs
npmenard: you can usually get a mutable reference from an &self after locking. I haven't tried the data upload yet so there might be a potential issue here indeed

micro-rdk/src/common/conn/viam.rs
npmenard: yep but technically non-blocker and might be a ticket for later if we wanted too

micro-rdk/src/common/conn/viam.rs
npmenard: This where we would implement the code handling the restarting of  app_client and so on. I haven't added the logic around this yet

gvaradarajan: ah ok, makes sense

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,1281 @@
...
+        let app_client = self
+            .connect_to_app()
+            .await
+            .inspect_err(|error| {
+                if error.is_permission_denied() || error.is_unauthenticated() {
+                    let _ = self.storage.reset_robot_credentials().inspect_err(|err| {
+                        log::error!("error {:?} while erasing credentials", err)
+                    });
+                    let _ = self.storage.reset_robot_configuration().inspect_err(|err| {
+                        log::error!("error {:?} while erasing credentials", err)
```
npmenard, line 317: erase robot_configuration

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,1281 @@
...
+                    });
+                    panic!("erased credentials restart robot"); // TODO bubble up error and go back in provisioning
+                }
+                log::error!("couldn't connect to {} reason {:?}", self.app_uri, error);
+            })
+            .ok();
+
+        // The next step is to build the robot based on the config retrieved online, from storage or as an empty robot.
```
npmenard, line 325: poorly worded

acmorrow (commented):
Here are the accumulated comments from when @mattjperez, @gvaradarajan, and I took a look through this change. We didn't get to everything, but wanted to share small bits of feedback and thoughts that we had accumulated so far.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,756 @@
...
+
+pub struct ViamServerBuilder2<Storage> {
+    storage: Storage,
+    http2_server: HTTP2Server,
+    webrtc_configuration: WebRtcListener,
+    provisioning_info: ProvisioningInfo,
+    wifi_manager: Option<Box<dyn WifiManager>>,
+    component_registry: Box<ComponentRegistry>,
+    http2_server_port: u16,
+    http2_server_insecure: bool,
```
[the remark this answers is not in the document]
npmenard: Maybe we shoukld add a ticket for this. This is also somewhat connected to storing certs in storage

micro-rdk-server/native/main.rs
```diff
@@ -44,13 +56,33 @@ mod native {
...
         }
 
-        let max_connections = 3;
-
         let mut info = ProvisioningInfo::default();
         info.set_manufacturer("viam".to_owned());
         info.set_model("test-esp32".to_owned());
 
-        serve_with_network(Some(info), repr, max_connections, storage, network);
+        let webrtc_certs = Rc::new(Box::new(WebRtcCertificate::new()) as Box<dyn Certificate>);
```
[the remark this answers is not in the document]
npmenard: I don't feel strongly so I'll defer to you opinion

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,756 @@
...
+        ViamServerBuilder2 {
+            storage,
+            http2_server: HTTP2Server::Empty,
+            webrtc_configuration: WebRtcListener::Empty,
+            provisioning_info: Default::default(),
+            wifi_manager: None,
+            component_registry: Default::default(),
+            http2_server_port: 12346,
+            http2_server_insecure: true,
+            initial_app_uri: "https://app.viam.com:443".try_into().unwrap(),
```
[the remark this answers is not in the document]
npmenard: let's make a ticket. This should be the last place where there is a reference

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,756 @@
...
+            app_client_tasks: Default::default(),
+            max_concurrent_connections: 1,
+        }
+    }
```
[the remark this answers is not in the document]
npmenard: agreed

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,756 @@
...
+    <Storage as RobotConfigurationStorage>::Error: Debug,
+    ServerError: From<<Storage as RobotConfigurationStorage>::Error>,
+    C: ViamH2Connector + 'static,
+    M: Mdns,
+{
+    pub fn run_forever(&mut self) {
+        let exec = self.executor.clone();
+        exec.block_on(self.run());
+    }
+    async fn run(&mut self) {
```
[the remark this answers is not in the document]
npmenard: yes

micro-rdk/Cargo.toml
```diff
@@ -29,6 +29,7 @@ esp-idf-logs = ["esp32"]
 [dev-dependencies]
 test-log.workspace = true
 env_logger.workspace = true
+rustls = { workspace = true, features = ["dangerous_configuration"] }
```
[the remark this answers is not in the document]
npmenard: for testing tlse server connection with self signed certs

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,1723 @@
...
+        // storage. If not, we should go straight to provisioning.
+        //
+        // truth table for 2nd check
+        // +------------------+------------------+------------------+
+        //| Has WifiManager  | Has Credential   | Should provision |
+        //+------------------+------------------+------------------+
+        //| false            | x                | false            |
+        //| true             | false            | true             |
+        //| true             | true             | false            |
+        //+------------------+------------------+------------------+
```
mattjperez, line 394: Finally! :sob: 
This is gonna make things way more straightforward!

micro-rdk/src/native/tcp.rs
```diff
@@ -1,116 +1,145 @@
...
+            self.0.take().unwrap(),
+        ))))
     }
 }
 
 /// Enum to represent a TCP stream (either plain or encrypted)
 pub enum NativeStream {
     LocalPlain(Async<TcpStream>),
-    TLSStream(Box<NativeTlsStream>),
+    NewTlsStream(futures_rustls::TlsStream<Async<TcpStream>>),
```
mattjperez, line 142: just `TlsStream`?

micro-rdk/src/esp32/tcp.rs
```diff
@@ -1,129 +1,519 @@
...
+}
+
+impl Esp32ClientConfig {
+    fn new() -> Self {
+        let mut alpn_proto = vec![ALPN_PROTOCOLS.as_ptr() as *const i8, std::ptr::null()];
+        // this is a root certificate to validate the server's certificate
+        let cert = include_bytes!("../../certs/google_gts_root_r1.crt");
+
+        let cfg = Box::new(esp_tls_cfg {
+            alpn_protos: alpn_proto.as_mut_ptr(),
```
mattjperez, line 120: does this need to be pinned or copied? I think the bug @acmorrow came across when adding http2 support was related to something like this or like what's below.

micro-rdk/src/esp32/tcp.rs
```diff
@@ -1,129 +1,519 @@
...
+use super::dtls::{AsyncSSLStream, SSLContext, SSLError};
+
+extern "C" {
+    fn esp_create_mbedtls_handle(
+        hostname: *const c_char,
+        hostlen: i32,
+        cfg: *const c_void,
+        tls: *mut esp_tls_t,
+    ) -> i32;
+}
```
mattjperez, line 32: is this not importable like the other esp-idf functions?

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,1723 @@
...
+            Box::new(network),
+        );
+        let cloned_exec = exec.clone();
+
+        let mut app = AppServerInsecure::default();
+        app.config_fn = Some(Rc::new(Box::new(|| {
+            let mut cfg = make_sample_config();
+            if let Some(cloud) = cfg.cloud.as_mut() {
+                cloud.fqdn = "test-bot.xxds65ui.viam.cloud".to_owned();
+                cloud.local_fqdn = "test-bot.xxds65ui.viam.local.cloud".to_owned();
```
mattjperez, line 1376: can we add a TODO to have these test addresses centralized or come a dev config?

mattjperez (commented):
Just some questions and comments.
This is a giant PR and I'd prefer to get it merged quickly so we can start testing live and fix any issues.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,1723 @@
...
+                .as_ref()
+                .as_ref()
+                .is_some_and(|_| !self.storage.has_wifi_credentials())
+        {
+            self.provision().await;
+        }
+
+        // Since provisioning was run and completed, credentials are properly populated
+        // if wifi manager is configured loop forever until wifi is connected
+        if let Some(wifi) = self.wifi_manager.as_ref().as_ref() {
```
acmorrow, line 407: Should it be possible to create the robot even if the local wifi is unreachable? You still could capture data. If you find later that the robot is no longer valid once you connect to wifi and then app you would destroy it.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -0,0 +1,1723 @@
...
+        //| Has WifiManager  | Has Credential   | Should provision |
+        //+------------------+------------------+------------------+
+        //| false            | x                | false            |
+        //| true             | false            | true             |
+        //| true             | true             | false            |
+        //+------------------+------------------+------------------+
+        if !self.storage.has_robot_credentials()
+            || self
+                .wifi_manager
+                .as_ref()
```
gvaradarajan, line 398: there's a lot of `as_ref().as_ref()` every time we reference the wifi_manager. Would this be cleaner if it was an `Option<Rc<>>` rather than an `Rc<Option<>>`? Then we could implement `WifiManager` for `Rc<Box<dyn WifiManager>>`?

micro-rdk-ffi/src/ffi/runtime.rs
```diff
@@ -258,36 +259,79 @@ pub unsafe extern "C" fn viam_server_start(ctx: *mut viam_server_context) -> via
...
+        builder.build(
+            NativeH2Connector::default(),
+            Executor::new(),
+            NativeMdns::new("".to_owned(), network.get_ip()).unwrap(),
+            Box::new(network),
         )
-    }
+    };
 
-    viam_code::VIAM_OK
```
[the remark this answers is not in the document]
npmenard: no it may still return an error

micro-rdk/src/esp32/tcp.rs
acmorrow: Not exposed, we are hijacking it.

micro-rdk/src/native/tcp.rs
acmorrow: Will rename

PR #321, opened 2024-10-09
 RSDK-8896 Request and accept heartbeats from signaling server 

benjirewis:
RSDK-8896

Updates proto. Includes `"heartbeats-allowed": "true"` metadata in the `Answer` call that initiates signaling. This will tell the signaling server to start sending `answer_request::Stage::Heartbeat`s every 15s. Catches these heartbeats in `wait_sdp_offer` and `next_remote_candidate`, logging and throwing away (no need to respond to the heartbeats).

Tested with Gautham


micro-rdk/src/common/grpc_client.rs
```diff
@@ -280,6 +281,9 @@ impl GrpcClient {
         if let Some(jwt) = jwt {
             r = r.header("authorization", jwt);
         };
+        if let Some(heartbeats_allowed) = heartbeats_allowed {
+            r = r.header("hearbeats_allowed", heartbeats_allowed);
```
gvaradarajan, line 285: "heartbeats" is misspelled here, would this still work?

micro-rdk/src/common/webrtc/api.rs
```diff
@@ -283,40 +287,51 @@ impl WebRtcSignalingChannel {
...
+                            answer_request::Stage::Done(_) => {
                                 return Ok(None);
                             }
-                        }
-                        answer_request::Stage::Error(s) => {
-                            if let Some(status) = s.status {
-                                return Err(WebRtcError::SignalingError(status.message));
+                            answer_request::Stage::Heartbeat(_) => {
+                                log::debug!("received a heartbeat from the signaling server");
+                                continue;
```
gvaradarajan, line 324: do we want this to potentially continue indefinitely? If not, I think we would want some max_iterations

micro-rdk/src/common/webrtc/api.rs
```diff
@@ -283,40 +287,51 @@ impl WebRtcSignalingChannel {
...
-                            return Err(WebRtcError::SignalingError("unknown".to_owned()));
-                        }
-                        answer_request::Stage::Done(_) => {
-                            return Ok(None);
-                        }
-                        _ => {
-                            return Err(WebRtcError::SignalingError("unexpected stage".to_owned()))
                         }
                     }
+                    return Ok(None);
```
gvaradarajan, line 333: not sure if you were aware (and you don't have to make this change), but you can break from loops with return values by declaring `break Ok(None)` for example

gvaradarajan (commented):
I want to pull this PR down and test this out before I approve, but had a couple comments

micro-rdk/src/common/webrtc/api.rs
benjirewis: I don't know about continue _indefinitely_ but I don't _think_ we want a max iteration here. My understanding is that this `loop` will receive the next `AnswerRequest` for the current signaling exchange. If it's a heartbeat, it should try to call `next()` on the `signaling_rx` again. If the signaling server has been disconnected, we should be able to exit the loop via L294 (`return Err(WebRtcError::SignalingDisconnected());`). If the exchange times out, we should be able to exit the loop via L315, as the signaling server will send us an `answer_request::Stage::Error` if there was a timeout.

micro-rdk/src/common/webrtc/api.rs
benjirewis: That's sick, did not know that. I think because the `answer_request::Stage::Done(_)` arm was already doing `return Ok(None)`, I may just use `return Ok(None)` here so there's some consistency.

micro-rdk/src/common/webrtc/api.rs
gvaradarajan: This logic checks out to me, can resolve

micro-rdk/src/common/conn/server.rs
```diff
@@ -491,10 +491,10 @@ where
             let sig = if let Some(webrtc_config) = self.webrtc_config.as_ref() {
                 let ip = self.network.get_ip();
                 let rguard = self.app_client.read().await;
-                let signaling = rguard
-                    .as_ref()
-                    .unwrap()
-                    .initiate_signaling(self.rpc_host.clone());
+                let signaling = rguard.as_ref().unwrap().initiate_signaling(
+                    self.rpc_host.clone(),
+                    "true".to_string(), /* heartbeats allowed */
```
npmenard, line 496: if this is going to be hard coded we can skip passing it as a variable for now

micro-rdk/src/common/app_client.rs
```diff
@@ -187,12 +189,14 @@ impl AppClient {
     pub(crate) fn initiate_signaling(
         &self,
         rpc_host: String,
+        heartbeats_allowed: String,
     ) -> impl Future<Output = Result<AppSignaling, AppClientError>> {
         let (sender, receiver) = async_channel::bounded::<Bytes>(1);
         let r = self.grpc_client.build_request(
```
npmenard, line 195: the fucntion returns the request so it make more sense to mutate the request by adding the header after

micro-rdk/src/gen/proto.rpc.webrtc.v1.rs
```diff
@@ -272,14 +272,19 @@ pub struct AnswerRequestErrorStage {
     #[prost(message, optional, tag="1")]
     pub status: ::core::option::Option<super::super::super::super::google::rpc::Status>,
 }
+/// AnswerRequestHeartbeatStage is sent periodically to verify liveness of answerer.
```
npmenard, line 275: did you generate it with make buf?

micro-rdk/src/common/app_client.rs
```diff
@@ -187,12 +189,14 @@ impl AppClient {
     pub(crate) fn initiate_signaling(
         &self,
         rpc_host: String,
+        heartbeats_allowed: String,
```
npmenard, line 192: if we are going to hard-code it and opt out it not expected then we shouldn't need an argument for that

npmenard (changes requested):
done wiht first pass.
Ideally should be tested before merging

micro-rdk/src/gen/proto.rpc.webrtc.v1.rs
benjirewis: Yep! Lmk if that's not the correct method.

npmenard (approved):
LGTM did you test this?

benjirewis:
> LGTM did you test this?

I tested it with @gvaradarajan (thanks so much!), and I made a typo (`heartbeats_allowed` -> `heartbeats-allowed`), but otherwise things seem to work. Will fix that typo and rebase.

gvaradarajan (approved):
LGTM

PR #323, opened 2024-10-10
RSDK-8754 Fixup bad link and page title

acmorrow:
@gvaradarajan - I missed updating the link while renaming a file in the prior review, and then noticed a copy-paste error in the title of the new page.


gvaradarajan (approved):
Sorry I missed this as well

PR #325, opened 2024-10-10
[RSDK-8489] Upgrade mdns-sd for native

acmorrow:
I seem to get better mDNS behavior when testing with the native server after making this upgrade.


mattjperez (approved):
LGTM. Seems there were some nice improvements between .10 and .11. 

Shouldnt have compatibility issues as long as our domain ends in `.local`

npmenard (approved):

PR #327, opened 2024-10-14
Add support for timeout on getting the next ssl record for a connection

npmenard:


micro-rdk/src/esp32/dtls.rs
```diff
@@ -107,14 +91,36 @@ unsafe extern "C" fn mbedtls_net_read<S: Read>(
...
+    let buf = std::slice::from_raw_parts_mut(buf as *mut _, len);
+    match state.read(buf) {
+        Ok(len) => {
+            // if operation succeed cancel running timeout
+            state.reset_read_timeout();
+            len as c_int
+        }
+        Err(e) => {
+            let _ = state.error.insert(e);
+            if state.error.as_ref().unwrap().kind() == std::io::ErrorKind::WouldBlock {
```
gvaradarajan, line 115: match statement instead?

micro-rdk/src/esp32/dtls.rs
```diff
@@ -139,37 +145,114 @@ extern "C" fn ssl_debug(
...
+    fn get_delay(&self) -> i32 {
+        if self.fin.is_none() {
+            return -1;
+        }
+        let now = Instant::now();
+        if now > *self.intermediate.as_ref().unwrap() {
+            log::debug!("intermetidate timer expired");
+            return 2;
+        }
+        if now > *self.fin.as_ref().unwrap() {
```
gvaradarajan, line 211: shouldn't we check fin before intermediate?

micro-rdk/src/esp32/dtls.rs
```diff
@@ -426,24 +510,24 @@ impl SSLContext {
     }
 }
 
-pub struct SSLStream<S> {
+pub(crate) struct SSLStream<S> {
```
gvaradarajan, line 513: this is holding a pointer and interacting with it in an unsafe way, is it ok that you don't have a drop implementation for de-allocating the memory behind the pointer?

gvaradarajan (changes requested):
Had a couple questions

micro-rdk/src/esp32/dtls.rs
npmenard: good point

micro-rdk/src/esp32/dtls.rs
npmenard: nice catch that was really wrong :)

micro-rdk/src/esp32/dtls.rs
npmenard: if you are talking about bio_ptr it should be dropped on line 519

gvaradarajan: Ah I missed that, my bad

gvaradarajan (approved):
LGTM

PR #328, opened 2024-10-15
[RSDK-8999] Bulk upgrade crates

acmorrow:
There are many crates with newer versions that are source compatible with the micro-rdk. Apply those upgrades, pin those that cannot be upgraded, and ticket identified future work.

- Apply manual pins
- cargo upgrade --rust-version 1.76 --recursive -i
- cargo update
- Upgrade mentions of rust 1.75 to rust 1.76
- Add TODOS with tickets for pins and other future upgrades

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -76,6 +76,7 @@ futures-util = "0.3.31"
 http-body-util = "0.1.2"
 hyper = { version = "1.5", default-features = false, features = ["server", "client", "http2"] }
 ignore = "0.4"
+# TODO(RSDK-9045): Stop using lazy-static
 lazy_static = "1.5.0"
 local-ip-address = "0.6.3"
 log = "0.4.22"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,7 +74,7 @@ futures-lite = "2"
 futures-rustls = "~0.24"
 futures-util = "0.3.31"
 http-body-util = "0.1.2"
-hyper = { version = "1.4", default-features = false, features = ["server", "client", "http2"] }
+hyper = { version = "1.5", default-features = false, features = ["server", "client", "http2"] }
 ignore = "0.4"
 lazy_static = "1.5.0"
 local-ip-address = "0.6.3"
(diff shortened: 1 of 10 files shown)


templates/project/Cargo.toml
```diff
@@ -18,26 +19,27 @@ git = "https://github.com/viamrobotics/micro-rdk.git"
...
 [build-dependencies]
 cargo_metadata = "0.18"
-embuild = "0.29"
-regex = "1.10"
-serde = { version = "1.0.156", features = ["derive"] }
-serde_json = "1.0.97"
+embuild = "0.32"
+regex = "1.11"
+serde = { version = "1.0", features = ["derive"] }
+serde_json = "1.0"
```
mattjperez, line 39: since this deals with serialization stuff, should we pin the serde crates to the latest patch version? or are we relying on Cargo.lock when it comes to this?

Cargo.toml
```diff
@@ -38,79 +39,86 @@ debug = true # Symbols are nice and they don't increase the size on Flash
...
+futures-rustls = "=0.24"
+futures-util = "0.3.31"
+http-body-util = "0.1.2"
 hyper = { version = "1.4", default-features = false, features = ["server", "client", "http2"] }
-ignore = "=0.4.20"
-lazy_static = "1.4.0"
-local-ip-address = "0.5.3"
-log = "0.4.20"
+ignore = "0.4"
+lazy_static = "1.5.0"
```
mattjperez, line 79: should we add a TODO to use the std versions and remove this dependency?

mattjperez (changes requested):
looks good, but have a couple questions. 

do we have a preference on what gets pinned to minor versions or patch versions?

gvaradarajan (approved):
LGTM, I agree with @mattjperez 's point of making a ticket for moving us off of lazy_static

npmenard (approved):
LGTM

templates/project/Cargo.toml
acmorrow: Relying on `Cargo.lock`. This isn't a 'pin' really, to my mind, because there is still a ^ dependency, implicitly, so would allow any semver compatible upgrade (e.g. could go to 1.1 but not 2.0).

Cargo.toml
acmorrow: Done: https://viam.atlassian.net/browse/RSDK-9045

mattjperez (approved):
LGTM

acmorrow:
@mattjperez - Updated, PTAL:
- Converted `=x.y` style pins to `~x.y` per https://stackoverflow.com/questions/79098874/what-are-the-semantics-of-an-underspecified-exact-version-pin-in-cargo.
- Added TODO with ticket to remove `lazy-static`.
- Removed `ignore` crate which was unused
- Did another round of cargo update / cargo upgrade to pull in the latest versions of things.

acmorrow:
> * Removed `ignore` crate which was unused

Scratch that, I didn't remove that yet, I was waiting on an update from @gvaradarajan about whether or not it was being used somewhere I wasn't seeing.

PR #330, opened 2024-10-17
[RSDK-9049] cache TLS certificates for isolated network connection

gvaradarajan:


acmorrow (approved):
LGTM as well, with an optional logging suggestion

micro-rdk/src/common/conn/viam.rs
```diff
@@ -553,22 +555,29 @@ where
...
         if self.http2_server.has_http2_server() && !self.http2_server_insecure {
-            // TODO implement certificate storage
-            // in the situation where we are not connected to app or retrieving certs fails AND there are
-            // no certs stored then HTTP2 **should** be disabled.
-            let mut certs = None;
+            // Try to obtain and store a fresh TLS certificate. If this fails or we cannot reach
+            // app, then we'll end up falling back on whatever TLS certificate was cached
             if let Some(app) = app_client.as_ref() {
-                certs = app.get_certificates().await.ok();
+                if let Ok(cert_resp) = app.get_certificates().await {
```
npmenard, line 561: Can we change the logic so that we only load certs from the nvs if the app client is absent? If storing the cert failed but the response passed then we should be able to continue regardless.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -553,22 +555,29 @@ where
...
+                    log::error!("no TLS certificates found in storage or after contacting app");
+                    // At no point were we ever able to obtain a valid TLS certificate, so we disable HTTP2
                     let _ = std::mem::replace(&mut self.http2_server, HTTP2Server::Empty);
                 }
                 Some(certs) => {
                     if let HTTP2Server::HTTP2Connector(s) = &mut self.http2_server {
                         s.set_server_certificates(
-                            certs.tls_certificate.into_bytes(),
-                            certs.tls_private_key.into_bytes(),
+                            certs.certificate.clone(),
```
npmenard, line 579: can't we just move the certs?

micro-rdk/src/common/conn/viam.rs
```diff
@@ -553,22 +555,29 @@ where
...
+                        Ok(_) => log::debug!("storing TLS certificate received by app"),
+                        Err(err) => log::error!("error storing TLS cert: {:?}", err),
+                    };
+                }
             }
+            let certs = self.storage.get_tls_certificate().ok();
+
             match certs {
                 None => {
+                    log::error!("no TLS certificates found in storage or after contacting app");
```
npmenard, line 572: is that true all the time or only if the app was indeed contacted?

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -75,6 +78,29 @@ impl WifiCredentials {
     }
 }
 
+#[derive(Clone, Debug, Default)]
+pub struct TlsCertificate {
```
npmenard, line 82: can we fold this trait into the RobotStorage trait?

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -192,6 +195,35 @@ impl WifiCredentialStorage for NVSStorage {
...
+        let private_key = self.get_blob(NVS_TLS_PRIVATE_KEY_KEY)?;
+        Ok(TlsCertificate {
+            certificate,
+            private_key,
+        })
+    }
+    fn store_tls_certificate(&self, creds: TlsCertificate) -> Result<(), Self::Error> {
+        self.set_blob(
+            NVS_TLS_CERTIFICATE_KEY,
+            Bytes::from(creds.certificate.clone()),
```
npmenard, line 215: if we are going to clone we should take a ref instead

npmenard (changes requested):
Done with first pass

micro-rdk/src/common/conn/viam.rs
gvaradarajan: I wanted to avoid changing the trait, but if we feel it's worth it I can

micro-rdk/src/common/conn/viam.rs
gvaradarajan: it wasn't before, but after the change you requested above it will be I believe

npmenard (approved):
LGTM

PR #331, opened 2024-10-21
fix canary not notifying channel when every connection attempt failed

gvaradarajan:


canary/daily_summary.py
```diff
@@ -67,6 +67,8 @@ def get_summary_for_platform(coll: collection.Collection, platform: str) -> Tupl
     avg_connection_attempts = connection_attempts / num_results
     
     total_board_calls = board_api_successes + board_api_failures
+    if total_board_calls == 0:
```
stevebriskin, line 70: what will the record in the DB look like for these cases?
I'm looking at the `num_results == 0` condition and thinking if that'll trigger before this statement. And whether no_results should also result in an alert as that implies something is horribly wrong.

stevebriskin (approved):

PR #332, opened 2024-10-21
[RSDK-9080] fix incorrect logic validating proper ESP32 GPIO pins

gvaradarajan:


gvaradarajan:
Opening this PR because it works, but it seems to use the opposite logic from [here](https://github.com/espressif/esp-idf/blob/a97a7b0962da148669bb333ff1f30bf272946ade/components/esp_driver_gpio/include/driver/gpio.h#L28), so I don't know why? @npmenard , @acmorrow , any insights?

gvaradarajan:
> Not sure why you think it is the opposite logic according the the constant definition we have :
> 
> ```
> #define SOC_GPIO_VALID_GPIO_MASK        (0xFFFFFFFFFFULL & ~(0ULL | BIT24 | BIT28 | BIT29 | BIT30 | BIT31))
> // GPIO >= 34 are input only
> #define SOC_GPIO_VALID_OUTPUT_GPIO_MASK (SOC_GPIO_VALID_GPIO_MASK & ~(0ULL | BIT34 | BIT35 | BIT36 | BIT37 | BIT38 | BIT39))
> 
> // digital I/O pad powered by VDD3P3_CPU or VDD_SPI(GPIO_NUM: 1, 3, 5, 6, 7, 8, 9, 10, 11, 16, 17, 18, 19, 21, 22, 23)
> #define SOC_GPIO_VALID_DIGITAL_IO_PAD_MASK 0xEF0FEAULL
> ```
> 
> which seems to follow definitions at https://www.espressif.com/sites/default/files/documentation/esp32-wrover-e_esp32-wrover-ie_datasheet_en.pdf

@npmenard yeah it does seem to align with the datasheet, I think I was just reading it wrong. Well this works. It passes for the pins that had failed in the ticket and fails for input pins like 34. (But I think we should consider this a temporary measure, you should be allowed to get high/low from input pins)

acmorrow (approved):
LGTM

PR #333, opened 2024-10-22
[RSDK-6522] implement cloud metadata endpoint

gvaradarajan:


micro-rdk/src/common/robot.rs
```diff
@@ -265,6 +274,11 @@ impl LocalRobot {
             data_manager_collection_task: None,
             start_time: Instant::now(),
         };
+        if let Some(cloud_cfg) = config.cloud.as_ref() {
```
acmorrow, line 277: Is there some nice method on `Option` you could use above on 263-265 to achieve this in one go? An `or_default` type thing?

acmorrow, line 277: Under what circumstances will `config.cloud()` actually be `None`? Should it be possible to build a `LocalRobot` from a `config` lacking a `cloud` data?

micro-rdk/src/common/robot.rs
```diff
@@ -867,6 +881,16 @@ impl LocalRobot {
         }
         Ok(())
     }
+
+    pub fn get_cloud_metadata(&self) -> robot::v1::GetCloudMetadataResponse {
```
acmorrow, line 885: Should this method return an error if these values were not available? What does RDK do?

acmorrow (commented):
Code looks fine if this is the desired behavior, I just have some questions / small suggestions about the behavior itself.

micro-rdk/src/common/robot.rs
gvaradarajan: there's something I can do that is less efficient, I'll push it up and see if you prefer it

gvaradarajan: I think you should be able to from a library perspective, we can create a robot without starting a server. But maybe that means this shouldn't be stored on the robot?

micro-rdk/src/common/robot.rs
gvaradarajan: it should, I'll make the change

acmorrow (approved):
LGTM

.gitignore
```diff
@@ -6,3 +6,4 @@ Cargo.lock
 .vscode
 components_esp32*.lock
 .envrc
+/build
```
npmenard, line 9: why this addition in the ignores?

micro-rdk/src/common/robot.rs
```diff
@@ -254,6 +265,11 @@ impl LocalRobot {
         let mut robot = LocalRobot {
             executor: exec,
             part_id,
+            cloud_metadata: config.cloud.as_ref().map(|cfg| CloudMetadata {
+                org_id: cfg.primary_org_id.clone(),
+                location_id: cfg.location_id.clone(),
+                machine_id: cfg.machine_id.clone(),
```
npmenard, line 271: is part_id == machine_id?

micro-rdk/src/common/robot.rs
```diff
@@ -106,6 +113,7 @@ pub struct LocalRobot {
     // is NOT a valid timestamp. For actual timestamps, the real time should be set on the system
     // at some point using settimeofday (or something equivalent) and referenced thereof.
     pub(crate) start_time: Instant,
+    cloud_metadata: Option<CloudMetadata>,
```
npmenard, line 116: even if it's somewhat incorrect I think we can avoid the Option here. In can't see and case where we wouldn't be able to populate CloudMetadata from the config. And if a robot is constructed by default then we should be in provisioning

npmenard (approved):
Some comments lgtm otherwise

micro-rdk/src/common/robot.rs
gvaradarajan: no, there's a separate "machine_id" and "machine_part_id"

.gitignore
gvaradarajan: CMake kept generating a build directory because of the C code and I was tired of accidentally committing it

micro-rdk/src/common/robot.rs
gvaradarajan: Won't we be able to serve on an isolated network without cloud metadata?

gvaradarajan: Scratch that previous question, my actual concern is this: the cloud meta data is an option in the config proto itself, so by leaving it as an option I hope to stop a potential error in the config endpoint that should be inconsequential from breaking the ability to create a robot from the config

PR #334, opened 2024-10-23
[RSDK-9116] Return proper error when message is larger than store capacity

gvaradarajan:


micro-rdk/src/common/data_store.rs
```diff
@@ -253,7 +253,9 @@ impl DataStore for DefaultDataStore {
         }
         let encode_len = message.encoded_len();
         let total_encode_len = length_delimiter_len(encode_len) + encode_len;
-
+        if encode_len > buffer.capacity() {
+            return Err(DataStoreError::DataTooLarge);
```
mattjperez, line 257: can we include the `encoded_len` in the error for people to adjust properly?

mattjperez (approved):
LGTM, just a comment

micro-rdk/src/common/data_store.rs
```diff
@@ -253,7 +253,14 @@ impl DataStore for DefaultDataStore {
         }
         let encode_len = message.encoded_len();
         let total_encode_len = length_delimiter_len(encode_len) + encode_len;
-
+        let buffer_capacity = buffer.capacity();
```
acmorrow, line 256: The first time I read this code, I thought I was looking at a bug, because why would you be comparing the total encoded length to the _capacity_ of the buffer, rather than the available length. But I realized after looking a little more closely that it is in fact correct because after the loop below starting on line 264, `buffer.vacant_len` will eventually reach `buffer.capacity` as messages are dropped. So checking against `capacity` here is correct, but I think a comment is warranted.

micro-rdk/src/common/data_store.rs
```diff
@@ -5,7 +5,7 @@
 use crate::proto::app::data_sync::v1::SensorData;
 use bytes::{Buf, BufMut, BytesMut};
 use prost::{encoding::decode_varint, length_delimiter_len, DecodeError, EncodeError, Message};
-use ringbuf::{ring_buffer::RbBase, Consumer, LocalRb, Producer};
+use ringbuf::{ring_buffer::RbBase, Consumer, LocalRb, Producer, Rb};
```
acmorrow, line 8: Is `Rb` now used directly somewhere? I didn't quite see where.

acmorrow (approved):
LGTM, though I think a comment is a good idea.

micro-rdk/src/common/data_store.rs
npmenard: +1 it can be difficult to decipher

micro-rdk/src/common/data_store.rs
```diff
@@ -253,7 +253,14 @@ impl DataStore for DefaultDataStore {
         }
         let encode_len = message.encoded_len();
         let total_encode_len = length_delimiter_len(encode_len) + encode_len;
-
+        let buffer_capacity = buffer.capacity();
+        if encode_len > buffer_capacity {
+            return Err(DataStoreError::DataTooLarge(
```
npmenard, line 258: would we drop the full SensorData object?

npmenard (approved):
LGTM

micro-rdk/src/common/data_store.rs
gvaradarajan: I don't think so, huh. I wonder why clippy didn't catch it

gvaradarajan: Ah, it's a trait. Needs to be imported in order to call `buffer.capacity()`

micro-rdk/src/common/data_store.rs
gvaradarajan: I believe so, because `write_message` takes ownership of the SensorData object

PR #335, opened 2024-10-23
[RSDK-8924] have GrpcClient accept address as Uri rather than String

gvaradarajan:


micro-rdk/src/common/grpc_client.rs
```diff
@@ -267,8 +275,12 @@ impl GrpcClient {
         rpc_host: &str,
         body: B,
     ) -> Result<Request<B>, GrpcClientError> {
-        let mut uri = self.uri.to_owned();
-        uri.push_str(path);
+        let mut uri_base_parts = self.uri.clone().into_parts();
+        // we assume here that the API does not support nesting paths
+        // like /SomeOrg/proto.rpc.v1.AuthService/Authenticate and simply
+        // replace whatever existing path is there on the Uri instance with `path`
```
acmorrow, line 281: Rather than doing this as a comment, maybe in `GrpcClient::new` you could destructure the provided URI into `Scheme` and `Authority` objects by calling `into_parts` _there_, and reject it if you get a non-empty `PathAndQuery` component. Then, here, just create a new `Parts` from clones of the owned `Scheme` and `Authority` objects, with the `path.parse::<PathAndQuery>` for the last component.

acmorrow (approved):
LGTM with an optional suggestion, if it works.

micro-rdk/src/common/grpc_client.rs
gvaradarajan: Discussed offline, I've opted not to go with this because I believe we may someday support nested APIs and this is a more natural place to look for changing the logic

PR #338, opened 2024-10-24
[RSDK-8923] Acquire app address from cloud credentials

gvaradarajan:
Instead of hardcoding the signaling address to http://app.viam.com:443, we now use the "app_address" key given in cloud credentials

IMPORTANT NOTE: As a follow-up, the logic in the project template under the "has_robot_config" condition must be updated upon version bump!


templates/project/src/main.rs
```diff
@@ -80,6 +81,7 @@ fn main() {
 
     if !storage.has_robot_configuration() {
         // check if any were statically compiled
+        // TODO: update with app address storage logic when version is incremented
```
acmorrow, line 84: I'd suggest making a ticket with a fixVersion and adding the ticket into the TODO here as `TODO(RSDK-WXYZ): Update ...`.

acmorrow (approved):
LGTM with a few small comments / suggestions

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -29,6 +33,10 @@ impl RobotCredentials {
     pub(crate) fn robot_secret(&self) -> &str {
         &self.robot_secret
     }
+
+    pub(crate) fn app_address(&self) -> Uri {
+        self.app_address.parse::<Uri>().unwrap()
```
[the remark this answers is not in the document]
gvaradarajan: There's no good reason, I'll change it accordingly

templates/project/src/main.rs
gvaradarajan: will do

micro-rdk/src/common/conn/viam.rs
```diff
@@ -1403,7 +1394,7 @@ mod tests {
         let creds = CloudConfig {
             id: "".to_string(),
             secret: "".to_string(),
-            app_address: "".to_owned(),
+            app_address: "http://localhost:56563".to_owned(),
```
mattjperez, line 1397: lets make this a local const since it's referenced so often

mattjperez (approved):
LGTM, just a suggestion

PR #342, opened 2024-10-28
[RSDK-9158] fix key write problem for app address in installer

gvaradarajan:


acmorrow (approved):
The fix is fine, so LGTM, but:
- The commit message should reference https://viam.atlassian.net/browse/RSDK-9158, which I've just assigned to you and put in progress.
- Please file a follow-up ticket to find a way to re-use the constants from `nvs_storage.rs`. It should be a follow-up because right now those constants aren't `pub` and the relevant file is esp32 gated so there isn't a way to use them without a bigger refactoring.

gvaradarajan:
Follow-up ticket for refactor here: https://viam.atlassian.net/browse/RSDK-9162

PR #344, opened 2024-10-30
adjust README module dependency example

gvaradarajan:
The previous line was not working for some reason when used in an actual project. This new way is more canonical to Cargo anyway I believe.

### DIFF
--- a/examples/modular-drivers/README.md
+++ b/examples/modular-drivers/README.md
@@ -35,7 +35,7 @@ index 79fbc5c..949b5ab 100644
  smol = "1.2"
  futures-lite = "1.12.0"
  micro-rdk = {version = "0.0.3", git = "https://github.com/viamrobotics/micro-rdk.git", features = ["esp32"]}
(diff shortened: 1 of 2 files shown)


examples/modular-drivers/README.md
```diff
@@ -35,7 +35,7 @@ index 79fbc5c..949b5ab 100644
  smol = "1.2"
  futures-lite = "1.12.0"
  micro-rdk = {version = "0.0.3", git = "https://github.com/viamrobotics/micro-rdk.git", features = ["esp32"]}
-+micro-rdk-modular-driver-example = { git = "https://github.com/viamrobotics/micro-rdk/tree/main/examples/modular-drivers" }
```
mattjperez, line 38: does it need the `package` specification? I figured the workspace would direct it if the name is the same as the intended package

mattjperez (approved):
LGTM, non-blocking question/suggestion

acmorrow (approved):
LGTM

examples/modular-drivers/README.md
gvaradarajan: it doesn't but I think it's better communication to a reader about how it actually works to include it

PR #346, opened 2024-11-05
[RSDK-9216] - allocate buffer for response message per RPC call

gvaradarajan:


gvaradarajan:
Camera code is untested on hardware, will refrain from merging until it is but wanted the code review regardless

mattjperez:
tested with fake camera and esp32camera, still works

micro-rdk/src/common/grpc.rs
```diff
@@ -1357,7 +1356,7 @@ impl<'a> GrpcServerInner<'a> {
             .get_camera_by_name(req.name)
             .ok_or(GrpcError::RpcUnavailable)?;
 
-        let mut buffer = RefCell::borrow_mut(&self.buffer).split_off(0);
+        let mut buffer = BytesMut::with_capacity(CAMERA_BUFFER_SIZE);
```
mattjperez, line 1359: shouldn't this use the new allocator?

micro-rdk/src/common/grpc.rs
```diff
@@ -1384,7 +1383,8 @@ impl<'a> GrpcServerInner<'a> {
...
```
mattjperez, line 1386: same here

micro-rdk/src/esp32/rpc_heap.rs
```diff
@@ -0,0 +1,52 @@
...
+    // We source the memory for this Vec from ESP-IDF's heap_caps_malloc
+    // because it will error when not enough space is available rather than panic.
+    // As a result we must wrap the Vec in a ManuallyDrop to prevent double-free, since
+    // the memory will not be managed by Rust's allocators
+    inner: ManuallyDrop<Vec<u8>>,
+    ptr: *mut u8,
+}
+
+impl RpcAllocation for Esp32RpcHeapAllocation {
+    fn get_allocation(size: usize) -> Result<Self, GrpcError> {
```
mattjperez, line 22: I think these deserve their own error types and get propagated upward to `GrpcError`

micro-rdk-server/native/main.rs
```diff
@@ -83,7 +83,7 @@ mod native {
             mdns,
             Box::new(network),
         );
-        server.run_forever();
+        server.run_forever::<bytes::BytesMut>();
```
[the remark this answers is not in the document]
gvaradarajan: I don't quite see a reason to?

micro-rdk/src/common/grpc.rs
gvaradarajan: I didn't bother because it would simply change immediately in the next PR

micro-rdk/src/esp32/rpc_heap.rs
gvaradarajan: I'd prefer to save doing that until we have an implementation of the trait that would actually want to surface any other error

micro-rdk-server/native/main.rs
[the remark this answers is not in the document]
mattjperez: just that the leap from `BytesMut` to `NativeHeapAllocation` feels odd. One is external with a generic name, not hinting on how it might work underneath, the other is the opposite (granted, because we're the ones implementing it)

mattjperez (approved):
LGTM, but I think @acmorrow would like to coordinate this with his PR

acmorrow:
@gvaradarajan - Yes, please don't merge this until we land #347.

acmorrow:
> @gvaradarajan - Yes, please don't merge this until we land #347.

Also, please let me get a round of review in on it.

micro-rdk-server/native/main.rs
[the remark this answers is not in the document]
gvaradarajan: Ok, that's a fair point about clarity, I'll make this change

micro-rdk/src/esp32/rpc_heap.rs
```diff
@@ -0,0 +1,52 @@
...
+        let ptr = unsafe { heap_caps_malloc(size, MALLOC_CAP_SPIRAM | MALLOC_CAP_8BIT) } as *mut u8;
+        if ptr.is_null() {
+            Err(GrpcError::RpcResourceExhausted)
+        } else {
+            let inner = ManuallyDrop::new(unsafe { Vec::from_raw_parts(ptr, size, size) });
```
npmenard, line 27: you shouldn't set the len it implies that the underlying data is initialized

micro-rdk/src/esp32/rpc_heap.rs
```diff
@@ -0,0 +1,52 @@
...
+        }
+        if 5 + m.encoded_len() > buffer.capacity() {
+            return Err(GrpcError::RpcResourceExhausted);
+        }
+        buffer.put_u8(0);
+        buffer.put_u32(m.encoded_len().try_into().unwrap());
+        let mut msg = buffer.split_off(5);
+        m.encode(&mut msg).map_err(|_| GrpcError::RpcInternal)?;
+        buffer.unsplit(msg);
+        Ok(buffer.freeze())
```
npmenard, line 44: the Esp32RpcHeapAllocation object will be dropped here and memory will be release.

micro-rdk-server/esp32/main.rs
```diff
@@ -152,7 +153,7 @@ mod esp32 {
         };
         #[cfg(not(feature = "qemu"))]
         let mut server = { builder.build(Esp32H2Connector::default(), Executor::new(), mdns) };
-        server.run_forever();
+        server.run_forever::<Esp32RpcHeapAllocation>();
```
npmenard, line 156: I don't think we should add a Generic here

gvaradarajan:
> I am not sure the code works the way you intended it to. Perhaps we should use the allocator_api and replace all Bytes with Vec<u8, AllocatorESP32> since it implements Buf + BufMut that should be possible. To simplify the code I would use a global type so Vec<u8, A> get swapped for Vec in native builds

@npmenard, we can't use the allocator API because it panics rather than returns an error when unable to allocate (except in Nightly Rust)

micro-rdk/src/esp32/rpc_heap.rs
gvaradarajan: isn't that good?

npmenard: yes but it doesn't look like we ever use the memory allocated by get_allocation.

npmenard:
esp toolchain should be nightly no?

micro-rdk-server/esp32/main.rs
mattjperez: should it be an input parameter to `run_forever`?

gvaradarajan:
> esp toolchain should be nightly no?

yes but not for native builds. So making a consistent API requires wrapping

gvaradarajan:
This PR has a lot of fundamental problems, closing

PR #347, opened 2024-11-06
[RSDK-8489] Implement local webrtc signaling

acmorrow:
I will add inline annotations to provide context on the current state of things, and what needs to happen next.

### DIFF
--- a/micro-rdk/src/common/conn/viam.rs
+++ b/micro-rdk/src/common/conn/viam.rs
@@ -14,6 +14,7 @@ use crate::common::app_client::{
     AppClient, AppClientBuilder, AppClientError, PeriodicAppClientTask,
 };
 use crate::common::credentials_storage::TlsCertificate;
+use crate::common::webrtc::signaling_server::SignalingServer;
 use std::marker::PhantomData;
 use std::net::{SocketAddr, TcpListener};
 use std::pin::Pin;
@@ -622,7 +623,9 @@ where
             }
         }
 
-        let (tx, rx) = async_channel::bounded(1);
+        // One slot for remote signaling, one for local.
+        let (tx, rx) = async_channel::bounded(2);
+
         let mut inner = RobotServer {
             http2_server: &self.http2_server,
             http2_server_port: self.http2_server_port,
@@ -636,6 +639,10 @@ where
                 self.max_concurrent_connections as usize,
             ),
             robot_config: &config,
+            local_signaling_server: Arc::new(SignalingServer::new(
+                self.executor.clone(),
+                tx.clone(),
+            )),
         };
 
         if let Some(cfg) = config.cloud.as_ref() {
@@ -774,9 +781,10 @@ struct RobotServer<'a, M> {
     network: &'a dyn Network,
     incomming_connection_manager: IncomingConnectionManager,
     robot_config: &'a RobotConfig,
+    local_signaling_server: Arc<SignalingServer>,
 }
 
-pub(crate) enum IncomingConnection2 {
+pub(crate) enum IncomingConnection {
     HTTP2Connection(std::io::Result<(Async<TcpStream>, SocketAddr)>),
     WebRTCConnection(Result<Box<WebRtcSignalingChannel>, WebRtcError>),
 }
@@ -791,9 +799,10 @@ where
     ) -> Task<Result<(), errors::ServerError>> {
         let exec = self.executor.clone();
         let robot = self.robot.clone();
+        let ss = self.local_signaling_server.clone();
         self.executor.spawn(async move {
-            let srv = GrpcServer::new(robot, GrpcBody::new());
-
+            let mut srv = GrpcServer::new(robot, GrpcBody::new());
+            srv.register_signaling_server(ss);
             http2::Builder::new(exec)
                 .initial_connection_window_size(2048)
                 .initial_stream_window_size(2048)
@@ -806,10 +815,10 @@ where
     }
     async fn serve_incoming_connection(
         &mut self,
-        incoming: IncomingConnection2,
+        incoming: IncomingConnection,
     ) -> Result<(), errors::ServerError> {
         match incoming {
-            IncomingConnection2::HTTP2Connection(conn) => {
+            IncomingConnection::HTTP2Connection(conn) => {
                 if let HTTP2Server::HTTP2Connector(h) = self.http2_server {
                     if self.incomming_connection_manager.get_lowest_prio() < u32::MAX {
                         let stream = conn?;
@@ -822,7 +831,8 @@ where
                     }
                 }
             }
-            IncomingConnection2::WebRTCConnection(conn) => {
+
+            IncomingConnection::WebRTCConnection(conn) => {
                 let sig = conn.map_err(|e| errors::ServerError::Other(e.into()))?;
                 let ip = self.network.get_ip();
                 if let WebRtcListener::WebRtc(conf) = self.webrtc_config {
@@ -861,7 +871,7 @@ where
                     "_rpc",
                     "_tcp",
                     self.http2_server_port,
-                    &[("grpc", "")],
+                    &[("grpc", ""), ("webrtc", "")],
                 )
                 .map_err(|e| errors::ServerError::Other(e.into()))?;
                 mdns.add_service(
@@ -869,7 +879,7 @@ where
                     "_rpc",
                     "_tcp",
                     self.http2_server_port,
-                    &[("grpc", "")],
+                    &[("grpc", ""), ("webrtc", "")],
                 )
                 .map_err(|e| errors::ServerError::Other(e.into()))?;
             }
@@ -882,28 +892,28 @@ where
         };
 
         loop {
-            let h2_conn: Pin<Box<dyn Future<Output = IncomingConnection2>>> =
+            let h2_conn: Pin<Box<dyn Future<Output = IncomingConnection>>> =
                 if let HTTP2Server::Empty = self.http2_server {
                     Box::pin(async {
-                        IncomingConnection2::HTTP2Connection(futures_lite::future::pending().await)
+                        IncomingConnection::HTTP2Connection(futures_lite::future::pending().await)
                     })
                 } else {
                     // safe to unwrap, always exists
                     Box::pin(async {
-                        IncomingConnection2::HTTP2Connection(
+                        IncomingConnection::HTTP2Connection(
                             http2_listener.as_ref().unwrap().accept().await,
                         )
                     })
                 };
 
-            let webrtc_conn: Pin<Box<dyn Future<Output = IncomingConnection2>>> =
+            let webrtc_conn: Pin<Box<dyn Future<Output = IncomingConnection>>> =
                 if let WebRtcListener::Empty = self.webrtc_config {
                     Box::pin(async {
-                        IncomingConnection2::HTTP2Connection(futures_lite::future::pending().await)
+                        IncomingConnection::HTTP2Connection(futures_lite::future::pending().await)
                     })
                 } else {
                     Box::pin(async {
-                        IncomingConnection2::WebRTCConnection(
+                        IncomingConnection::WebRTCConnection(
                             self.webrtc_signaling
                                 .recv()
                                 .await
--- a/micro-rdk/src/common/grpc.rs
+++ b/micro-rdk/src/common/grpc.rs
@@ -13,11 +13,11 @@ use crate::{
         webrtc::grpc::WebRtcGrpcService,
     },
     google::rpc::Status,
-    proto::{self, component, robot},
+    proto::{self, component, robot, rpc::webrtc::v1::CallResponse},
 };
 use bytes::{BufMut, BytesMut};
-use futures_lite::{future, Future};
-use http_body_util::BodyExt;
+use futures_lite::{Future, StreamExt};
+use http_body_util::{combinators::BoxBody, BodyExt, StreamBody};
 use hyper::{
     body::{self, Body, Bytes, Frame},
     http::{uri::InvalidUri, HeaderValue},
@@ -27,12 +27,12 @@ use hyper::{
 use log::*;
 use prost::Message;
 use std::{
-    cell::RefCell,
     pin::Pin,
-    rc::Rc,
     task::{Context, Poll},
 };
 use thiserror::Error;
+
+use super::webrtc::signaling_server::SignalingServer;
 
 #[cfg(feature = "camera")]
 static GRPC_BUFFER_SIZE: usize = 1024 * 30; // 30KB
@@ -125,27 +125,15 @@ pub trait GrpcResponse {
 
 #[derive(Clone)]
 pub struct GrpcServer<R> {
-    pub(crate) response: R,
-    pub(crate) buffer: Rc<RefCell<BytesMut>>,
+    pub(crate) _response: R,
     robot: Arc<Mutex<LocalRobot>>,
+    signaling_server: Option<Arc<SignalingServer>>,
 }
 
 pub struct GrpcServerInner<'a> {
-    buffer: &'a Rc<RefCell<BytesMut>>,
+    buffer: Option<BytesMut>,
     robot: &'a Arc<Mutex<LocalRobot>>,
-}
-
-impl<R> Debug for GrpcServer<R>
-where
-    R: Debug,
-{
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(
-            f,
-            "GrpcServer {{ response : {:?} }}, {{ buffer : {:?} }}",
-            self.response, self.buffer
-        )
-    }
+    signaling_server: &'a Option<Arc<SignalingServer>>,
 }
 
 impl<R> GrpcServer<R>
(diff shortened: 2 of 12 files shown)


micro-rdk/src/common/grpc.rs
```diff
@@ -1435,35 +1439,99 @@ impl<'a> GrpcServerInner<'a> {
...
+
+        match self.signaling_server {
+            Some(ss) => {
+                let cr = proto::rpc::webrtc::v1::CallRequest::decode(message);
+                let ss = ss.clone();
+                ss.executor
+                    .clone()
+                    .spawn(async move {
+                        match cr {
+                            Ok(cr) => match ss.call(cr, sender.clone()).await {
```
acmorrow, line 1470: I think this pattern of having
```
async fn(request, sender) -> Result<(), Error>
```

Is going to be somewhat common. You want to be able to write the function normally with `?` and such to check preconditions, but then once you are into the streaming part you are sending results/errors to the channel, rather than returning them.

I'd like to have a combinator that wraps a function of that shape and feeds any `Err` result back into the channel.

micro-rdk/src/common/grpc.rs
```diff
@@ -1435,35 +1439,99 @@ impl<'a> GrpcServerInner<'a> {
...
-        }
-        let mut grpc = GrpcServerInner {
-            buffer: &self.buffer,
+        let grpc = GrpcServerInner {
+            buffer: Some(BytesMut::with_capacity(GRPC_BUFFER_SIZE)),
             robot: &self.robot,
+            signaling_server: &self.signaling_server,
         };
-        grpc.handle_request(method, data)
+        grpc.handle_unary_request(method, data)
```
acmorrow, line 1522: I did not extend streaming support into WebRTC yet, so this can and should bypass the stream based response path and go directly to the result returning path.

micro-rdk/src/common/grpc.rs
```diff
@@ -1474,17 +1542,14 @@ impl<R> Service<Request<body::Incoming>> for GrpcServer<R>
 where
     R: GrpcResponse + Body + Clone + 'static,
 {
-    type Response = Response<R>;
+    type Response = Response<BoxBody<Bytes, GrpcError>>;
```
acmorrow, line 1545: Note that `R` no longer appears here, and we aren't using the `GrpcResponse` stuff at all. For that matter, we aren't using the `GrpcBody` trait either. I left all of that alone in the interest of getting this up for review, since removing that stuff would require compensating changes in the WebRTC based GrpcServer infra. A follow-up will refactor this, and hopefully simplify it all.

micro-rdk/src/common/grpc.rs
```diff
@@ -1499,11 +1564,68 @@ where
...
+            trailers.insert("grpc-status", "0".parse().unwrap());
+            struct UnfoldState {
+                trailers: HeaderMap,
+                #[allow(clippy::type_complexity)]
+                stream: Option<
+                    Pin<
+                        Box<
+                            dyn futures_lite::Stream<Item = Result<Bytes, ServerError>>
+                                + Sync
+                                + Send,
```
acmorrow, line 1584: The type system needs to know that the stream future we are putting in here is `Sync + Send` in order to let it be provided as the body. But my understanding is that the compiler can't derive that knowledge since `handle_request` returns a `dyn Stream`. So we need to be explicit about it.

micro-rdk/src/common/grpc.rs
```diff
@@ -1499,11 +1564,68 @@ where
...
+                }
+                Err(e) => {
+                    state.trailers.insert("grpc-status", e.status_code().into());
+                    state
+                        .trailers
+                        .insert("grpc-message", e.to_string().parse().unwrap());
+                }
+            };
+
+            let stream = futures_lite::stream::unfold(state, |mut state| async move {
```
acmorrow, line 1606: The `unfold` gadget is interesting: it lets you have a stateful stream. This might be better handled by implementing our own future on which we can `poll_frame` as GrpcBody does currently, but this was enough to get off the ground.

micro-rdk/src/common/grpc.rs
```diff
@@ -1499,11 +1564,68 @@ where
...
+                if let Some(ref mut stream) = state.stream {
+                    match stream.next().await {
+                        Some(Ok(bytes)) => {
+                            return Some((Ok(Frame::<Bytes>::data(bytes)), state));
+                        }
+                        Some(Err(e)) => {
+                            let _ = state.stream.take();
+                            state.trailers.insert("grpc-status", e.status_code().into());
```
acmorrow, line 1614: This should be dry'd with 1599 above.

micro-rdk/src/common/grpc.rs
```diff
@@ -1612,30 +1734,3 @@ impl fmt::Display for ServerError {
...
-    type Future = future::Ready<Result<Self::Response, Self::Error>>;
-
-    fn call(&self, _: T) -> Self::Future {
-        {
-            info!("reserve memory");
-            RefCell::borrow_mut(&self.server.buffer).reserve(10240);
-        }
-        future::ready(Ok(self.server.clone()))
-    }
-}
```
acmorrow, line 1641: This was all dead code

micro-rdk/src/common/webrtc/api.rs
```diff
@@ -79,27 +84,23 @@ pub enum WebRtcError {
     #[error("cannot parse candidate")]
     CannotParseCandidate,
     #[error("Operation timeout")]
-    OperationTiemout,
+    OperationTimeout,
     #[error(transparent)]
     GrpcClientError(#[from] GrpcClientError),
 }
 
+pub(crate) type WebRtcSignaling = Either<AppSignaling, LocalSignaling>;
```
acmorrow, line 92: This is the key change. We now have two different types that can represent a signaling request. One for app signaling, and one for local signaling.

I'm open to a better name for this type given how similar it is to `WebRtcSignalingChannel`. Maybe `WebRtcSignalingInvocation`?

micro-rdk/src/common/webrtc/api.rs
```diff
@@ -299,47 +300,92 @@ impl WebRtcSignalingChannel {
...
-            })),
-        };
-
-        if let Err(e) = self.signaling_tx.send_message(answer).await {
-            log::error!("error sending signaling message: {:?}", e);
-            Err(WebRtcError::SignalingDisconnected())
-        } else {
-            log::warn!("too many active connections");
-            Ok(())
+        match self.signaling.as_mut() {
```
acmorrow, line 303: The app signaling case is now the `Left` case, and the code there hasn't changed. The `Right` case is local signaling and is new code. That goes for many of the following methods.

I have made no effort to DRY things up between the app signaling and the local signaling case, but will do so in a follow-up

micro-rdk/src/common/webrtc/api.rs
```diff
@@ -349,88 +395,145 @@ impl WebRtcSignalingChannel {
...
+                let answer = AnswerResponse {
+                    uuid,
+                    stage: Some(answer_response::Stage::Done(AnswerResponseDoneStage {})),
+                };
+                match app_signaling.tx.send_message(answer).await {
+                    Err(_) => Err(WebRtcError::SignalingDisconnected()),
+                    Ok(_) => Ok(()),
+                }
+            }
+            Either::Right(ref mut local_signaling) => match local_signaling.tx.close() {
```
acmorrow, line 524: This is actually buggy. We should not be closing the channel here, because we may race with a CallUpdate that comes after we have sent all of the information from our side.

micro-rdk/src/common/webrtc/api.rs
```diff
@@ -349,88 +395,145 @@ impl WebRtcSignalingChannel {
...
+                true => Ok(()),
+                false => Err(WebRtcError::SignalingDisconnected()),
+            },
         }
     }
 }
 
-pub struct WebRtcApi<S, E> {
+pub struct WebRtcApi<C, E> {
```
acmorrow, line 532: fly-by: this was inconsistent with later usage.

micro-rdk/src/common/webrtc/ice.rs
```diff
@@ -193,6 +192,7 @@ impl ICEAgent {
                     Err(io::Error::new(io::ErrorKind::TimedOut, ""))
                 })
                 .await;
+
```
acmorrow, line 195: This is just ws noise from when I had some logging in this file, I'll revert it.

micro-rdk/src/common/webrtc/signaling_server.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        let channel = Box::new(WebRtcSignalingChannel::new(
+            Either::Right(local_signaling),
+            sdp,
+        ));
+
+        match self.sender.send(channel).await {
+            Ok(()) => {
+                let uuid: Option<String> = None;
+                let mut uuid_guard = scopeguard::guard(uuid, |uuid| {
```
acmorrow, line 106: This is actually also incorrect, in that we remove the uuid when we leave the loop due to a channel close (see the other noted problem), but we still may see a `CallUpdate` for that uuid.

micro-rdk/src/common/webrtc/signaling_server.rs
```diff
@@ -0,0 +1,176 @@
...
+        &self,
+        request: CallRequest,
+        responses: Sender<Result<CallResponse, ServerError>>,
+    ) -> Result<(), ServerError> {
+        if request.disable_trickle {
+            return Err(ServerError::new(GrpcError::RpcUnimplemented, None));
+        }
+
+        // TODO: Can we re-use this engine?
+        let engine: Box<general_purpose::GeneralPurpose> = general_purpose::STANDARD.into();
```
acmorrow, line 56: Building the SDP from the CallRequest is very similar to how we build it from the `AnswerRequest` in `wait_for_sdp`, and should be DRY'd.

micro-rdk/src/common/webrtc/signaling_server.rs
```diff
@@ -0,0 +1,176 @@
...
+        let (response_tx, response_rx) =
+            async_channel::bounded::<Result<CallResponse, ServerError>>(1);
+
+        let local_signaling = LocalSignaling {
+            tx: response_tx,
+            rx: update_rx,
+        };
+
+        let channel = Box::new(WebRtcSignalingChannel::new(
+            Either::Right(local_signaling),
```
acmorrow, line 99: This lights up the `Right` paths in api.rs.

micro-rdk/src/common/webrtc/signaling_server.rs
```diff
@@ -0,0 +1,176 @@
...
+        match self.sender.send(channel).await {
+            Ok(()) => {
+                let uuid: Option<String> = None;
+                let mut uuid_guard = scopeguard::guard(uuid, |uuid| {
+                    if let Some(uuid) = uuid {
+                        self.pending.lock().unwrap().remove(uuid.as_str());
+                    }
+                });
+
+                loop {
```
acmorrow, line 112: Here, we pull responses out of the channel as the api.rs machinery feeds them to us, and we stream them back out to the Sender that was passed in, which in turn results in a stream body up in grpc.rs.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -636,6 +639,10 @@ where
                 self.max_concurrent_connections as usize,
             ),
             robot_config: &config,
+            local_signaling_server: Arc::new(SignalingServer::new(
```
acmorrow, line 642: We need to adjust the default `max_concurrent_connections`, and probably also disable local signaling if the value we are provided is too small, indicating a very resource limited host.

acmorrow (commented):
Here are some comments, TODOs, and guideposts for this review. Some of these I will fold into comments before merging, others will become TODOs and tickets.

micro-rdk/src/common/webrtc/api.rs
gvaradarajan: why aren't we using a similar pattern to the `AnswerStage` structure? If we did, could we then differentiate between "done" and "should close"?

gvaradarajan (commented):
Just a first pass, still trying to understand how everything fits together before I dive deeper

micro-rdk/src/common/grpc.rs
```diff
@@ -1499,11 +1564,68 @@ where
...
+                            state.trailers.insert("grpc-status", e.status_code().into());
+                            state
+                                .trailers
+                                .insert("grpc-message", e.to_string().parse().unwrap());
+                        }
+                        None => {}
+                    }
+                };
+                Some((Ok(Frame::<Bytes>::trailers(state.trailers.clone())), state))
+            });
```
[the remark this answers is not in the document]
acmorrow: No, this is just a logic error. Thanks for catching it. It is supposed to end up returning `None` on the next invocation after trailers are sent. I will fix it up.

micro-rdk/src/common/webrtc/api.rs
acmorrow: The `Answer*` types have additional states that we don't have access to as a pure `Call` machine. There actually isn't anything we need to do here, except fix it to not close the channel. The channel close should happen for `call_update_request::Update::Done(_)`, instead.

micro-rdk/src/common/grpc.rs
```diff
@@ -125,27 +125,15 @@ pub trait GrpcResponse {
 
 #[derive(Clone)]
 pub struct GrpcServer<R> {
-    pub(crate) response: R,
-    pub(crate) buffer: Rc<RefCell<BytesMut>>,
+    pub(crate) _response: R,
```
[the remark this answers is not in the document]
acmorrow: Sure, done.

micro-rdk/src/common/grpc.rs
```diff
@@ -225,7 +210,20 @@ impl<'a> GrpcServerInner<'a> {
     }
 
     pub(crate) fn handle_request(
-        &mut self,
+        self,
+        path: &str,
+        payload: &[u8],
+    ) -> Pin<Box<dyn futures_lite::Stream<Item = Result<Bytes, ServerError>> + Sync + Send>> {
+        match path {
```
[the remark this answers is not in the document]
acmorrow: Yes, I plan to fold a lot of what I mentioned in this review into comments and `TODO(RSDK-XXXX)` with a filed ticket.

micro-rdk/src/common/grpc.rs
acmorrow: Do you mean the  `UnfoldState` struct, or the `Pin<Box<dyn Stream<Item=...>> + Sync + Send>` type? For the former, no, I'd prefer to keep the struct local to the function that uses it, since it doesn't serve a purpose anywhere else. For the future type, sure, that's fine by me.

micro-rdk/src/common/grpc.rs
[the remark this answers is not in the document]
acmorrow: I've updated the logic here, and it should be clearer now, and does explicitly return None to end the stream.

micro-rdk/src/common/grpc.rs
gvaradarajan: I did mean the former, but I see your point. I don't think defining the latter outside is useful (although I am surprised clippy didn't complain)

micro-rdk/src/common/webrtc/api.rs
```diff
@@ -299,47 +300,92 @@ impl WebRtcSignalingChannel {
...
-            uuid: sdp.uuid.clone(),
-            stage: Some(answer_response::Stage::Init(AnswerResponseInitStage {
-                sdp: answer,
-            })),
-        };
-        match self.signaling_tx.send_message(answer).await {
-            Err(e) => {
-                log::error!("error sending signaling message: {:?}", e);
-                Err(WebRtcError::SignalingDisconnected())
+        match self.signaling.as_mut() {
```
gvaradarajan, line 341: could some of this be simplified by a function like `fn send_answer_stage(&mut self, uuid: String, stage: AnswerStage) -> Result<(), WebRtcError>`?

gvaradarajan (commented):
Minor second pass, will do a final one after you've made all the changes you want to make from your own comments as well

micro-rdk/src/common/webrtc/signaling_server.rs
npmenard: we can add a ticket for making the engine global

npmenard (approved):
LGTM mainly some nits. We should merge and start testing it. 
I think further work should focus on removing code duplication in webrtc/api.rs and simplify both the rpc stream and the local signaling machinery so the flow is easier to understand. 
I would also like to have local signaling be a configurable option in the builder.

acmorrow:
Per discussion with @npmenard, I'm merging this in its current state, and I'll address comments and add ticketed TODOs in a near-term follow-up PR.

micro-rdk/src/common/webrtc/signaling_server.rs
acmorrow: Ticket added, and TODO'd: https://viam.atlassian.net/browse/RSDK-9247

micro-rdk/src/common/webrtc/signaling_server.rs
```diff
@@ -0,0 +1,179 @@
...
+        Ok(OptionalWebRtcConfigResponse { config: None })
+    }
+
+    pub async fn call(
+        &self,
+        request: CallRequest,
+        responses: Sender<Result<CallResponse, ServerError>>,
+    ) -> Result<(), ServerError> {
+        if request.disable_trickle {
+            return Err(ServerError::new(GrpcError::RpcUnimplemented, None));
```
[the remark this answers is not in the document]
acmorrow: Done

micro-rdk/src/common/webrtc/signaling_server.rs
```diff
@@ -0,0 +1,179 @@
...
+                    }
+                });
+
+                loop {
+                    match response_rx.recv().await {
+                        Ok(Ok(response)) => {
+                            match uuid_guard.as_ref() {
+                                Some(uuid) => {
+                                    if uuid != &response.uuid {
+                                        break Err(ServerError::new(GrpcError::RpcInternal, None));
```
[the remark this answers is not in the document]
acmorrow: Done

acmorrow (commented):
Follow-ups on open comments. I will create a new PR with changes where they were made in response to notes here.

acmorrow:
Changes in response to comments are in #349

PR #348, opened 2024-11-11
[RSDK-9216] - allocate buffer for response message per RPC call

gvaradarajan:


gvaradarajan:
still need to test on an actual ESP32 with camera

micro-rdk/src/common/grpc.rs
```diff
@@ -836,7 +822,9 @@ impl<'a> GrpcServerInner<'a> {
             .get_generic_readings()
             .map_err(|err| ServerError::new(GrpcError::RpcInternal, Some(err.into())))?;
         let resp = proto::common::v1::GetReadingsResponse { readings };
-        self.encode_message(resp)
+        let res = GrpcServerInner::get_encoded_message(resp);
+        println!("res: {:?}", res);
```
acmorrow, line 826: Looks like a stale debugging print here.

micro-rdk/src/common/grpc.rs
```diff
@@ -154,28 +148,20 @@ where
 }
 
 impl<'a> GrpcServerInner<'a> {
-    fn encode_message_into<M: Message>(m: M, mut buffer: BytesMut) -> Result<Bytes, ServerError> {
+    fn get_encoded_message<M: Message>(m: M) -> Result<Bytes, ServerError> {
```
acmorrow, line 151: I think I like the old name better. This method isn't getting an encoded message from somewhere, it is taking a message and encoding it.

Since you no longer need the "new buffer" vs "internal buffer" distinction with the three existing methods were intended to support and now only need the former, you could just call this method `encode_message`.

micro-rdk/src/common/camera/fake_camera.rs
```diff
@@ -130,25 +111,26 @@ mod tests {
...
+        let listener: async_io::Async<TcpListener> = tcp_server.try_into().unwrap();
 
         let robot = Arc::new(Mutex::new(setup_robot().unwrap()));
 
         loop {
             let incoming = listener.accept().await;
             assert!(incoming.is_ok());
-            let stream = incoming.unwrap();
+            let incoming = incoming.unwrap();
+            let stream: NativeStream = NativeStream::LocalPlain(incoming.0);
```
acmorrow, line 133: What is the context for these changes? They seem somewhat unrelated to the rest?

micro-rdk/src/esp32/camera.rs
```diff
@@ -173,36 +173,9 @@ impl Esp32Camera {
...
-        let msg = HttpBody {
-            content_type: "image/jpeg".to_string(),
-            data: frame.as_bytes().to_vec(),
-            ..Default::default()
-        };
-        // safety: message must be encoded before the frame is dropped from scope
-        msg.encode(&mut buffer)
-            .map_err(CameraError::MessageEncodeError)?;
-        return Ok(buffer);
+        Ok(frame.as_bytes())
```
acmorrow, line 178: Does this allocate?

acmorrow (changes requested):

micro-rdk/src/common/camera/fake_camera.rs
gvaradarajan: the camera tests don't pass without it, ever since [this commit](https://github.com/viamrobotics/micro-rdk/commit/b4f543140e2bb919d7294c97352f35d69b27dbed)

mattjperez (approved):
Pretty floored that the solution was so straight forward. 
LGTM, verified worki with camera

acmorrow (approved):
LGTM

PR #350, opened 2024-11-14
[RSDK-9195] ota async task

mattjperez:
The following PR:
- [RSDK-9195](https://viam.atlassian.net/browse/RSDK-9195?atlOrigin=eyJpIjoiOGMxMmZmOTAxNWUxNDM4MWI4YjZjMGIzMzRjYTczZTMiLCJwIjoiaiJ9): moves the ota logic to an async task attached to viam-server (this will enable retries/incremental-download/etc)
- [RSDK-9198](https://viam.atlassian.net/browse/RSDK-9198?atlOrigin=eyJpIjoiNGQ1NTgxNTU0NmFiNGUyOWI4YjIxYTM3MGUxYWQ4MmMiLCJwIjoiaiJ9): uses async hyper::http2 instead of the blocking EspHttpClient

The errors will get handled better as part of [RSDK-9214](https://viam.atlassian.net/browse/RSDK-9214?atlOrigin=eyJpIjoiNzhmMmRkZDNiMGEwNGJlZGE2YjkyZDcxOWQxNjkyMzQiLCJwIjoiaiJ9)

[RSDK-9195]: https://viam.atlassian.net/browse/RSDK-9195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSDK-9198]: https://viam.atlassian.net/browse/RSDK-9198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSDK-9214]: https://viam.atlassian.net/browse/RSDK-9214?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


micro-rdk/Cargo.toml
```diff
@@ -86,6 +86,8 @@ pin-project.workspace = true
 async-lock.workspace = true
 dns-message-parser.workspace = true
 printf-compat = { workspace = true, optional = true }
+bincode = "2.0.0-rc.3"
+
```
mattjperez, line 90: will remove

mattjperez, line 90: I can make this an optional dependency for `ota` feature moving forward. That way the changes it doesn't get into the regular build

will rm ws

micro-rdk/src/common/conn/viam.rs
```diff
@@ -622,6 +600,42 @@ where
             }
         }
 
+        #[cfg(feature = "ota")]
```
mattjperez, line 603: moved the whole ota section to after certs have been retrieved and stored

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+            parts.authority = Some(
+                auth.parse()
+                    .map_err(|_| OtaError::Other("failed to parse authority".to_string()))?,
+            );
+            uri = hyper::Uri::from_parts(parts).map_err(|e| OtaError::Other(e.to_string()))?;
+        };
+        let io = self
+            .connector
+            .connect_to(&uri)
+            .unwrap()
```
mattjperez, line 199: will change to map_err in follow-up commit

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+                if total_downloaded < core::mem::size_of::<EspAppDesc>() {
+                    log::error!("initial frame too small to retrieve esp_app_desc_t");
+                } else {
+                    log::info!("data length {}", data.len());
+                    #[cfg(feature = "esp32")]
+                    {
+                        let mut loader = EspFirmwareInfoLoader::new();
+                        loader.load(&data).map_err(OtaError::EspError)?;
+                        let new_fw = loader.get_info().map_err(OtaError::EspError)?;
+                        log::info!("current firmware: {:?}", running_fw_info);
```
gvaradarajan, line 286: `log::debug` instead (for both statements?

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+            .url
+            .parse::<hyper::Uri>()
+            .map_err(|_| OtaError::ConfigError(format!("invalid url: {}", self.url)))?;
+        if uri.port().is_none() {
+            if uri.scheme_str() != Some("https") {
+                log::error!("no port found and not https");
+            }
+
+            let mut auth = uri.authority().expect("str").to_string();
+            auth.push_str(":443");
```
gvaradarajan, line 188: Does this represent a port? What determines this value?

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+            .connector
+            .connect_to(&uri)
+            .unwrap()
+            .await
+            .map_err(|e| OtaError::Other(e.to_string()))?;
+
+        let (mut sender, conn) = {
+            http2::Builder::new(self.exec.clone())
+                .keep_alive_interval(Some(std::time::Duration::from_secs(120)))
+                .keep_alive_timeout(std::time::Duration::from_secs(10))
```
gvaradarajan, line 206: Is the minimum amount of time we can use for this timeout? If a smaller timeout works, I think it would be preferred

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+            log::error!("aborting ota");
+            #[cfg(feature = "esp32")]
+            update_handle.abort()?;
+
+            return Err(OtaError::Other(
+                "nbytes written did not match file size".to_string(),
+            ));
+        }
+
+        drop(conn);
```
gvaradarajan, line 347: could we drop connection before aborting the OTA? It seems to me that we'd like to keep the connection alive for as little time as possible and the abort might take a variable amount of time

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+    InitError,
+    #[error("new image size is invalid: {0} bytes")]
+    InvalidImageSize(usize),
+    #[error("{0}")]
+    Other(String),
+}
+
+#[cfg(feature = "esp32")]
+type OtaConnector = crate::esp32::tcp::Esp32H2Connector;
+#[cfg(not(feature = "esp32"))]
```
gvaradarajan, line 123: maybe I forgot from the previous PR, but why are we allowing cases where we're not on an esp32? Isn't OTA meaningless otherwise?

micro-rdk/src/common/ota.rs
mattjperez: this represents a port. This is specifically in the case that the scheme is `https` and no port is specified. `443` is the known port for https (connection attempt will fail without it). hyper doesn't make any assumptions for you so you need to handle the case that a port isn't included.

If you include a port in the url (ex `https://my_url.com:9000`) it will use that port instead.

micro-rdk/src/common/ota.rs
mattjperez: sure, I can move this up a bit to before the `nwritten != file_len` check.

micro-rdk/src/common/ota.rs
mattjperez: is there a lower limit you have in mind?

micro-rdk/src/common/ota.rs
mattjperez: testing, specifically around hyper, config checks, and inspecting the binary to be downloaded

micro-rdk/src/common/ota.rs
mattjperez: will do

micro-rdk/src/common/ota.rs
gvaradarajan: no, I figured we could arrive at this by lowering the duration until it no longer works on a relatively unstable network (probably do something more like binary search)

micro-rdk/src/common/ota.rs
gvaradarajan: That's fair for development purposes maybe, but isn't this confusing for merged-in code? It took me a while to see that OTA compiles for native but makes http calls that essentially do nothing.

mattjperez: Not sure what you mean by nothing. The calls are for downloading the OTA binary and inspecting it. This would also be useful for  verifying future work we talked about for editing the binary signatures directly. 

Should I put it back and take it out between every PR for this feature? Or clean it up at the end if it doesn't prove useful beyond the planned work?

mattjperez: I guess to that point, what is the purpose of having a native build?

gvaradarajan: I mean the native build still tests the connection infrastructure abstraction and acts as a server. In native "OTA", there's no updating happening so it doesn't feel like a fair comparison. But we can leave it in and revisit at the end of the epic

micro-rdk/src/common/ota.rs
gvaradarajan: ok, sounds good

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+    /// compile date
+    date: [u8; 16],
+    idf_ver: [u8; 32],
+    app_elf_sha256: [u8; 32],
+    /// minimal eFuse block revision supported by image, in format: major * 100 + minor
+    min_efuse_blk_rev_full: u16,
+    /// maximal eFuse block revision supported by image, in format: major * 100 + minor
+    max_efuse_blk_rev_full: u16,
+    reserv2: [u32; 19],
+}
```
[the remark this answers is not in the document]
acmorrow: This is fine for now, let's make a ticket to revisit. I'd prefer we ultimately use the C FFI one when possible, and declare our own for native. Ideally with some sanity checks on esp32 that the native one seems to line up with the FFI one.

micro-rdk/src/common/ota.rs
acmorrow: Is there a reason that we know we need to adjust these from the defaults?

micro-rdk/src/common/ota.rs
acmorrow: Overall though, I do feel like this method should probably have some well thought out user facing logs that provide context about what the OTA system is doing.

micro-rdk/src/common/conn/viam.rs
acmorrow: Why do we need certs? I thought the certs were for us _serving_

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+};
+use bincode::Decode;
+use futures_lite::AsyncWriteExt;
+use http_body_util::{BodyExt, Empty};
+use hyper::{body::Bytes, client::conn::http2, Request};
+use thiserror::Error;
+
+// TODO(RSDK-9200): set according to active partition scheme
+const OTA_MAX_IMAGE_SIZE: usize = 1024 * 1024 * 4; // 4MB
+pub const OTA_MODEL_TYPE: &str = "ota_service";
```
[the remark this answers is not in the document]
mattjperez: I tried doing that initially, but was having issues with using a const str to define another const str, will look into it again.

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+                    #[cfg(feature = "esp32")]
+                    {
+                        let mut loader = EspFirmwareInfoLoader::new();
+                        loader.load(&data).map_err(OtaError::EspError)?;
+                        let new_fw = loader.get_info().map_err(OtaError::EspError)?;
+                        log::info!("current firmware: {:?}", running_fw_info);
+                        log::info!("new firmware: {:?}", new_fw);
+                        if let Some(ref running_fw) = running_fw_info {
+                            if running_fw.version == new_fw.version
+                                && running_fw.released == new_fw.released
```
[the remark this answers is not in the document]
mattjperez: this is the compiled timestamp we talked about before

micro-rdk/src/common/conn/viam.rs
mattjperez: oh, I am using the same certs for the hyper client's https connection.  should I not be?

acmorrow: Let's try the empirical answer: does it still work if you don't use them?

micro-rdk/src/common/conn/viam.rs
```diff
@@ -622,6 +600,42 @@ where
...
+                    .services
+                    .iter()
+                    .find(|&service| service.model == ota::OTA_MODEL_TRIPLET)
+                {
+                    if let Ok(mut ota) =
+                        ota::OtaService::from_config(service, certs, self.executor.clone())
+                    {
+                        self.ota_service_task
+                            .replace(self.executor.spawn(async move {
+                                if let Err(e) = ota.update().await {
```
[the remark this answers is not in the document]
mattjperez: part of why we made this a separate async task was so we can do backoff and retries while still letting the rest of the program move forward, but the idea was for the task to eventually terminate.

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+        let mut uri = self
+            .url
+            .parse::<hyper::Uri>()
+            .map_err(|_| OtaError::ConfigError(format!("invalid url: {}", self.url)))?;
+        if uri.port().is_none() {
+            if uri.scheme_str() != Some("https") {
+                log::error!("no port found and not https");
+            }
+
+            let mut auth = uri.authority().expect("str").to_string();
```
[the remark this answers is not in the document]
mattjperez: will map the error

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+        };
+        #[cfg(feature = "esp32")]
+        let mut update_handle = ota.initiate_update().map_err(|_| OtaError::InitError)?;
+        #[cfg(not(feature = "esp32"))]
+        let mut update_handle = Vec::new();
+        let mut nwritten: usize = 0;
+        let mut total_downloaded: usize = 0;
+        let mut got_info = false;
+
+        while let Some(next) = response.frame().await {
```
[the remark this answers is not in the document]
mattjperez: it appears the closest thing would be the `range` [header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range), to request a specific range of bytes to be served (could be useful for resumable download).

However, there's no guarantee that the server supports ranges and that would have to be [verified](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range#checking_if_a_server_supports_range_requests) and handled.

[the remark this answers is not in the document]
acmorrow: - https://httpwg.org/specs/rfc7540.html#SETTINGS_MAX_FRAME_SIZE
- https://docs.rs/h2/latest/h2/server/struct.Builder.html#method.max_frame_size
- https://docs.rs/h2/latest/h2/client/struct.Builder.html#method.max_frame_size

micro-rdk/src/common/ota.rs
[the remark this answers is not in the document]
mattjperez: [RSDK-9342](https://viam.atlassian.net/browse/RSDK-9342?atlOrigin=eyJpIjoiZDgzNGNlNTlmOTlhNGJlYTk3OTNiMjE3YzcwM2I3M2IiLCJwIjoiaiJ9)

micro-rdk/src/common/conn/viam.rs
mattjperez: you were right. we just need the certificate authority, which the H2 connector is already using. removed the certificates completely.

micro-rdk/src/common/ota.rs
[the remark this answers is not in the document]
mattjperez: accomplished using [once_cell::sync::Lazy](https://docs.rs/once_cell/latest/once_cell/sync/struct.Lazy.html), though not a fan of the deref, but it works.

micro-rdk/src/common/ota.rs
mattjperez: confirmed working without setting `keep_alive_interval` and using default timeout (20 seconds).

micro-rdk/Cargo.toml
mattjperez: moved to just being compiled when target isn't esp32 for now. the [ticket](https://viam.atlassian.net/browse/RSDK-9342?atlOrigin=eyJpIjoiZDgzNGNlNTlmOTlhNGJlYTk3OTNiMjE3YzcwM2I3M2IiLCJwIjoiaiJ9) for having a native representation for the app description might use a different crate eventually so defering until then.

micro-rdk/src/common/grpc_client.rs
```diff
@@ -219,7 +219,7 @@ impl Future for AsyncioSleep {
 }
 
 #[derive(Default, Clone, Debug)]
-struct H2Timer;
+pub struct H2Timer;
```
gvaradarajan, line 222: `pub(crate)` instead?

gvaradarajan (approved):
LGTM

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,357 @@
...
+            .method("GET")
+            .uri(uri)
+            .body(Empty::<Bytes>::new())
+            .map_err(|e| OtaError::Other(e.to_string()))?;
+        let mut response = sender
+            .send_request(request)
+            .await
+            .map_err(|e| OtaError::Other(e.to_string()))?;
+
+        if response.status() != 200 {
```
acmorrow, line 226: Please add a TODO here to use `is_success` or similar.

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,357 @@
...
+            return Err(OtaError::Other(
+                format!("Bad Request - Status:{}", response.status()).to_string(),
+            ));
+        }
+        let headers = response.headers();
+        log::debug!("ota response headers: {:?}", headers);
+
+        if !headers.contains_key(hyper::header::CONTENT_LENGTH) {
+            log::error!("header");
```
acmorrow, line 235: This log statement should either be elaborated or removed.

acmorrow (approved):
LGTM with a few small nits; I think this is in good enough shape we should merge it so we can continue iterating and testing.

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,359 @@
...
+                            }
+                        }
+                    }
+                    #[cfg(not(feature = "esp32"))]
+                    {
+                        if let Ok(decoded) = bincode::decode_from_slice::<
+                            EspAppDesc,
+                            bincode::config::Configuration,
+                        >(
+                            &data[..256], bincode::config::standard()
```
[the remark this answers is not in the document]
mattjperez: changed to const since `EspAppDesc` and `esp_app_desc_t` are on different targets.

micro-rdk/src/common/ota.rs
mattjperez: neither hyper nor http crates define the https default port, I'm guessing because these crates try to minimize assumptions. we could look at another dependency, but I'd rather leave this as-is and add a comment.

note: `reqwest` handles it in-place, [link](https://github.com/search?q=repo%3Aseanmonstar%2Freqwest%20443&type=code)

micro-rdk/src/common/ota.rs
```diff
@@ -0,0 +1,357 @@
...
+        })
+    }
+
+    pub(crate) async fn update(&mut self) -> Result<(), OtaError> {
+        let mut uri = self
+            .url
+            .parse::<hyper::Uri>()
+            .map_err(|_| OtaError::ConfigError(format!("invalid url: {}", self.url)))?;
+        if uri.port().is_none() {
+            if uri.scheme_str() != Some("https") {
```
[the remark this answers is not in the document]
mattjperez: interestingly, hyper doesn't appear to expose `http::uri::scheme`, just `http::uri::URI`, so we don't have direct access to `Scheme` without importing `http` explicitly.

We actually do the current pattern [in micro-rdk already](https://github.com/viamrobotics/micro-rdk/blob/c2a2baf587194e4fbded85689afa2c1979f7439e/micro-rdk/src/native/tcp.rs#L65). So we could define the strings and default ports if we wanted to

micro-rdk/src/common/ota.rs
mattjperez: removed

micro-rdk/src/common/ota.rs
mattjperez: [RSDK-9346](https://viam.atlassian.net/browse/RSDK-9346?atlOrigin=eyJpIjoiODBkYzA4MzAzNmUwNDk0Yzg5NGYyZmI0MTM1MGI0MjYiLCJwIjoiaiJ9)

micro-rdk/src/common/grpc_client.rs
mattjperez: added

PR #351, opened 2024-11-14
[RSDK-9213] restart on fd error

mattjperez:


acmorrow:
The ticket says that we should either raise the fd limit, or restart when out of fds.
- Did we investigate raising the fd limit? If so, what were the results? Related: did we check whether there is any sort of sdkconfig that makes file descrriptor exhaustion a crash?
- This only handles fd exhaustion in the app signaling path, but that isn't the only place we create files (sockets). Is there something particularly unique about failing to obtain an fd in the app signaling path that leads to a hang such that it requires special handling, but that we can get away with not doing elsewhere?

mattjperez:
> The ticket says that we should either raise the fd limit, or restart when out of fds.
> 
>     * Did we investigate raising the fd limit? If so, what were the results? Related: did we check whether there is any sort of sdkconfig that makes file descrriptor exhaustion a crash?
> 
>     * This only handles fd exhaustion in the app signaling path, but that isn't the only place we create files (sockets). Is there something particularly unique about failing to obtain an fd in the app signaling path that leads to a hang such that it requires special handling, but that we can get away with not doing elsewhere?

There is an option in [sdkconfig](https://docs.espressif.com/projects/esp-idf/en/v4.4.4/esp32/api-reference/kconfig.html?highlight=sdkconfig#config-fatfs-fs-lock), however we don't appear to be setting it yet and I don't see any sane default set in the documentation, might need to dive into source. We could choose to set it, but we would need to make sure we do that in all the sdkconfigs **and** ensure whoever uses our ffi has these settings somehow (maybe an option in esp-idf component options). I didn't see any direct way to set behavior when fds are exhausted

As for handling outside of signaling:
- we could write a general wrapper/macro for all EspErrors, that way they can be identified at any point throughout the stack
-  signaling happens fairly regularly, no? as the exhaustion would be system-wide, should another part of the stack experience this error first, signaling may eventually run into the same issue, triggering the restart anyway.

acmorrow:
> UPDATE: looking at [esp-idf](https://github.com/espressif/esp-idf/blob/7a305c0284b7af7cd8b8f12b48f72e2685d9a363/components/fatfs/Kconfig#L141), the default is 0 (no limit).

I don't think that configuration parameter controls the number of FDs - I think it is specifically about protecting FAT files from corruption due to simultaneous operations.

acmorrow:
@mattjperez - Let's close this PR, I will put up another with what I believe is a fix for the root cause.

PR #353, opened 2024-11-25
[RSDK-9283] add ota metadata storage, compare stored and latest

mattjperez:
Instead of storing the OTA ServiceConfig in storage like originally planned, using the cached robot config achieves the same effect.

Still testing on hardware, will update when verified

### DIFF
--- a/micro-rdk/src/common/conn/viam.rs
+++ b/micro-rdk/src/common/conn/viam.rs
@@ -51,10 +51,7 @@ use super::server::{IncomingConnectionManager, WebRtcConfiguration};
 use crate::common::provisioning::server::AsNetwork;
 
 #[cfg(feature = "ota")]
-use crate::{
-    common::{credentials_storage::OtaMetadataStorage, ota},
-    proto::app::v1::ServiceConfig,
-};
+use crate::common::{credentials_storage::OtaMetadataStorage, ota};
 
 pub struct RobotCloudConfig {
     local_fqdn: String,
--- a/micro-rdk/src/common/conn/viam.rs
+++ b/micro-rdk/src/common/conn/viam.rs
@@ -542,7 +542,7 @@ where
                 .iter()
                 .find(|&service| service.model == *ota::OTA_MODEL_TRIPLET)
             {
-                match ota::OtaService::from_config_boxed(
+                match ota::OtaService::from_config(
                     service,
                     self.storage.clone(),
                     self.executor.clone(),
--- a/micro-rdk/src/common/conn/viam.rs
+++ b/micro-rdk/src/common/conn/viam.rs
@@ -529,6 +529,13 @@ where
             log::error!("couldn't store the robot configuration reason {:?}", err);
         }
 
+        let config_monitor_task = Box::new(ConfigMonitor::new(
+            config.clone(),
+            self.storage.clone(),
+            || std::process::exit(0),
+        ));
+        self.app_client_tasks.push(config_monitor_task);
+
         #[cfg(feature = "ota")]
         {
             log::debug!("ota feature enabled");
@@ -559,13 +566,6 @@ where
                 );
             }
         }
-
-        let config_monitor_task = Box::new(ConfigMonitor::new(
-            config.clone(),
-            self.storage.clone(),
-            || std::process::exit(0),
-        ));
-        self.app_client_tasks.push(config_monitor_task);
 
         let mut robot = LocalRobot::from_cloud_config(
             self.executor.clone(),
--- a/micro-rdk/src/common/conn/viam.rs
+++ b/micro-rdk/src/common/conn/viam.rs
@@ -506,20 +506,9 @@ where
             |resp| (resp.0.config.map_or(Box::default(), Box::new), resp.1),
         );
 
-        if let Err(err) = self.storage.store_robot_configuration(&config) {
-            log::error!("couldn't store the robot configuration reason {:?}", err);
-        }
-
-        let config_monitor_task = Box::new(ConfigMonitor::new(
-            config.clone(),
-            self.storage.clone(),
-            || std::process::exit(0),
-        ));
-        self.app_client_tasks.push(config_monitor_task);
-
         #[cfg(feature = "ota")]
         {
-            use crate::common::ota;
+            use crate::{common::ota, proto::app::v1::ServiceConfig};
             log::debug!("ota feature enabled");
 
             if let Some(service) = config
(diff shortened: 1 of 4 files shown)


mattjperez:
this PR needs more work, looks like I might need to go back to the NVS partition for OtaMetadata.

Because the robot config (source of truth in this case) gets updated in the viam-server *while* ota is being done, an unrecoverable crash in OTA (ex a stack overflow) will not prevent the robot config from getting updated.

TLDR: an unrecoverable crash in OTA will lead to the program thinking is up-to-date when it actually is not

micro-rdk/src/common/conn/viam.rs
```diff
@@ -86,15 +89,28 @@ impl From<&proto::app::v1::CloudConfig> for RobotCloudConfig {
...
+#[cfg(feature = "ota")]
+pub trait ViamServerStorage:
+    RobotConfigurationStorage + WifiCredentialStorage + OtaMetadataStorage + Clone + 'static
+{
+}
+#[cfg(feature = "ota")]
+impl<T> ViamServerStorage for T where
+    T: RobotConfigurationStorage + WifiCredentialStorage + OtaMetadataStorage + Clone + 'static
+{
+}
```
mattjperez, line 112: I'm not a fan of how this looks. Open to suggestions

micro-rdk/src/common/conn/viam.rs
```diff
@@ -86,15 +89,28 @@ impl From<&proto::app::v1::CloudConfig> for RobotCloudConfig {
...
+
```
mattjperez, line 113: not a fan of how this looks, open to suggestions.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -569,6 +579,13 @@ where
             }
         }
 
+        let config_monitor_task = Box::new(ConfigMonitor::new(
+            config.clone(),
+            self.storage.clone(),
+            || std::process::exit(0),
+        ));
+        self.app_client_tasks.push(config_monitor_task);
+
```
mattjperez, line 588: this is just code movement. can shift back to before ota

micro-rdk/src/common/ota.rs
```diff
@@ -347,11 +394,25 @@ impl OtaService {
...
+        }
+
+        log::info!("updating metadata in NVS");
+        self.storage
+            .store_ota_metadata(OtaMetadata {
+                version: self.pending_version.clone()
+            })
+            .map_err(|e| OtaError::Other(e.to_string()))?;
+
+        log::info!("firmware update complete");
```
mattjperez, line 408: @acmorrow @gvaradarajan 

in this section, there's the possibility we succeed in setting the next boot partition (`update_handle.complete()`) but fail to update NVS (`store_ota_metadata()`).

In this scenario, the device will boot from the new firmware but, given the old nvs metadata is still intact/corrupted/non-existant, will think it is out of date and attempt to update again (download firmware, set boot partition, update nvs).

This is similar to the initial boot-up anyway since there is no OtaMetadata in the NVS on a clean build.

It seems preferable to the alternative if we swap the order of updates; the device will (falsely) believeit is up-to-date and will not perform the update.

mattjperez (commented):
@acmorrow @gvaradarajan I've updated the PR to use NVS storage again. 

Still testing on hardware and will update this when done.

micro-rdk/src/common/conn/viam.rs
gvaradarajan: curious, what bothers you about it?

gvaradarajan (approved):
LGTM

acmorrow:
FYI @mattjperez I think you have a transposed digit in the ticket in the commit message / PR title.

micro-rdk/src/common/conn/viam.rs
mattjperez: just having 4 things that look very similar and all the feature-gating. nothing functional

acmorrow (changes requested):

micro-rdk/src/common/ota.rs
```diff
@@ -281,18 +337,8 @@ impl OtaService {
...
-                            if running_fw.version == new_fw.version
-                                && running_fw.released == new_fw.released
-                            {
-                                log::info!("current firmware is up to date");
-                                update_handle.abort()?;
-                                return Ok(());
-                            }
-                        }
-                        log::info!("applying new ota firmware");
+                        log::debug!("current firmware app description: {:?}", running_fw_info);
```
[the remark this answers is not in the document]
mattjperez: not at the moment. if anything only `loader.load(&data).map_err(OtaError::EspError)` has any functional purpose in that it acts as verification that this firmware actually has a header (and therefore is valid firmware and not a random binary)

micro-rdk/src/common/ota.rs
```diff
@@ -332,7 +379,7 @@ impl OtaService {
         }
 
         drop(conn);
-        log::info!("ota download complete");
+        log::info!("download complete");
```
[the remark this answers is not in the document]
mattjperez: handled in the dynamic max size PR.

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -135,6 +135,29 @@ const NVS_WIFI_PASSWORD_KEY: &str = "WIFI_PASSWORD";
...
+    type Error = NVSStorageError;
+    fn has_ota_metadata(&self) -> bool {
+        self.has_string(NVS_OTA_VERSION_KEY).unwrap_or(false)
+    }
+    fn get_ota_metadata(&self) -> Result<OtaMetadata, Self::Error> {
+        let version = self.get_string(NVS_OTA_VERSION_KEY)?;
+        Ok(OtaMetadata { version })
+    }
+    fn store_ota_metadata(&self, ota_metadata: OtaMetadata) -> Result<(), Self::Error> {
+        self.set_string(NVS_OTA_VERSION_KEY, &ota_metadata.version)
```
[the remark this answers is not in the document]
mattjperez: after offline talk, going with 128 char limit, failing to initialize ota and logging descriptive error

acmorrow (approved):
LGTM

micro-rdk/src/common/ota.rs
```diff
@@ -159,19 +181,79 @@ impl OtaService {
...
+            return Ok(());
+        }
+
+        log::info!(
+            "proceeding with update from version `{}` to version `{}`",
+            stored_metadata.version,
+            self.pending_version
+        );
+
+        let mut uri = Box::new(
```
acmorrow, line 251: Instead of Box'ing these various bits of ephemeral state, could we try Box'ing the `OtaService` itself? It seems to indirectly hold cryptographic material which may be making it fairly hefty to hold on the stack.

mattjperez: Tried boxing the `OtaService` itself and still had the stack overflow. 
Boxing the spawned async handle to `conn` however is working consistently.

PR #354, opened 2024-11-26
[RSDK-8992] Upgrade to rust 1.82

acmorrow:
Rust 1.82 is the latest that esp-rs offers.


.cargo/config.toml
```diff
@@ -1,23 +1,23 @@
 [target.xtensa-esp32-espidf]
 linker = "ldproxy"
 runner = "espflash --monitor"
-rustflags = ["-C", "default-linker-libraries", "-Clink-args=-Wl,-Map=./esp-build.map"]
+rustflags = ["-C", "default-linker-libraries", "-Clink-args=-Wl,-Map=./esp-build.map", "--cfg", "espidf_time32"]
```
acmorrow, line 4: Required due to `libc` crate upgrade and https://github.com/rust-lang/libc/pull/3993.

Cargo.toml
```diff
@@ -39,7 +39,7 @@ debug = true # Symbols are nice and they don't increase the size on Flash
 opt-level = "s"
 
 [workspace.dependencies]
-anyhow = "1.0.89"
+anyhow = "1.0.93"
```
acmorrow, line 42: Updates to this section per `cargo upgrade` with rust 1.82 enabled.

Cargo.toml
```diff
@@ -49,12 +49,12 @@ atomic-waker = "1.1.2"
...
 bytecodec = "0.4.15"
-bytes = "1.7.2"
-cargo_metadata = "0.18.1"
+bytes = "1.8.0"
+cargo_metadata = "0.19.0"
 cbindgen = "0.27.0"
 chrono = "0.4.38"
 chrono-tz = "0.10.0"
-clap = "4.5.20"
+clap = "4.5.21"
```
acmorrow, line 57: ⬇️ - I had hoped to upgrade `embedded-svc` and `esp-idf-svc`, but did not succeed due to a puzzling error. I'll take another go at it, since I believe all that was blocking it was the rust version.

micro-rdk-ffi/build.rs
```diff
@@ -18,8 +18,10 @@ pub struct Cloud {
 }
 
 fn main() {
-    #[cfg(not(feature = "test"))]
+    #[cfg(not(test))]
     {
+        println!("cargo:rustc-check-cfg=cfg(has_robot_config)");
```
acmorrow, line 23: Without this, newer clippy yelled at me about `cfg(has_robot_config)` as an unknown config check.

micro-rdk-ffi/build.rs
```diff
@@ -18,8 +18,10 @@ pub struct Cloud {
 }
 
 fn main() {
-    #[cfg(not(feature = "test"))]
+    #[cfg(not(test))]
```
acmorrow, line 21: I think this was an error before? I'm not sure.

micro-rdk/build.rs
```diff
@@ -1,6 +1,7 @@
 use regex::Regex;
 
 fn main() {
+    println!("cargo::rustc-check-cfg=cfg(esp32)");
```
acmorrow, line 4: Like above, clippy would complain that this was an unknown config in esp32/board.rs.

micro-rdk/src/esp32/board.rs
```diff
@@ -216,12 +216,9 @@ impl EspBoard {
                 vec![]
             };
 
-            let i2c_confs = if let Ok(i2c_confs) = cfg.get_attribute::<Vec<Esp32I2cConfig>>("i2cs")
-            {
-                i2c_confs
-            } else {
-                vec![]
-            };
+            let i2c_confs = cfg
```
acmorrow, line 219: Newer clippy suggestion

micro-rdk/src/esp32/coredump.rs
```diff
@@ -115,7 +115,7 @@ impl DoCommand for Coredump {
         command_struct: Option<protobuf::Struct>,
     ) -> Result<Option<protobuf::Struct>, crate::common::generic::GenericError> {
         if let Some(cmd) = command_struct {
-            if cmd.fields.get("sizes").is_some() {
+            if cmd.fields.contains_key("sizes") {
```
acmorrow, line 118: This and ⬇️ are from newer clippy suggestions

acmorrow (commented):
Some notes. Also, need to think about how to get this green in CI because right now it fails in anything using the docker image since they haven't been updated to 1.82, but we would need to merge this in order to build new docker images that had 1.82, so it is a bit of a chicken/egg thing.

mattjperez (approved):
LGTM

acmorrow:
FYI I'm not planning to merge this until next week

gvaradarajan (approved):
LGTM

acmorrow:
@mattjperez and @gvaradarajan - I held back on merging this due to the license finder failures.

@stevebriskin - Please have a look at the license finder failure on this if you have a minute. Is Unicode-3 something we can add to our accepted licenses, should we need to? I also intend to check whether we can actually disable the dependency edge that has dragged in the relevant unicode stuff, as it all appears to be [IDNA](https://en.wikipedia.org/wiki/Internationalized_domain_name) related and I suspect we can live without that.

acmorrow:
Discussed with @stevebriskin today and got approval to add Unicode-3.0 as an approved license, pushed a commit to do so per `license_finder permitted_licenses add Unicode-3.0`

acmorrow (commented):
@gvaradarajan and @mattjperez - Well, rust 1.83 came out while we were iterating on this with the license finder stuff. So, here is 1.83 supported. Please take a look, as I'm hoping to merge this either later tonight or first thing monday morning.

acmorrow:
Here's just the new commits on this: https://github.com/viamrobotics/micro-rdk/pull/354/files/1d3de65fd6e33ada99562ed4fabccec84497ac0f..52a210d50b6492f1a297fb72adf7a0e24cb4dea1

micro-rdk/src/common/provisioning/server.rs
```diff
@@ -522,7 +522,7 @@ pub trait WifiManager: Network {
         &'a self,
         ssid: &'a str,
         password: &'a str,
-    ) -> Pin<Box<dyn Future<Output = Result<(), WifiManagerError>> + '_>>;
```
mattjperez, line 525: interesting that this inverse change is needed opposed to the other lifetime changes. I guess because it's the returned result and not the input parameter.

mattjperez (approved):
LGTM

PR #355, opened 2024-12-04
[RSDK-9213] Ensure ice agent is terminated if signaling doesn't complete

acmorrow:
Inside `run_ice_until_connected` a detached task is spawned for `ICEAgent::run`. That task can be canceled by calling `done` on the associated `AtomicSync` (`WebRtcApi::ice_agent`). In normal operation, that happens when the `AtomicSync` is stashed into a created `WebRTCConnection` which does indeed call `done` on the `AtomicSync` in its `Drop` implementation.

However, if `WebRtcApi::connect` launches the `ICEAgent::run` task in `run_ice_until_connected`, but does not proceed far enough to create the `WebRTCConnection`, then the detached task will never stop.

This exact scenario can happen if a client issues two WebRTC connection requests, one via app.viam.com signaling and the other via local signaling, and one of them gets cancelled during the critical window. It could also happen if a client crashes or is ^C'd in the middle of signaling.

I've written a not great fix for this in the interest time and minimal change to the complex code in `api.rs`: it simply creates a scope guard that ensures that `done` is called on the `AtomicSync` if no `WebRTCConnection` object is produced to take over that responsibility.

### DIFF
--- a/micro-rdk/src/common/webrtc/api.rs
+++ b/micro-rdk/src/common/webrtc/api.rs
@@ -683,9 +683,7 @@ where
         // Make sure that if we leave this scope without passing
         // ownership of the `ice_agent` notifier to a new
         // WebRTCConnection that the ICEAgent is terminated.
-        let ice_done_guard = scopeguard::guard(
-            self.ice_agent.clone(), |ia| ia.done(),
-        );
+        let ice_done_guard = scopeguard::guard(self.ice_agent.clone(), |ia| ia.done());
 
         self.run_ice_until_connected(&answer)
             .or(async {
--- a/micro-rdk/src/common/webrtc/api.rs
+++ b/micro-rdk/src/common/webrtc/api.rs
@@ -35,6 +35,7 @@ use base64::{engine::general_purpose, Engine};
 use either::Either;
 use futures_lite::{Future, FutureExt, StreamExt};
 use prost::{DecodeError, EncodeError};
+use scopeguard::ScopeGuard;
 use sdp::{
     description::{
         common::{Address, ConnectionInformation},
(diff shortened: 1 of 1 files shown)


gvaradarajan (approved):
LGTM, did this turn out to be the sole reason for the unclosed sockets?

acmorrow:
> LGTM, did this turn out to be the sole reason for the unclosed sockets?

Hard to give a definitive "yes" to that, because I've definitely not proven that there aren't *other* potential socket leaks, but I had a consistent repro where I would fail with the `{ code: 23, kind: Uncategorized, message: "Too many open files in system" })` issue, and once this change is in place, my repro no longer results in those errors. If you would like I can show you in person tomorrow.

mattjperez (approved):
LGTM, 

did you want to add a ticket for follow-up work when we have more bandwidth?

acmorrow:
> LGTM,
> 
> did you want to add a ticket for follow-up work when we have more bandwidth?

In terms of looking for other socket leaks? I'm not sure such a ticket would be actionable without some observed misbehavior. If we keep seeing the fd exhaustion in the wild, then we can make a ticket and start a new investigation.

mattjperez:
> I've written a not great fix for this in the interest time and minimal change to the complex code in api.rs

No, not for socket leaks. What you said here made me think there was more work to be done. if not, cool.

acmorrow:
> > I've written a not great fix for this in the interest time and minimal change to the complex code in api.rs
> 
> No, not for socket leaks. What you said here made me think there was more work to be done. if not, cool.

To clarify, I think this fix is "not great" because it is more or less a band-aid. I'd prefer that the code be correct by construction, such that it was impossible for the relationship between the task and the notification controlling its lifetime to be inadvertently severed by using the type system and rusts's safety mechanisms. But that would require refactoring other parts of this code, and I don't think that's the best use of time right now.

PR #356, opened 2024-12-04
[RSDK-9200] determine max firmware size dynamically

mattjperez:
The maximum firmware size was statically set for 4MB, however partition tables can vary in size.

I've left the max size when downloading natively at 4MB (for debug/dev purposes, can be bumped easily).

On an esp32, we get the partition info for the next ota partition to be written to and use that to determine the limit.
This is great since we can use the destination's size instead of inferring. 

Won't merge until after final changes are verified on hardware.

### DIFF
--- a/micro-rdk/src/common/ota.rs
+++ b/micro-rdk/src/common/ota.rs
@@ -229,10 +229,12 @@ impl<S: OtaMetadataStorage> OtaService<S> {
                 unsafe { esp_ota_get_next_update_partition(std::ptr::null()) };
 
             if ptr.is_null() {
-                log::warn!("pointer to next update partition was null");
-                return Err(OtaError::Other(
+                log::warn!("pointer to next update partition was null, device may not be correctly partitioned");
+                let e = OtaError::Other(
                     "failed to obtain a handle to the next OTA update partition".to_string(),
-                ));
+                );
+                log::warn!(e.to_string());
+                return Err(e);
             }
             let size = unsafe { (*ptr).size } as usize;
             let address = unsafe { (*ptr).address } as usize;
@@ -268,7 +270,7 @@ impl<S: OtaMetadataStorage> OtaService<S> {
         }
 
         log::info!(
-            "proceeding with update from version `{}` to version `{}`",
+            "firmware is out of date, proceeding with update from version `{}` to version `{}`",
             stored_metadata.version,
             self.pending_version
         );
@@ -389,17 +391,17 @@ impl<S: OtaMetadataStorage> OtaService<S> {
         log::info!("writing new firmware to address `{:#x}`", self.address,);
 
         while let Some(next) = response.frame().await {
-            // hyper::Error(Body, Error { kind: Io(Custom { kind: Other, error: "ssl other error -26624" })
             let frame = next.map_err(|e| OtaError::Other(e.to_string()))?;
             if !frame.is_data() {
                 return Err(OtaError::Other(
                     "download contained non-data frame".to_string(),
                 ));
             }
+
             // Err variant returns the original frame, not an impl Error
             let data = frame
                 .into_data()
-                .map_err(|_| OtaError::Other("".to_string()))?;
+                .map_err(|_| OtaError::Other("failed to get data from frame".to_string()))?;
             total_downloaded += data.len();
 
             if !got_info {
@@ -453,7 +455,12 @@ impl<S: OtaMetadataStorage> OtaService<S> {
                 .map_err(|e| OtaError::Other(e.to_string()))?;
             // TODO change back to 'n' after impl async writer
             nwritten += data.len();
-            log::info!("updating: {}/{} bytes written", nwritten, file_len);
+            log::info!(
+                "updating next OTA partition at {}: {}/{} bytes written",
+                self.address,
+                nwritten,
+                file_len
+            );
         }
 
         drop(conn);
--- a/micro-rdk/src/common/ota.rs
+++ b/micro-rdk/src/common/ota.rs
@@ -229,9 +229,8 @@ impl<S: OtaMetadataStorage> OtaService<S> {
                 unsafe { esp_ota_get_next_update_partition(std::ptr::null()) };
 
             if ptr.is_null() {
-                log::warn!("pointer to next update partition was null, device may not be correctly partitioned");
                 let e = OtaError::Other(
-                    "failed to obtain a handle to the next OTA update partition".to_string(),
+                    "failed to obtain a handle to the next OTA update partition, device may not be partitioned properly for OTA".to_string(),
                 );
                 log::warn!(e.to_string());
                 return Err(e);
--- a/micro-rdk/src/common/ota.rs
+++ b/micro-rdk/src/common/ota.rs
@@ -48,7 +48,7 @@ use crate::{
 #[cfg(feature = "esp32")]
 use crate::esp32::esp_idf_svc::{
     ota::{EspFirmwareInfoLoader, EspOta},
-    sys::EspError,
+    sys::{esp_ota_get_next_update_partition, esp_partition_t, EspError},
 };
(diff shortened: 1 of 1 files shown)


micro-rdk/src/common/ota.rs
```diff
@@ -159,11 +158,30 @@ impl OtaService {
...
+        let max_size = {
+            log::debug!("getting size of dst partition...");
+            let ptr: *const esp_partition_t =
+                unsafe { esp_ota_get_next_update_partition(std::ptr::null()) };
+            if ptr.is_null() {
+                return Err(OtaError::Other(
+                    "failed to get dst partition info - may not exist".to_string(),
+                ));
+            }
+            let size = unsafe { (*ptr).size };
```
mattjperez, line 174: this field is a u32 according to esp-idf-svc [docs](https://docs.esp-rs.org/esp-idf-svc/esp_idf_sys/struct.esp_partition_t.html)

micro-rdk/src/common/ota.rs
```diff
@@ -159,11 +158,30 @@ impl OtaService {
...
+        #[cfg(not(feature = "esp32"))]
+        let max_size = 1024 * 1024 * 4; // 4MB
+        #[cfg(feature = "esp32")]
+        let max_size = {
+            log::debug!("getting size of dst partition...");
+            let ptr: *const esp_partition_t =
+                unsafe { esp_ota_get_next_update_partition(std::ptr::null()) };
+            if ptr.is_null() {
+                return Err(OtaError::Other(
+                    "failed to get dst partition info - may not exist".to_string(),
```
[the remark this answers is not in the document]
mattjperez: it will definitely have its own `OtaError` type. that's the final major ticket is doing a full sweep and transpose of all errors into specific `OtaErrors`.

If this fails then the `OtaService::from_config` would return an error. But I can see if there's something specific I can check at the start to confirm that `otadata` exists

micro-rdk/src/common/ota.rs
```diff
@@ -159,11 +164,41 @@ impl OtaService {
...
+            let label = {
+                // char[17] zero-terminated ascii string
+                let label_ptr = unsafe { (*ptr).address as *const std::ffi::c_char };
+                let cstr = unsafe { std::ffi::CStr::from_ptr(label_ptr) };
+                cstr.to_str()
+                    .inspect_err(|e| log::error!("failed to get label information: {}", e))
+                    .unwrap_or_default()
+                    .to_owned()
+            };
+            (size, address, label)
```
mattjperez, line 191: @acmorrow added some extra info while I figured would be good to surface about the target partition.

need to verify this works consistently on hardware before I merge it.

mattjperez: my attempts at getting the `label` led to invalid accesses. I'm sure there are ways to do it properly but I'm removing it for now as it's not essential to the PR

micro-rdk/src/common/ota.rs
```diff
@@ -364,14 +386,20 @@ impl<S: OtaMetadataStorage> OtaService<S> {
         let mut total_downloaded: usize = 0;
         let mut got_info = false;
 
+        log::info!("writing new firmware to address `{:#x}`", self.address,);
+
         while let Some(next) = response.frame().await {
-            let frame = next.unwrap();
+            // hyper::Error(Body, Error { kind: Io(Custom { kind: Other, error: "ssl other error -26624" })
```
gvaradarajan, line 392: is this an accident?

gvaradarajan (approved):
LGTM

micro-rdk/src/common/ota.rs
mattjperez: ah, yeah. was notes to myself for when I implement [RSDK-9214](https://viam.atlassian.net/browse/RSDK-9214?atlOrigin=eyJpIjoiYWVmYTU0YzI1OGRhNDNkMGEwNWU2YTMyZmIwMDc5ZDYiLCJwIjoiaiJ9), the final code-based ticket to improve the error messaging. will remove it before merging

acmorrow: Please remember to remove before commit

micro-rdk/src/common/ota.rs
```diff
@@ -364,14 +386,20 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
             if !frame.is_data() {
                 return Err(OtaError::Other(
                     "download contained non-data frame".to_string(),
                 ));
             }
-            let data = frame.into_data().unwrap();
+            // Err variant returns the original frame, not an impl Error
+            let data = frame
+                .into_data()
+                .map_err(|_| OtaError::Other("".to_string()))?;
```
acmorrow, line 402: Is there nothing more to say in the error than the emtpy string?

micro-rdk/src/common/ota.rs
```diff
@@ -220,12 +220,34 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+        #[cfg(not(feature = "esp32"))]
+        let (max_size, address) = (1024 * 1024 * 4, 0xabcd);
+        #[cfg(feature = "esp32")]
+        let (max_size, address) = {
+            log::debug!("getting handle to next OTA update partition");
+            let ptr: *const esp_partition_t =
+                unsafe { esp_ota_get_next_update_partition(std::ptr::null()) };
+
+            if ptr.is_null() {
+                log::warn!("pointer to next update partition was null");
```
acmorrow, line 232: - Maybe add "device may not be correctly partitioned for OTA" to the message.
- Similar thoughts on message in the `OtaError` below.
- Maybe use the same message for both? Do you have a way to log the message from an error?
```
let err = OtaError::Other("some good message here);
log::warn!(err.to_string())
return err
```

micro-rdk/src/common/ota.rs
```diff
@@ -241,7 +263,7 @@ impl<S: OtaMetadataStorage> OtaService<S> {
         };
 
         if self.pending_version == stored_metadata.version() {
-            log::info!("firmware is up-to-date: {}", stored_metadata.version);
+            log::info!("firmware is up-to-date: `{}`", stored_metadata.version);
```
acmorrow, line 266: ⬇️ - If you say "firmware is up-to-date" here, then in the log message on 270, say "firmware is out of date, proceeding with ..."

micro-rdk/src/common/ota.rs
```diff
@@ -403,30 +431,33 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+                .write(&data)
+                .map_err(|e| OtaError::Other(e.to_string()))?;
+            #[cfg(not(feature = "esp32"))]
+            let _n = update_handle
+                .write(&data)
+                .await
+                .map_err(|e| OtaError::Other(e.to_string()))?;
+            // TODO change back to 'n' after impl async writer
+            nwritten += data.len();
+            log::info!("updating: {}/{} bytes written", nwritten, file_len);
```
acmorrow, line 456: This could use a little work. Maybe:
```
log::info("Updating next OTA partition at {}: {}/{} bytes written", address, nwritten, file_len)
```

acmorrow (approved):
LGTM. I left some small suggestions on various logs, please adjust as you see fit before merging.

micro-rdk/src/common/ota.rs
mattjperez: added it. though we already log the returned error [here](https://github.com/viamrobotics/micro-rdk/blob/a5d7a4a0a21a51c932b61e590796ab6f6193b229/micro-rdk/src/common/conn/viam.rs#L554), so you might see it twice now.

micro-rdk/src/common/ota.rs
mattjperez: updated

acmorrow (approved):

PR #357, opened 2024-12-04
[RSDK-9196] add init connection retry

mattjperez:
This adds a retry in the event a connection cannot be established (ex no network connection).
It does not implement resumable download.

It compiles, but need to verify on native and hardware. won't merge before doing so

### DIFF
--- a/micro-rdk/src/common/ota.rs
+++ b/micro-rdk/src/common/ota.rs
@@ -283,9 +283,8 @@ impl<S: OtaMetadataStorage> OtaService<S> {
 
         let (mut sender, conn) = loop {
             match self.connector.connect_to(&uri) {
-                Ok(acceptor) => {
-                    // failed to initialize the acceptor for some reason
-                    match acceptor.await {
+                Ok(connection) => {
+                    match connection.await {
                         Ok(io) => {
                             match http2::Builder::new(self.exec.clone())
                                 .max_frame_size(16_384) // lowest configurable
@@ -295,7 +294,7 @@ impl<S: OtaMetadataStorage> OtaService<S> {
                             {
                                 Ok(pair) => break pair,
                                 Err(e) => {
-                                    log::error!("failed to build request: {}", e);
+                                    log::error!("failed to build http request: {}", e);
                                 }
                             }
                         }
@@ -305,12 +304,15 @@ impl<S: OtaMetadataStorage> OtaService<S> {
                     }
                 }
                 Err(e) => {
-                    // log and continue
-                    log::error!("failed to initiate connection: {}", e);
+                    log::error!("failed to create http connection: {}", e);
                 }
             };
 
-            log::debug!("retrying connection in {} seconds", CONN_RETRY_SECS);
+            log::debug!(
+                "retry establishing http connection to {} in {} seconds",
+                &uri,
+                CONN_RETRY_SECS
+            );
             Timer::after(Duration::from_secs(CONN_RETRY_SECS)).await;
         };
 
--- a/micro-rdk/src/common/ota.rs
+++ b/micro-rdk/src/common/ota.rs
@@ -274,13 +274,6 @@ impl<S: OtaMetadataStorage> OtaService<S> {
             uri = hyper::Uri::from_parts(parts).map_err(|e| OtaError::Other(e.to_string()))?;
         };
 
-        let io = self
-            .connector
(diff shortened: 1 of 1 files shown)


micro-rdk/src/common/ota.rs
```diff
@@ -189,20 +192,38 @@ impl OtaService {
...
+                        }
+                        Err(e) => {
+                            log::error!("failed to create tcp stream: {}", e);
+                        }
+                    }
+                }
+                Err(e) => {
+                    // log and continue
+                    log::error!("failed to initiate connection: {}", e);
+                }
```
acmorrow, line 222: - Is there some way to structure this that isn't so branchy / procedural? Could the logs be generated via `inspect_err`?
- Like in the other review, let's take the time to make these log messages better.

micro-rdk/src/common/ota.rs
```diff
@@ -189,20 +192,38 @@ impl OtaService {
...
+            };
+
+            log::debug!("retrying connection in {} seconds", CONN_RETRY_SECS);
```
acmorrow, line 225: Helpful to say what is being retried here.

micro-rdk/src/common/ota.rs
```diff
@@ -49,15 +49,18 @@ use crate::esp32::esp_idf_svc::{
...
 use hyper::{body::Bytes, client::conn::http2, Request};
 use once_cell::sync::Lazy;
+use std::time::Duration;
 use thiserror::Error;
 #[cfg(not(feature = "esp32"))]
 use {bincode::Decode, futures_lite::AsyncWriteExt};
 
 // TODO(RSDK-9200): set according to active partition scheme
 const OTA_MAX_IMAGE_SIZE: usize = 1024 * 1024 * 4; // 4MB
+const CONN_RETRY_SECS: u64 = 60;
```
acmorrow, line 63: Rather than the constant being a u64 with `SECS` in the name, and then constructing a Duration from it each time, could this constant just **be** the Duration?

micro-rdk/src/common/ota.rs
```diff
@@ -189,20 +192,38 @@ impl OtaService {
...
-        let (mut sender, conn) = {
-            http2::Builder::new(self.exec.clone())
-                .max_frame_size(16_384) // lowest configurable value
-                .timer(H2Timer)
-                .handshake(io)
-                .await
-                .map_err(|e| OtaError::Other(e.to_string()))?
+        let (mut sender, conn) = loop {
+            match self.connector.connect_to(&uri) {
+                Ok(acceptor) => {
```
acmorrow, line 198: Why `acceptor` here, when you are doing a `connect`?

acmorrow (changes requested):
The logic looks pretty sound but I think there are stylistic and diagnostic improvements to be made.

micro-rdk/src/common/ota.rs
mattjperez: yeah, i did that initially as a local variable to reuse in logs, but `Duration` doesn't impl display. though now that I think about it, Duration does have an `as_secs` method that returns a `u64`, so I'll use that.

micro-rdk/src/common/ota.rs
mattjperez: you're right. it should be `connector` according to the [return value](https://github.com/viamrobotics/micro-rdk/blob/0fedde4460f0f88510ad4fe14877521ffae30e20/micro-rdk/src/esp32/tcp.rs#L480). I got it mixed up with the method above it, `accept_connection`.

micro-rdk/src/common/ota.rs
```diff
@@ -189,20 +192,40 @@ impl OtaService {
...
+                    }
+                }
+                Err(e) => {
+                    log::error!("failed to create http connection: {}", e);
+                }
+            };
+
+            log::debug!(
+                "retry establishing http connection to {} in {} seconds",
+                &uri,
```
mattjperez, line 225: @acmorrow added the uri that was tried here and modified the other logs slightly. As the other errors will also hit this log I don't think it's necessary to add the uri to them as well, especially since the uri's with embedded permissions tend to be very lengthy, though can add it if you want.

Other than that, I'm sure what more I can say above that would provide better context than surfacing the underlying error.

acmorrow (approved):
LGTM

PR #358, opened 2024-12-09
[RSDK-9296] [RSDK-9310] Improve interactions with STUN server

acmorrow:
Failing to resolve the STUN server should not crash, since this is a realistic scenario during an offline reboot.

Similarly, failing to resolve or reach the STUN server shouldn't inhibit generating a local candidate based on the local host/port, since this can be used now by local signaling.

### DIFF
--- a/micro-rdk/src/common/webrtc/ice.rs
+++ b/micro-rdk/src/common/webrtc/ice.rs
@@ -153,7 +153,12 @@ impl ICEAgent {
             return Ok(());
         }
 
-        log::debug!("looking for srv reflexive candidate");
+        log::debug!("local_candidates: registering intrinsic local candidate");
+        let our_ip = SocketAddrV4::new(self.local_ip, self.transport.local_address().map_err(|_| IceError::IceIoError)?.port());
+        let local_cand = Candidate::new_host_candidate(our_ip);
+        self.local_candidates.push(local_cand);
+
+        log::debug!("local_candidates: looking for srv reflexive candidate");
 
         let message = stun_codec::Message::<stun_codec::rfc5389::Attribute>::new(
             stun_codec::MessageClass::Request,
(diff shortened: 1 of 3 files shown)


mattjperez (approved):
LGTM

micro-rdk/src/common/webrtc/ice.rs
```diff
@@ -165,10 +176,21 @@ impl ICEAgent {
         let bytes = Bytes::from(encoder.encode_into_bytes(message).unwrap());
 
         // TODO(RSDK-3063) Twilio address is hard-coded, we should support additional server via WebRTCOptions
-        let mut stun_ip = "global.stun.twilio.com:3478".to_socket_addrs().unwrap();
+        let mut stun_ip = match "global.stun.twilio.com:3478".to_socket_addrs() {
+            Ok(stun_ip) => stun_ip,
+            Err(err) => {
+                log::warn!("Failed trying to resolve STUN server address; no reflexive candidate will be generated: {}", err);
+                return Ok(());
```
gvaradarajan, line 183: is there ever a time where we actually want to return an Error rather than Ok? Like what if we are not in an isolated network and we still can't resolve the address?

gvaradarajan (approved):
LGTM, just one question

micro-rdk/src/common/webrtc/ice.rs
acmorrow: Yes, I thought about that a bit, and also the other case, where we could add the intrinsic local candidate, and then even though we were able to resolve the STUN server, something goes wrong afterwards. In that case, we actually do return an error, but we sort of don't want to fail the method because we still did create one local candidate.

Overall, my view is that `local_candidates` probably shouldn't really be returning a `Result` at all. It is a best effort "how can I describe myself". If the answer is "I can't", it should be up to the higher level (i.e. the caller of `local_candidates` up in `api.rs`) to decide how to handle that.

PR #359, opened 2024-12-10
Pin IDF component micro-RDK version to v0.3.0

gvaradarajan:
This is due to a regression regarding a client, when v0.3.5 is released this should be changed back to the latest stable version


acmorrow (approved):
LGTM, for when we are ready to merge it (which isn't quite yet).

acmorrow:
@gvaradarajan - Also, I filed https://viam.atlassian.net/browse/RSDK-9497 so we remember to undo this before our next release, since the usual "grep for the previous version and bump it" technique would miss after the pin.

PR #361, opened 2024-12-10
[RSDK-8932] ota docs, local server

mattjperez:
This PR does three things
1. adds basic documentation about the OTA feature
2. makes the `micro-rdk-server` build use OTA by default
3. an http2 static file server for development and testing


Makefile
```diff
@@ -99,12 +99,15 @@ doc-open:
 size:
 	find . -name "esp-build.map" -exec ${IDF_PATH}/tools/idf_size.py {} \;
 
-build-esp32-bin:
-	cargo +esp espflash save-image --package micro-rdk-server --merge --chip esp32 target/xtensa-esp32-espidf/micro-rdk-server-esp32.bin -T micro-rdk-server/esp32/partitions.csv -s 4mb  --bin micro-rdk-server-esp32 --target=xtensa-esp32-espidf  -Zbuild-std=std,panic_abort --release
+build-esp32-bin: build-esp32-ota
```
mattjperez, line 102: only annoying thing is you'll see the same terminal output twice, since the output doesn't differentiate between a full build or app image build.

acmorrow (changes requested):

etc/ota-server/src/main.rs
```diff
@@ -0,0 +1,37 @@
...
+#[tokio::main]
+async fn main() {
+    tracing_subscriber::registry()
+        .with(
+            tracing_subscriber::EnvFilter::try_from_default_env().unwrap_or_else(|_| {
+                format!("{}=debug,tower_http=debug", env!("CARGO_CRATE_NAME")).into()
+            }),
+        )
+        .with(tracing_subscriber::fmt::layer())
+        .init();
```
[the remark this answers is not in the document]
mattjperez: it goes to stdout and is configured the same as normal logging (ex `RUST_LOG=debug cargo r`, `RUST_LOG make serve-dev-ota`). in it's current use, it's the same as the log crate.

This program is from axum's examples, and the `tracing` ecosystem is used as much as `tokio` is in rust web development. It can easily tie into language-agnostic log collectors, syslogs, etc. 

can port it to use just the log crate or something if you want

Makefile
```diff
@@ -99,11 +99,34 @@ doc-open:
...
+	cargo +esp espflash save-image \
```
acmorrow, line 103: These are looking much better. A few small things:

- Order of `--bin` and `--partition-table` args is not consistent between the two lists.
- Let's move any "only in one list" flags/args to the end of each. So for `build-esp32-bin` you already have `--merge` there, but `--flash-size` is in the middle.
- Let's make the actual target file always the last argument.

So the order of the 'sections' is:
- identical
- common flag but different args (e.g. partition table)
- unique (e.g. `--merge` and `--flash-size`)
- file target

OTA.md
```diff
@@ -0,0 +1,112 @@
...
+	  "name": "OTA",
+      "namespace": "rdk",
+      "type": "generic",
+      "model": "ota_service",
+      "attributes": {
+        "url": <firmware-download-url>,
+        "version": <some-tag>
+	  }
+	}
+```
```
acmorrow, line 24: Indentation here is still not right.

OTA.md
```diff
@@ -0,0 +1,112 @@
...
+The terms 'firmware' or 'binary' can be a bit generic.
+In this section, we will refer to two types of binaries that can be built.
+1. a Merged Image
+2. an [Application (App) Image](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/system/app_image_format.html)
+
+> Note: in the absence of a `factory` partition `ota_0` fills the same initial role.
+
+## Full Build
+
+If a device is built with the above partition table, the `make build-esp32-bin` command creates a Merged Binary that includes
```
acmorrow, line 70: Above you say "Merged Image" but here "Merged Binary".

OTA.md
```diff
@@ -0,0 +1,112 @@
...
+- the bootloader
+- the partition table mapping (`partitions.csv`)
+- populated partitions (with partition headers) according to the mapping
+/
```
acmorrow, line 74: Stray `/`?

OTA.md
```diff
@@ -0,0 +1,112 @@
...
+This is the build workflow which must be used to:
+- flash a new device for the first time
+- update a device's partition table
+  - for example, make a device capable of OTA.
+
+**This is not the build that should be hosted at the `url` in the service config.**
+You can confirm this by using `ls -l** in your build directory to compare the size of the binary to your partition table.
+
+## OTA Build
+The `ota` build produces an app image (described above), which internally consists of:
```
acmorrow, line 86: In the `## Full Build` section above, you describe `make build-esp32bin` early, as the task to be performed, but this section doesn't have that. I thihnk it should

OTA.md
```diff
@@ -0,0 +1,112 @@
...
+- the type-specific partition header, `esp_app_desc_t**
+- the application image that contains the program instructions
+
+**This app image is what you must host, see [Firmware Hosting Options](#firmware-hosting-options).**
+
+This build must be within the size limits of the smallest `ota_*` partition in a device's *current* partition table. 
+
+In this document's example, both the `ota_0` and `ota_1` partitions are ~3.4 MB. 
+The command underlying `make build-esp32-ota` takes the partition table as input so it **should** fail, additionally, target devices must be programmed to verify at runtime that the assertion holds.
```
acmorrow, line 95: Have you tried this? Does it actually fail? Does `--target-app-partition` help?

etc/ota-server/src/main.rs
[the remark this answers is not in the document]
acmorrow: No, if this is canonical for axum then that's fine.

acmorrow (changes requested):
Some new comments and notes on updates. Also looks like a few original comments not addressed quite yet.

OTA.md
```diff
@@ -0,0 +1,90 @@
...
+
+```
+# ESP-IDF Partition Table
+# Name,   Type, SubType, Offset,  Size, Flags
+nvs,      data, nvs,     0x9000,  0x4000,
+otadata,  data, ota,     0xd000,  0x2000,
+phy_init, data, phy,     0xf000,  0x1000,
+factory,  app,  factory, 0x10000,  1M,
+ota_0,    app,  ota_0,   0x110000, 1M,
+ota_1,    app,  ota_1,   0x210000, 1M,
```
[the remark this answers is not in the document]
mattjperez: updated to use partitions in `micro-rdk-server/ota8mb_partitions.csv`

OTA.md
```diff
@@ -0,0 +1,90 @@
...
+
+```
+
+The terms 'firmware' or 'binary' can be a bit generic.
+In this section, we will refer to two types of binaries that can be built.
+1. a Merged Binary
+2. an App Image
```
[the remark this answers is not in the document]
mattjperez: I added links to the latest esp docs regarding it for clarification. App Image Format is what the format it called. I can see how it might be a bit confusing but wanted to keep things as close to official docs as possible so they can find it if they seach `esp app image` or something

OTA.md
```diff
@@ -0,0 +1,90 @@
...
+The command `make flash-esp32-bin` writes this entire merged binary to the device's flash memory.
+
+This is the build workflow which must be used if you want to update a device's partition table; for example, to make a device capable of OTA.
+
+**This is not the build that should be hosted at the `url` in the service config.**
+You can confirm this by using `ls -l` in your build directory to compare the size of the binary to your partition table.
+
+### OTA Build
+The `ota` build consists of *only*:
+- the type-specific partition header, `esp_app_desc_t`
```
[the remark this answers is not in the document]
mattjperez: added note that the output of this is what should be hosted, the download target of an ota update

etc/ota-server/src/main.rs
```diff
@@ -0,0 +1,37 @@
+use axum::Router;
+use local_ip_address::local_ip;
+use std::net::SocketAddr;
+use tower_http::{services::ServeDir, trace::TraceLayer};
+use tracing_subscriber::{layer::SubscriberExt, util::SubscriberInitExt};
+
+const MICRO_RDK_OTA_BIN: &str = "micro-rdk-server-esp32-ota.bin";
+const TARGET_DIR: &str = "target/xtensa-esp32-espidf";
```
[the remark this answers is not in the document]
mattjperez: yes, I didn't put more time into trying to make this better. I thought about using `CARGO_TARGET_DIR` but didn't seem like much of an improvement. happy to try ideas if anyone has

OTA.md
mattjperez: discussed offline. it does fail, but there is ambiguity (at least when not using this flag) as to which (possible all)  partitions of subtype `app` are being checked. 

ultimately, users of micro-rdk are most likely using the partition tables we provide and advanced users can find advanced uses/info in espressif official docs

Makefile
acmorrow: I spent some time seeing if there was a way to do the merge as its own step, or to request production of both merged and unmerged simultaneously, but it yielded nothing.

etc/ota-server/src/main.rs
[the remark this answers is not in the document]
acmorrow: I don't think it is worth further time right now. This is a developer facing tool so we can iterate on it as we see fit or feel the need.

Makefile
mattjperez: removed `--flash-size` as it doesn't appear to be necessary, both `make build-*` still work.

etc/ota-server/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+authors.workspace = true
+description.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+version.workspace = true
+rust-version.workspace = true
+
+[dependencies]
+axum = {version = "0.7.9", features = ["http2"]}
```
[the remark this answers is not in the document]
acmorrow: Ping on this? I think this and the json formatting are the only things left before merge.

OTA.md
```diff
@@ -0,0 +1,90 @@
...
+    {
+      "name": "OTA",
+      "namespace": "rdk",
+      "type": "generic",
+      "model": "ota_service",
+      "attributes": {
+        "url": "",
+        "version": ""
+		}
+	}
```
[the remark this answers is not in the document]
mattjperez: that was more annoying than expected. different editors accounted for whitespace and had different space/tab behavior. ended up just using github editor for it

Cargo.toml
```diff
@@ -8,6 +8,7 @@ members = [
   "micro-rdk-server",
   "micro-rdk-ffi",
   "examples/modular-drivers",
+  "etc/ota-server",
```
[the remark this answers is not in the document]
acmorrow: Oh, forgot about this one. Please remember to rename this before merge.

acmorrow (approved):
LGTM mod renaming `ota-server` and hoisting the new crate dependencies up into the root `Cargo.toml`.

Cargo.toml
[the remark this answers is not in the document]
mattjperez: yup, making those code changes now

mattjperez:
rendered here https://github.com/mattjperez/micro-rdk/blob/rsdk-8932-ota-docs/OTA.md

OTA.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+In the `url` field, enter the url where a firmware will be downloaded from
+   - if using a local endpoint on the same network, remember to use the proper private ip address
+   - for cloud-hosted resources, embedded auth in the url is easiest, we don't support tokens yet
+
+
+The `version` field is equivalent to a `tag` and can be any arbitrary string of up to 128 characters. 
+After successfully applying the new firmware, this `version` will be stored in NVS. 
+This values is compared to that of the latest machine config from app.viam.com and will trigger the update process.
```
gvaradarajan, line 34: `values` -> `value`? Also maybe clarify that a detected change in the version is what prompts the update process

OTA.md
```diff
@@ -0,0 +1,118 @@
...
+
+The command `make flash-esp32-bin` writes this entire Merged Image to the device's flash memory.
+
+This is the build workflow which must be used to:
+- flash a new device for the first time
+- update a device's partition table
+  - for example, make a device capable of OTA.
+
+**This is not the build that should be hosted at the `url` in the service config.**
+You can confirm this by using `ls -l** in your build directory to compare the size of the binary to your partition table.
```
gvaradarajan, line 83: should we clarify this with the expected size difference?

OTA.md
```diff
@@ -0,0 +1,118 @@
...
+  - use `make serve-dev-ota` to create a local endpoint for serving the ota app image
+  - the command will build the ota firmware first before serving the url
+
+### Cloud
+
+The OTA Service in the micro-rdk currently supports **only HTTP/2**, this means that the hosting platform must support HTTP/2 connections. 
+
+While not all blob storage platform support HTTP/2, many offer Content Delivery Network (CDN) solutions that do.
+
+We don't currently support tokens in the [OTA Service Config](#ota-service-config), so if permissions are required to access the endpoint they must be embedded in the URL itself.
```
gvaradarajan, line 110: `tokens` -> `authentication tokens`. `embedded in the URL itself` -> `embedded in the URL as query params`?

gvaradarajan (approved):
LGTM, optional suggestions

Makefile
mattjperez: put it back, turns out it did influence the final build size for some reason. It appears like it may be adding padding with `--flash-size` for the merged binary to be the same size as the flash target. building without is only slightly larger than the ota app image by itself.

PR #362, opened 2024-12-10
[RSDK-9244] [RSDK-9245] Feature gate local signaling and adjust max connections

acmorrow:
See notes inline for more details on each change.


examples/esp-idf-component/sdkconfig.defaults
```diff
@@ -4,6 +4,7 @@ CONFIG_ESP32_DEFAULT_CPU_FREQ_240=y
 CONFIG_ESP_SYSTEM_EVENT_TASK_STACK_SIZE=4096
 CONFIG_PTHREAD_TASK_STACK_SIZE_DEFAULT=8192
 CONFIG_ESP_TLS_SERVER=y
+CONFIG_LWIP_MAX_SOCKETS=13
```
acmorrow, line 7: Now that we have a fix for https://viam.atlassian.net/browse/RSDK-9213 in via #355, I'm more comfortable bumping this up a little bit. Arguably, we can have more sockets now since you may have both local WebRTC and app brokered running simultaneously.

micro-rdk-ffi/src/ffi/runtime.rs
```diff
@@ -226,24 +226,15 @@ pub unsafe extern "C" fn viam_server_start(ctx: *mut viam_server_context) -> via
...
-            use micro_rdk::esp32::esp_idf_svc::hal::sys::g_wifi_feature_caps;
-            use micro_rdk::esp32::esp_idf_svc::hal::sys::CONFIG_FEATURE_CACHE_TX_BUF_BIT;
-            if !g_spiram_ok {
-                log::info!("spiram not initialized disabling cache feature of the wifi driver");
-                g_wifi_feature_caps &= !(CONFIG_FEATURE_CACHE_TX_BUF_BIT as u64);
-                1
-            } else {
-                3
-            }
+    #[cfg(target_os = "espidf")]
```
acmorrow, line 229: Logic to determine the number of max conns was present everywhere that led to instantiating a `ViamServerBuilder`, and was slightly different everywhere. I've opted to consolidate the logic inside `ViamServerBuilder` and let it decide the default. Explicit calls to `with_max_concurrent_connection` are no longer required to raise the default above one.

micro-rdk/Cargo.toml
```diff
@@ -15,7 +15,7 @@ links = "micro_rdk"
 crate-type = ["lib"]
 
 [features]
-default = ["builtin-components", "data"]
+default = ["builtin-components", "data", "local-signaling"]
```
acmorrow, line 18: Added a default-enabled feature gate for `local-signaling` so it is easier to disable if we need.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -194,6 +194,33 @@ where
     <Storage as RobotConfigurationStorage>::Error: Debug,
     ServerError: From<<Storage as RobotConfigurationStorage>::Error>,
 {
+
+    pub fn get_default_max_concurrent_connections() -> usize {
```
acmorrow, line 198: This is now where we compute the default max number of concurrent connections.

micro-rdk/src/common/grpc.rs
```diff
@@ -145,6 +145,7 @@ where
     }
 
     // TODO(RSDK-9242): Use the builder pattern instead.
+    #[allow(dead_code)]
```
acmorrow, line 148: I opted not to attempt to actually avoid compiling the local signaling support when the `local-signaling` feature was off, because it got really messy with the `Either` type over the app signaling vs local signaling. I doubt the code savings are significant. If anyone thinks it would be worth it, I'm happy to file a ticket for sometime down the road to attempt it.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -194,6 +194,33 @@ where
...
+        #[allow(unused_mut)]
+        let mut max = 3;
+
+        // If local signaling is enabled, grant two more
+        #[cfg(feature = "local-signaling")]
+        {
+            max += 2;
+        }
+
+        // But on an esp32 lacking SPIRAM, assume only one connection can be realized
```
mattjperez, line 208: What's the behavior of an SDK trying to connect if local signaling is enabled but lacking spiram (when max connections is 1)? 

might be good to have a log message or something to give the user a heads-up about the limitation.
`log::warn!("this device can only support one connection at a time blah blah blah");`

micro-rdk/src/common/conn/viam.rs
```diff
@@ -205,7 +232,7 @@ where
             http2_server_port: 12346,
             http2_server_insecure: false,
             app_client_tasks: Default::default(),
-            max_concurrent_connections: 1,
+            max_concurrent_connections: Self::get_default_max_concurrent_connections(),
```
mattjperez, line 235: Nice

micro-rdk/src/common/grpc.rs
mattjperez: does this mean we can't complie if `local-signaling` flag is disabled?

This feels like a weird state to leave things in, and I personally think we should not have a flag for local signaling anyway. If anything, we can have an env var to check if local signaling is disabled and adds/removes it in the builder step.

mattjperez (approved):
LGTM, just some questions.

micro-rdk/src/common/grpc.rs
acmorrow: No, this is exactly so we *can* compile / clippy when the `local-signaling` flag is disabled. Otherwise we get warnings here that `register_signaling_server` is unused.

micro-rdk/src/common/conn/viam.rs
acmorrow: Yes, actually, you've caught something that I intended to do, but then forgot to execute on. My intention was that if max conns is too small, to disable local signaling, with a warning.

I will add logic for that and post an update.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -640,13 +667,16 @@ where
             webrtc_config: &self.webrtc_configuration,
             network,
             incomming_connection_manager: IncomingConnectionManager::new(
-                self.max_concurrent_connections as usize,
+                self.max_concurrent_connections,
             ),
             robot_config: &config,
-            local_signaling_server: Arc::new(SignalingServer::new(
+            #[cfg(feature = "local-signaling")]
+            local_signaling_server: Some(Arc::new(SignalingServer::new(
```
acmorrow, line 677: @mattjperez - Here is where we should decide to not instantiate the `SignalingServer` if the max conns is too low.

micro-rdk/src/common/grpc.rs
mattjperez: gotcha, thanks for clarifying.

micro-rdk/src/common/conn/viam.rs
acmorrow: @mattjperez - Fixed this in latest.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -837,8 +843,18 @@ where
...
+        // concurrent connections, don't enable local signaling. This
+        // value may need to be adjusted if the logic in
+        // `ViamServerBuilder::get_default_max_concurrent_connections`
+        // changes.
         #[cfg(feature = "local-signaling")]
-        let ss = self.local_signaling_server.clone();
+        let ss = self
+            .local_signaling_server
+            .clone()
+            .filter(|_| self.incomming_connection_manager.max_connections() >= 5);
```
mattjperez, line 856: that's pretty cool, forgot options can use `filter`. this way the `max_connections()` check only happens if `self.local_signaling_server.is_some()`.

My instinct would have been to go with this
```rs
let ss = if self.incomming_connection_manager.max_connections() >= 5 { 
    self.local_signaling_server.clone() 
} else {
    None
}
``` 

Did you still want to add the warning though?

mattjperez (approved):

PR #363, opened 2024-12-13
[RSDK-9214] add more expressive error types and messages

mattjperez:
I moved away from `EspError` and `#[error(transparent)]` and have it return a string since there isn't a single place EspError happens (i.e. getting the app image header, getting the write handle, writing, aborting, etc). The essence of the message should still be captured in the string conversion

### DIFF
--- a/micro-rdk/src/common/ota.rs
+++ b/micro-rdk/src/common/ota.rs
@@ -107,33 +107,36 @@ struct EspAppDesc {
 pub(crate) enum ConfigError {
     #[error("version {0} has length {1}, maximum allowed characters is {2}")]
     InvalidVersionLen(String, usize, usize),
-    #[error("the given url {0}, is not a valid URI: {1}")]
+    #[error("the url `{0}`, is not valid: {1}")]
     InvalidUrl(String, String),
     #[error(transparent)]
     AttributeError(#[from] AttributeError),
-    #[error("config missing attribute `{0}`")]
+    #[error("required config attribute `{0}` not found")]
     MissingAttribute(String),
-    #[error("config missing value for `{0}`")]
+    #[error("value missing for field `{0}`")]
     MissingValue(String),
     #[error("{0}")]
     Other(String),
 }
 
-// TODO(RSDK-9214): surface (and use) underlying errors properly, use transparent where possible
 #[derive(Error, Debug)]
 pub(crate) enum OtaError {
+    #[error("error occured during abort process: {0}")]
+    AbortError(String),
     #[error("{0}")]
     ConfigError(#[from] ConfigError),
     #[error(transparent)]
     NetworkError(#[from] hyper::Error),
-    #[error("error getting the writer handle for OTA: {0}")]
-    HandleError(String),
+    #[error("{0}")]
+    UpdateError(String),
     #[error("failed to initialize ota process")]
     InitError,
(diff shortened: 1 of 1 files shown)


micro-rdk/src/common/ota.rs
```diff
@@ -103,20 +103,37 @@ struct EspAppDesc {
     reserv2: [u32; 20],
 }
 
+#[derive(Error, Debug)]
+pub(crate) enum ConfigError {
+    #[error("version {0} has length {1}, maximum allowed characters is {2}")]
+    InvalidVersionLen(String, usize, usize),
+    #[error("the given url {0}, is not a valid URI: {1}")]
```
acmorrow, line 110: - Is the url vs uri distinction in this message intentional and making a subtle point, or should it be one or the other?
- If one is capitalized, I'd expect the other to be, or vice versa.

micro-rdk/src/common/ota.rs
```diff
@@ -103,20 +103,37 @@ struct EspAppDesc {
...
+    InvalidUrl(String, String),
+    #[error(transparent)]
+    AttributeError(#[from] AttributeError),
+    #[error("config missing attribute `{0}`")]
```
acmorrow, line 114: Maybe `expected` or `required` attribute if this is an error?

micro-rdk/src/common/ota.rs
```diff
@@ -103,20 +103,37 @@ struct EspAppDesc {
...
+    MissingAttribute(String),
+    #[error("config missing value for `{0}`")]
```
acmorrow, line 116: maybe `for field` if `{0}` is going to be the name of a present field (which I expect it would be).

micro-rdk/src/common/ota.rs
```diff
@@ -103,20 +103,37 @@ struct EspAppDesc {
...
+    MissingValue(String),
+    #[error("{0}")]
+    Other(String),
+}
+
 // TODO(RSDK-9214): surface (and use) underlying errors properly, use transparent where possible
 #[derive(Error, Debug)]
-pub enum OtaError {
+pub(crate) enum OtaError {
```
acmorrow, line 124: ⬆️ - Should the TODO above be removed with this commit?

micro-rdk/src/common/ota.rs
```diff
@@ -103,20 +103,37 @@ struct EspAppDesc {
...
     #[error("{0}")]
-    ConfigError(String),
-    #[cfg(feature = "esp32")]
+    ConfigError(#[from] ConfigError),
     #[error(transparent)]
-    EspError(#[from] EspError),
+    NetworkError(#[from] hyper::Error),
+    #[error("error getting the writer handle for OTA: {0}")]
+    HandleError(String),
```
acmorrow, line 130: I don't love `HandleError` - the error isn't really so much about handles, as it is the ability to use the handle to update. So, maybe `WriteError` or `UpdateError`?

acmorrow (changes requested):
Looks good, some small comments and nits.

micro-rdk/src/common/ota.rs
mattjperez: it was making a subtle point, but I've decided to go with URL (lowercase) to be consistent. The underlying error may mention it regardless

acmorrow (approved):

PR #364, opened 2024-12-16
[RSDK-9134] Document macOS challenges with Homebrew python and ESP-IDF

acmorrow:
Viewable here: https://github.com/acmorrow/micro-rdk/blob/RSDK-9134/DEVELOPMENT.md#use-crate_cc_no_defaults


mattjperez (approved):
LGTM

PR #365, opened 2024-12-16
[RSDK-9548] log storage usage on startup

gvaradarajan:


micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -126,6 +126,40 @@ impl NVSStorage {
     }
 }
 
+impl StorageDiagnostic for NVSStorage {
+    type Error = NVSStorageError;
+    fn log_space_diagnostic(&self) -> Result<(), Self::Error> {
+        let part_name = CString::new("nvs")
```
acmorrow, line 132: Does this only work if the name of the partition is `nvs`? I don't think esp-idf requires that it be called that, and I think the NVS library will use all partitions of type `data` and subtype `nvs`.

Short of dealing with that complexity, could you dig out the name from `self.nvs` instead? I think that way it would honor the `partition_name` `str` passed to `NVSStorage::new`.

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -126,6 +126,40 @@ impl NVSStorage {
...
+            .map_err(|_| EspError::from_non_zero(NonZeroI32::new(ESP_ERR_INVALID_ARG).unwrap()))?;
+        let mut stats: nvs_stats_t = Default::default();
+        esp!(unsafe { nvs_get_stats(part_name.as_ptr(), &mut stats as *mut _) })?;
+
+        // 32 bytes per entry
```
acmorrow, line 137: Can we link to docs that support this?

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -126,6 +126,40 @@ impl NVSStorage {
...
+        let used_entries = stats.used_entries;
+        let used_space = used_entries * 32;
+        let used_page_header_space = (used_entries % 125) * 32;
+
+        let used_space = used_space + used_page_header_space;
+        let total_space = stats.total_entries * 32;
+
+        log::info!("CREDENTIAL STORAGE STATS: {:?} bytes used", used_space);
+        // From experimentation we have found that NVS requires 4000 bytes of
```
acmorrow, line 146: Exactly 4000, or 4096?

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -126,6 +126,40 @@ impl NVSStorage {
...
+        // unused space for reasons unknown. The percentage portion of the calculation
+        // comes from the maximum allowed blob size stated in the ESP32 documentation
+        // on NVS
+        let total_usable_space = (0.976 * (total_space as f64)) - 4000.0;
+        let percentage_used = (used_space as f64) / total_usable_space;
+        if percentage_used > 0.9 {
+            log::warn!(
+                "{:?} of {:?} bytes used, consider examination of storage usage",
```
acmorrow, line 154: "bytes of NVS storage in partition {}", maybe.

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -300,6 +305,15 @@ impl WifiCredentialStorage for RAMStorage {
     }
 }
 
+impl StorageDiagnostic for RAMStorage {
+    type Error = Infallible;
+    fn log_space_diagnostic(&self) -> Result<(), Self::Error> {
+        let used_space = size_of_val(self);
+        log::info!("CREDENTIAL STORAGE STATS: {:?} bytes used", used_space);
```
mattjperez, line 312: does this just return the size of the `Rc<Mutex<>>`? or does it actually change depending on our usage in RAMStorage?

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -126,6 +126,40 @@ impl NVSStorage {
     }
 }
 
+impl StorageDiagnostic for NVSStorage {
+    type Error = NVSStorageError;
+    fn log_space_diagnostic(&self) -> Result<(), Self::Error> {
+        let part_name = CString::new("nvs")
+            .map_err(|_| EspError::from_non_zero(NonZeroI32::new(ESP_ERR_INVALID_ARG).unwrap()))?;
```
mattjperez, line 133: should we be unwrapping here?

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -126,6 +126,40 @@ impl NVSStorage {
...
+        let mut stats: nvs_stats_t = Default::default();
+        esp!(unsafe { nvs_get_stats(part_name.as_ptr(), &mut stats as *mut _) })?;
+
+        // 32 bytes per entry
+        let used_entries = stats.used_entries;
+        let used_space = used_entries * 32;
+        let used_page_header_space = (used_entries % 125) * 32;
```
mattjperez, line 140: could we make these `const` with descriptive names? can't tell what 125 is for

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -126,6 +126,40 @@ impl NVSStorage {
...
+
+        let used_space = used_space + used_page_header_space;
+        let total_space = stats.total_entries * 32;
+
+        log::info!("CREDENTIAL STORAGE STATS: {:?} bytes used", used_space);
+        // From experimentation we have found that NVS requires 4000 bytes of
+        // unused space for reasons unknown. The percentage portion of the calculation
+        // comes from the maximum allowed blob size stated in the ESP32 documentation
+        // on NVS
+        let total_usable_space = (0.976 * (total_space as f64)) - 4000.0;
```
mattjperez, line 150: why `0.976`? just to have room before panicking or something?
also, is this correct that we want to subtract 4000 from this percentage of total space? 
it's not `total_usable_space = ((total_space as f64) - 4000.0)`

mattjperez (changes requested):
looking good, have some clarifying questions though

acmorrow (changes requested):

micro-rdk/src/esp32/nvs_storage.rs
gvaradarajan: It was taken from the length restrictions here: https://docs.espressif.com/projects/esp-idf/en/v4.4.8/esp32/api-reference/storage/nvs_flash.html#keys-and-values

micro-rdk/src/esp32/nvs_storage.rs
gvaradarajan: I'm worried about posting what might be a dead link in the future

micro-rdk/src/common/credentials_storage.rs
gvaradarajan: I checked and you're right, this doesn't give the right size

micro-rdk/src/esp32/nvs_storage.rs
gvaradarajan: Yes, I think this is an appropriate use of unwrap, if this panics it means there is something serious wrong with ESP-IDF

micro-rdk/src/esp32/nvs_storage.rs
gvaradarajan: I was trying to account for entries dedicated to NVS page headers but it seems that the total entries allowed does include those in the count, so I'm removing this anyway

micro-rdk/src/common/credentials_storage.rs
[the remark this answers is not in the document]
acmorrow: Any thoughts on this?

micro-rdk/src/common/conn/viam.rs
```diff
@@ -454,6 +464,9 @@ where
     }
 
     pub(crate) async fn run(&mut self) -> ! {
+        if let Err(err) = self.storage.log_space_diagnostic() {
```
acmorrow, line 467: I'm not sure it makes sense for `log_space_diagnostic` to return an error at all. If it can't log, it is unlikely that this can either, and if it can log, it can report its own failures.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -643,6 +656,7 @@ where
                     })
                     .ok()
             } else {
+                log::info!("falling back to stored TLS certificate");
```
acmorrow, line 659: It says what but not why. Maybe `log::info!("Failed to obtain certificates from {}, will attempt to load any stored certificates", self.app_client.uri()")` or similar. You could decide against logging the app address if you want, since otherwise you'd need to do the work to expose the URI held in the inner GrpcClient.

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -126,6 +130,40 @@ impl NVSStorage {
...
+        let total_space = stats.total_entries * BYTES_PER_ENTRY;
+
+        // From experimentation we have found that NVS requires 4000 bytes of
+        // unused space for reasons unknown. The percentage portion of the calculation (0.976)
+        // comes from the blob size restriction as stated in the ESP32 documentation
+        // on NVS
+        let total_usable_space = (0.976 * (total_space as f64)) - 4000.0;
+        let percentage_used = (used_space as f64) / total_usable_space;
+        log::info!(
+            "Credential storage stats: {:?} of {:?} bytes of NVS storage used",
```
acmorrow, line 152: It's not just credential storage anymore though, there's OTA and other stuff. Maybe just `NVS stats: {:?} bytes used of {:?} bytes available".

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -126,6 +130,40 @@ impl NVSStorage {
...
+            used_space,
+            total_space
+        );
+        if percentage_used > 0.9 {
```
acmorrow, line 156: If you used just `log!` could you compute a level, instead of having two logs?

Something like:

```
log!(if percentage_used > 90 { Level::Warn } else { Level::Info },
       "NVS stats {:?} bytes used of {:?} available");
```

Then you get a warning on high utilization and you don't need to say the "consider examination".

acmorrow (changes requested):

micro-rdk/src/common/credentials_storage.rs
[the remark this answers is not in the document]
gvaradarajan: yeah, I'll get rid of it

acmorrow (approved):
LGTM!

PR #366, opened 2024-12-18
[RSDK-9344] Use snappy for robot configuration compression

acmorrow:
- Before adding `snap`: `App/part. size:    3,452,128/4,128,767 bytes, 83.61%`
- After adding `snap`: `App/part. size:    3,487,808/4,128,767 bytes, 84.48%`

To test, I pasted a well-known large configuration (obtained by looking at the debug expansion of a fragment) into an attribute section of a sensor. Logs say:

```
I (8390) micro_rdk::esp32::nvs_storage: robot configuration compressed for NVS storage: 6510 bytes before, 3735 bytes after
```

So a 2,775 byte savings here, which isn't nothing given our NVS size bounds. The choice of snappy was somewhat arbitrary here, in that I tried it and it worked.

### DIFF
--- a/micro-rdk/src/esp32/nvs_storage.rs
+++ b/micro-rdk/src/esp32/nvs_storage.rs
@@ -2,7 +2,12 @@
 use bytes::Bytes;
 use hyper::{http::uri::InvalidUri, Uri};
 use prost::{DecodeError, Message};
-use std::{cell::RefCell, rc::Rc};
+use snap::{read::FrameDecoder, write::FrameEncoder};
+use std::{
+    cell::RefCell,
+    io::{Read, Write},
+    rc::Rc,
+};
 use thiserror::Error;
 
 use crate::{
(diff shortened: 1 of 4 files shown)


micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -207,13 +214,44 @@ impl RobotConfigurationStorage for NVSStorage {
...
+            compressed.len()
+        );
+        self.set_blob(NVS_ROBOT_CONFIG_KEY, compressed.into())?;
         Ok(())
     }
 
     fn get_robot_configuration(&self) -> Result<RobotConfig, Self::Error> {
-        let robot_config = self.get_blob(NVS_ROBOT_CONFIG_KEY)?;
-        RobotConfig::decode(&robot_config[..]).map_err(NVSStorageError::NVSValueDecodeError)
+        let mut decompressed_robot_config = vec![];
```
acmorrow, line 249: One thing we could do to harden this here would be that if we fail to decompress the buffer with snappy, that we then try to just feed the raw blob `RobotConfig::decode`. That'd protect against a situation where you had a non-compressed robot config in NVS, got OTA updated to a version that expected a compressed robot config, and then for some reason when you came back online were unable to reach app and the expectation would be to use the cached config.

acmorrow:
Closing this, as we will be pursuing a different approach.

PR #367, opened 2024-12-19
fix native errors for ota

mattjperez:
the dead code and wrong error variant somehow got through ci on earlier merge. this fixes it. added a couple small changes to logs too


gvaradarajan (approved):
LGTM

PR #369, opened 2024-12-20
[RSDK-9593] move ota to config monitor, config monitor restart hook 

mattjperez:
This PR:
- moves ota logic to the config monitor
- ~uses a `while let Err()` loop on `ota.update()` to ensure retries, **blocking other config changes from being applied**~.
- uses the periodic nature of the `ConfigMonitor` (every 10sec by default) to retry failed OTA attemps
- ~addresses ticket [RSDK-9594](https://viam.atlassian.net/browse/RSDK-9594) in the config monitor, haven't checked if this change needs to be made in other parts of the codebase.~

[RSDK-9594]: https://viam.atlassian.net/browse/RSDK-9594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/micro-rdk/src/common/config_monitor.rs
+++ b/micro-rdk/src/common/config_monitor.rs
@@ -8,13 +8,16 @@ use crate::{
 };
 use async_io::Timer;
 use futures_lite::{Future, FutureExt};
-use std::fmt::Debug;
-use std::pin::Pin;
-use std::time::Duration;
+use std::{fmt::Debug, pin::Pin, time::Duration};
+
+#[cfg(feature = "ota")]
+use crate::common::{exec::Executor, ota};
 
 pub struct ConfigMonitor<'a, Storage> {
     curr_config: Box<RobotConfig>, //config for robot gotten from last robot startup, aka inputted from entry
     storage: Storage,
+    #[cfg(feature = "ota")]
+    executor: Executor,
     restart_hook: Box<dyn Fn() + 'a>,
 }
 
@@ -27,11 +30,14 @@ where
     pub fn new(
         curr_config: Box<RobotConfig>,
         storage: Storage,
+        #[cfg(feature = "ota")] executor: Executor,
         restart_hook: impl Fn() + 'a,
     ) -> Self {
         Self {
             curr_config,
             storage,
+            #[cfg(feature = "ota")]
+            executor,
             restart_hook: Box::new(restart_hook),
         }
     }
@@ -72,9 +78,54 @@ where
                 .await?;
 
             if new_config
+                .as_ref()
                 .config
-                .is_some_and(|cfg| cfg != *self.curr_config)
+                .as_ref()
+                .is_some_and(|cfg| *cfg != *self.curr_config)
             {
+                log::info!("new robot config detected");
+
+                #[cfg(feature = "ota")]
+                {
+                    let config = new_config.config.clone().unwrap();
+                    if let Some(service) = config
+                        .services
+                        .iter()
+                        .find(|&service| service.model == *ota::OTA_MODEL_TRIPLET)
+                    {
+                        match ota::OtaService::from_config(
+                            service,
+                            self.storage.clone(),
+                            self.executor.clone(),
+                        ) {
+                            Ok(mut ota) => {
+                                let curr_metadata = ota.stored_metadata().await;
+                                if curr_metadata.version != ota.pending_version {
+                                    log::info!(
+                                        "firmware is out of date, updating from `{}` to `{}`",
+                                        curr_metadata.version,
+                                        ota.pending_version
+                                    );
+                                    while let Err(ota_err) = ota.update().await {
+                                        log::error!("failed to update firmware: {}", ota_err);
+                                        log::info!("retrying firmware update");
+                                    }
+                                    log::info!("firmware update successful");
+                                };
+                                let curr_metadata = ota.stored_metadata().await;
+                                log::info!(
+                                    "firmware is up to date: version `{}`",
+                                    curr_metadata.version
+                                );
+                            }
+                            Err(e) => {
+                                log::error!("failed to build ota service: {}", e.to_string());
+                                log::error!("ota service config: {:?}", service);
+                            }
+                        };
+                    }
+                }
+
                 if let Err(e) = self.storage.reset_robot_configuration() {
                     log::warn!(
                         "Failed to reset robot config after new config detected: {}",
--- a/micro-rdk/src/common/config_monitor.rs
+++ b/micro-rdk/src/common/config_monitor.rs
@@ -30,7 +30,8 @@ where
     pub fn new(
         curr_config: Box<RobotConfig>,
         storage: Storage,
-        #[cfg(feature = "ota")] executor: Executor,
+        #[cfg(feature = "ota")]
+        executor: Executor,
         restart_hook: impl Fn() + 'a,
     ) -> Self {
         Self {
--- a/micro-rdk/src/common/config_monitor.rs
+++ b/micro-rdk/src/common/config_monitor.rs
@@ -77,17 +77,11 @@ where
                 })
                 .await?;
 
-            if new_config
-                .as_ref()
-                .config
-                .as_ref()
-                .is_some_and(|cfg| *cfg != *self.curr_config)
-            {
-                log::info!("new robot config detected");
+            if let Some(config) = new_config.as_ref().config.as_ref() {
+                let mut reboot = false;
 
                 #[cfg(feature = "ota")]
                 {
-                    let config = new_config.config.clone().unwrap();
                     if let Some(service) = config
                         .services
                         .iter()
@@ -99,25 +93,20 @@ where
                             self.executor.clone(),
                         ) {
                             Ok(mut ota) => {
-                                let curr_metadata = ota.stored_metadata().await;
-                                if curr_metadata.version != ota.pending_version {
+                                if ota.needs_update().await {
                                     log::info!(
-                                        "firmware is out of date, updating from `{}` to `{}`",
-                                        curr_metadata.version,
-                                        ota.pending_version
+                                        "firmware update to `{}` detected",
+                                        ota.pending_version(),
                                     );
-                                    while let Err(ota_err) = ota.update().await {
+                                    if let Err(ota_err) = ota.update().await {
                                         log::error!("failed to update firmware: {}", ota_err);
-                                        log::info!("retrying firmware update");
+                                    };
+                                    if ota.needs_reboot() {
+                                        reboot = true;
                                     }
-                                    log::info!("firmware update successful");
-                                };
-                                let curr_metadata = ota.stored_metadata().await;
-                                log::info!(
-                                    "firmware is up to date: version `{}`",
-                                    curr_metadata.version
-                                );
+                                }
                             }
+
                             Err(e) => {
                                 log::error!("failed to build ota service: {}", e.to_string());
                                 log::error!("ota service config: {:?}", service);
@@ -126,12 +115,18 @@ where
                     }
                 }
 
-                if let Err(e) = self.storage.reset_robot_configuration() {
-                    log::warn!(
-                        "Failed to reset robot config after new config detected: {}",
-                        e
-                    );
-                } else {
+                if *config != *self.curr_config {
+                    if let Err(e) = self.storage.reset_robot_configuration() {
+                        log::warn!(
+                            "Failed to reset robot config after new config detected: {}",
+                            e
+                        );
+                    } else {
+                        reboot = true;
+                    }
+                }
+
+                if reboot {
                     // TODO(RSDK-9464): flush logs to app.viam before restarting
                     self.restart();
                 }
--- a/micro-rdk/src/common/config_monitor.rs
+++ b/micro-rdk/src/common/config_monitor.rs
@@ -30,8 +30,7 @@ where
     pub fn new(
         curr_config: Box<RobotConfig>,
         storage: Storage,
-        #[cfg(feature = "ota")]
-        executor: Executor,
+        #[cfg(feature = "ota")] executor: Executor,
         restart_hook: impl Fn() + 'a,
     ) -> Self {
         Self {
--- a/micro-rdk/src/common/config_monitor.rs
+++ b/micro-rdk/src/common/config_monitor.rs
@@ -132,6 +132,7 @@ where
                         e
                     );
                 } else {
+                    // TODO(RSDK-9464): flush logs to app.viam before restarting
                     self.restart();
                 }
             }
--- a/micro-rdk/src/common/config_monitor.rs
+++ b/micro-rdk/src/common/config_monitor.rs
@@ -92,26 +92,16 @@ where
                             self.storage.clone(),
                             self.executor.clone(),
                         ) {
-                            Ok(mut ota) => {
-                                if ota.needs_update().await {
-                                    log::info!(
-                                        "firmware update to `{}` detected",
-                                        ota.pending_version(),
-                                    );
-                                    if let Err(ota_err) = ota.update().await {
-                                        log::error!("failed to update firmware: {}", ota_err);
-                                    };
-                                    if ota.needs_reboot() {
-                                        reboot = true;
-                                    }
-                                }
-                            }
-
-                            Err(e) => {
-                                log::error!("failed to build ota service: {}", e.to_string());
-                                log::error!("ota service config: {:?}", service);
-                            }
-                        };
+                            Ok(mut ota) => match ota.update().await {
+                                Ok(needs_reboot) => reboot = needs_reboot,
+                                Err(e) => log::error!("failed to complete ota update: {}", e),
+                            },
+                            Err(e) => log::error!(
+                                "failed to create ota service from config:{} - {:?}",
+                                e,
+                                service,
+                            ),
+                        }
                     }
                 }
 
--- a/micro-rdk/src/common/config_monitor.rs
+++ b/micro-rdk/src/common/config_monitor.rs
@@ -30,7 +30,8 @@ where
     pub fn new(
         curr_config: Box<RobotConfig>,
         storage: Storage,
-        #[cfg(feature = "ota")] executor: Executor,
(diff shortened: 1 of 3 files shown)


micro-rdk/src/common/conn/viam.rs
```diff
@@ -51,7 +51,7 @@ use super::server::{IncomingConnectionManager, WebRtcConfiguration};
 use crate::common::provisioning::server::AsNetwork;
 
 #[cfg(feature = "ota")]
-use crate::common::{credentials_storage::OtaMetadataStorage, ota};
```
mattjperez, line 54: I've left this here for now instead of shifting it to `config_monitor.rs`. I like that it's part of `ViamServerStorage`, especially if we remove the flag entirely to make it a builtin

micro-rdk/src/common/ota.rs
```diff
@@ -524,9 +511,11 @@ impl<S: OtaMetadataStorage> OtaService<S> {
         // Test experimental ffi accesses here to be recoverable without flashing
         #[cfg(feature = "esp32")]
         {
-            log::info!("rebooting to load firmware from `{:#x}`", self.address);
+            log::info!(
+                "next reboot will load firmware from `{:#x}`",
+                self.address
+            );
             // TODO(RSDK-9464): flush logs to app.viam before restarting
-            esp_idf_svc::hal::reset::restart();
```
mattjperez, line 529: this shifts the responsibility of rebooting to whoever is calling `ota::update()`

micro-rdk/src/common/config_monitor.rs
```diff
@@ -72,9 +78,54 @@ where
...
+                                if curr_metadata.version != ota.pending_version {
+                                    log::info!(
+                                        "firmware is out of date, updating from `{}` to `{}`",
+                                        curr_metadata.version,
+                                        ota.pending_version
+                                    );
+                                    while let Err(ota_err) = ota.update().await {
+                                        log::error!("failed to update firmware: {}", ota_err);
+                                        log::info!("retrying firmware update");
+                                    }
```
mattjperez, line 112: wasn't sure if I should put a delay here at all since there is one internally in `update` for connection establishment already.

mattjperez: additionally, if the error is due to specific errors `ConfigError`, `InvalidImage`, etc, I'm not sure what the behavior should be.

we can 
a) continue applying the other configs (might be risky if the new configs are only compatible with new firmware)
b) continue to block on applying the current update and surface an error

micro-rdk/src/common/config_monitor.rs
```diff
@@ -72,15 +78,61 @@ where
...
+                        .iter()
+                        .find(|&service| service.model == *ota::OTA_MODEL_TRIPLET)
+                    {
+                        match ota::OtaService::from_config(
+                            service,
+                            self.storage.clone(),
+                            self.executor.clone(),
+                        ) {
+                            Ok(mut ota) => {
+                                let curr_metadata = ota.stored_metadata().await;
```
gvaradarajan, line 102: I feel like the logic here could go back to the OTA service and this could be reduced to a function call on the service? You would no longer need the `stored_metadata` function and the `pending_version` would no longer need to be `pub(crate)`

micro-rdk/src/common/config_monitor.rs
gvaradarajan: I think a) is the right move here. Consumers of the pre-packaged micro-rdk-server can see why the configs incompatible with the old firmware failed to function appropriately because the version on app will be the old version. Others can look at the logs (which is admittedly not ideal, but the solution to that is to change/revamp version reporting as we have previously discussed)

gvaradarajan (changes requested):
No logical issues, but I've suggested some shifting of code

micro-rdk/src/common/config_monitor.rs
mattjperez: you're right. because this would run every 10 seconds when the config gets checked, I wanted to control how often logging gets propogated, otherwise we's see `version up to date` all the time.

I've since moved a single log to when viam server starts to just log the current firmware once, and report only the update target when attempting ota

micro-rdk/src/common/config_monitor.rs
```diff
@@ -71,16 +77,57 @@ where
...
+                        log::warn!(
+                            "Failed to reset robot config after new config detected: {}",
+                            e
+                        );
+                    } else {
+                        reboot = true;
+                    }
+                }
+
+                if reboot {
```
mattjperez, line 129: looks cleaner to use a flag instead of fiddling with feature gates

micro-rdk/src/common/conn/viam.rs
```diff
@@ -486,6 +480,14 @@ where
             self.provision().await;
         }
 
+        #[cfg(feature = "ota")]
+        {
+            if self.storage.has_ota_metadata() {
+                let metadata = self.storage.get_ota_metadata().unwrap_or_default();
+                log::info!("firmware version: {}", metadata.version);
+            }
+        }
```
mattjperez, line 489: adding this here avoids complicating config monitor's ota stuff to get the right logging behavior.

micro-rdk/src/common/ota.rs
```diff
@@ -256,31 +269,22 @@ impl<S: OtaMetadataStorage> OtaService<S> {
             pending_version,
             max_size,
             address,
+            needs_reboot: false,
```
gvaradarajan, line 272: I think it's better to have `update` return a boolean as to whether an update was actually performed rather than store state?

gvaradarajan (approved):
LGTM with optional suggestion

micro-rdk/src/common/ota.rs
```diff
@@ -304,11 +319,19 @@ impl<S: OtaMetadataStorage> OtaService<S> {
             uri = hyper::Uri::from_parts(parts).map_err(|e| OtaError::Other(e.to_string()))?;
         };
 
+        let mut num_tries = 0;
         let (mut sender, conn) = loop {
+            num_tries += 1;
```
gvaradarajan, line 324: because of where the increment is, this ends up only trying a maximum of twice rather than thrice?

gvaradarajan (approved):
LGTM, provided the one logical issue is resolved

micro-rdk/src/common/ota.rs
mattjperez: ah you're right. should go at the bottom of the loop. fixed

micro-rdk/src/common/conn/viam.rs
```diff
@@ -486,6 +480,14 @@ where
             self.provision().await;
         }
 
+        #[cfg(feature = "ota")]
+        {
+            if self.storage.has_ota_metadata() {
+                let metadata = self.storage.get_ota_metadata().unwrap_or_default();
+                log::info!("firmware version: {}", metadata.version);
```
acmorrow, line 487: This will log the empty string if there is no OTA metadata in NVS. I'm not sure that's the best outcome. Can we be more explicit?

micro-rdk/src/common/conn/viam.rs
```diff
@@ -573,48 +575,20 @@ where
             log::error!("couldn't store the robot configuration reason {:?}", err);
         }
 
+        #[cfg(feature = "esp32")]
+        let hook = || crate::esp32::esp_idf_svc::hal::reset::restart();
+        #[cfg(not(feature = "esp32"))]
+        let hook = || std::process::exit(0);
```
acmorrow, line 581: - Before we just called `std::process::exit`, but this goes back to an older style where on esp32 we do `hal::reset::restart`. What was the motivation for that change. I think I know, but I'd like to be sure.
- If this is to effect the change I suspect, I think I'd want to see the same change in the `RestartMonitor`.
- A block nearly identical to this exists in `grpc::robot_shutdown`. Can we DRY this?

micro-rdk/src/common/ota.rs
```diff
@@ -172,6 +170,17 @@ pub(crate) struct OtaService<S: OtaMetadataStorage> {
 }
 
 impl<S: OtaMetadataStorage> OtaService<S> {
+    pub(crate) async fn stored_metadata(&self) -> OtaMetadata {
+        if !self.storage.has_ota_metadata() {
+            log::info!("no ota metadata currently stored in NVS");
```
acmorrow, line 175: nit: `OTA`

micro-rdk/src/common/ota.rs
```diff
@@ -172,6 +170,17 @@ pub(crate) struct OtaService<S: OtaMetadataStorage> {
 }
 
 impl<S: OtaMetadataStorage> OtaService<S> {
+    pub(crate) async fn stored_metadata(&self) -> OtaMetadata {
```
acmorrow, line 173: Is this function really `async`?

micro-rdk/src/common/ota.rs
```diff
@@ -259,28 +268,17 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
-        let stored_metadata = if !self.storage.has_ota_metadata() {
-            log::info!("no ota metadata currently stored in NVS");
-            OtaMetadata::default()
-        } else {
-            self.storage
-                .get_ota_metadata()
-                .inspect_err(|e| log::warn!("failed to get ota metadata from nvs: {}", e))
-                .unwrap_or_default()
-        };
+    pub(crate) async fn needs_update(&self) -> bool {
```
acmorrow, line 271: Is this function really `async`?

micro-rdk/src/common/ota.rs
```diff
@@ -520,15 +526,18 @@ impl<S: OtaMetadataStorage> OtaService<S> {
             .map_err(|e| OtaError::Other(e.to_string()))?;
 
         log::info!("firmware update complete");
-
-        // Test experimental ffi accesses here to be recoverable without flashing
+        // verifies nvs was stored correctly
+        let curr_metadata = self.stored_metadata().await;
+        log::info!(
+            "firmware update successful: version `{}`",
```
acmorrow, line 532: If you have this, do you still need the "firmware update complete" line above?

micro-rdk/src/common/ota.rs
```diff
@@ -259,28 +268,17 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+    pub(crate) async fn needs_update(&self) -> bool {
+        self.stored_metadata().await.version != self.pending_version
+    }
 
-        if self.pending_version == stored_metadata.version() {
-            log::info!("firmware is up-to-date: `{}`", stored_metadata.version);
-            return Ok(());
+    /// Attempts to perform an OTA update.
+    /// On success, returns an `Ok(true)` or `Ok(false)` indicating if a reboot is necessary.
+    pub(crate) async fn update(&mut self) -> Result<bool, OtaError> {
```
acmorrow, line 277: I don't particularly love the `bool` result. Maybe my unease is more to do with the name of the method?

micro-rdk/src/common/ota.rs
```diff
@@ -337,6 +342,7 @@ impl<S: OtaMetadataStorage> OtaService<S> {
                 CONN_RETRY_SECS
             );
             Timer::after(Duration::from_secs(CONN_RETRY_SECS)).await;
+            num_tries += 1;
```
acmorrow, line 345: - Is there a way to frame this loop without an explicitly managed counter?
- If you really need the counter, I'd recommend incrementing at or near the top so it can't somehow become skipped later.

micro-rdk/src/common/ota.rs
```diff
@@ -304,11 +302,18 @@ impl<S: OtaMetadataStorage> OtaService<S> {
             uri = hyper::Uri::from_parts(parts).map_err(|e| OtaError::Other(e.to_string()))?;
         };
 
+        let mut num_tries = 0;
         let (mut sender, conn) = loop {
+            if num_tries == NUM_RETRY_CONN {
```
acmorrow, line 307: Is this a necessary part of this change? It seems independent of moving the driver of OTA checks to be the ConfigMonitor.

micro-rdk/src/common/ota.rs
```diff
@@ -172,6 +170,17 @@ pub(crate) struct OtaService<S: OtaMetadataStorage> {
...
 impl<S: OtaMetadataStorage> OtaService<S> {
+    pub(crate) async fn stored_metadata(&self) -> OtaMetadata {
+        if !self.storage.has_ota_metadata() {
+            log::info!("no ota metadata currently stored in NVS");
+        }
+
+        self.storage
+            .get_ota_metadata()
+            .inspect_err(|e| log::warn!("failed to get ota metadata from nvs: {}", e))
+            .unwrap_or_default()
```
acmorrow, line 181: Why return a default value rather than an error?

micro-rdk/src/common/config_monitor.rs
```diff
@@ -27,11 +30,14 @@ where
     pub fn new(
         curr_config: Box<RobotConfig>,
         storage: Storage,
+        #[cfg(feature = "ota")] executor: Executor,
```
acmorrow, line 33: nit: inconsistent with how we usually do `cfg` things.

micro-rdk/src/common/config_monitor.rs
```diff
@@ -71,16 +77,47 @@ where
...
+                            "Failed to reset robot config after new config detected: {}",
+                            e
+                        );
+                    } else {
+                        reboot = true;
+                    }
+                }
+
+                if reboot {
+                    // TODO(RSDK-9464): flush logs to app.viam before restarting
```
acmorrow, line 120: This ticket would be another reason to DRY the shutdown handling.

acmorrow (changes requested):

micro-rdk/src/common/ota.rs
mattjperez: I've tried exploring different paths for a day or two:
- wrapping the connection initialization logic into its own function
  - getting the return type right (even with generics) was a rabbit hole of satisfying the compiler and enabling an unstable/nightly features that are used to define a lower-level trait (something with Allocator if I remember right)
- declaring the variables outside the loop
   -  same issue as above
 - using `for i in 0..CONN_RETRY`
   - can't use `break` in a `for-loop`, only in `loop`

I think I tried a couple other things, (like inspect,map_err, ok_and, etc) but it seemed like too much time spend on trying to avoid the messiness and have the correct behavior.

mattjperez: I can move the increment to the top of the loop and set the range accordingly

micro-rdk/src/common/config_monitor.rs
mattjperez: as in, have a global shutdown function that handles log-flushing?

micro-rdk/src/common/conn/viam.rs
mattjperez: the motivation was to get rid of the coredump we see everytime on config change restarts since I'm already modifying the config_monitor's behavior. 

I mention it in the description about partially addressing [RSDK-9594](https://viam.atlassian.net/browse/RSDK-9594), but just in config monitor.

I can look into `grpc::robot_shutdown`, but that depends if you want that done across the repo in RSDK-9594 or just here. Which would you prefer?

micro-rdk/src/common/ota.rs
mattjperez: Would you prefer a field in `OtaService` that the caller checks after calling `update`?

micro-rdk/src/common/ota.rs
mattjperez: It isn't, but shouldn't it be? The underlying calls are essentially I/O, getting data from flash storage. But this is maybe is more of a "should our storage APIs be async?" question.

I'll remove it here though.

micro-rdk/src/common/conn/viam.rs
mattjperez: per offline discussion, changing this back to just use `std::process::exit`. the right changes will be done in RSDK-9594

npmenard (approved):
Ok to merge for testing but not really a fan of the approach here. Having the OTA logic coupled to the config monitor feels clunky and complicated. It may lead to premature cancellation of the OTA process if connection to app.viam.com is lost. It also make any future changes to config monitor harder

acmorrow (approved):

micro-rdk/src/common/config_monitor.rs
```diff
@@ -71,16 +77,47 @@ where
...
+                        .iter()
+                        .find(|&service| service.model == *ota::OTA_MODEL_TRIPLET)
+                    {
+                        match ota::OtaService::from_config(
+                            service,
+                            self.storage.clone(),
+                            self.executor.clone(),
+                        ) {
+                            Ok(mut ota) => match ota.update().await {
+                                Ok(needs_reboot) => reboot = needs_reboot,
```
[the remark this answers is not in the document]
mattjperez: maybe not a likely scenario, but if app.viam becomes unreachable during the download period, at least it would be operating off of the new config at reboot.

we can iterate on it

micro-rdk/src/common/config_monitor.rs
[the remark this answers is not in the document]
mattjperez: if either a) we decide to revert back to a separate ota task or b) if OTA becomes default and we remove the feagure-gate, we could remove this.

micro-rdk/src/common/config_monitor.rs
```diff
@@ -71,16 +77,47 @@ where
...
+                let mut reboot = false;
+
+                #[cfg(feature = "ota")]
+                {
+                    if let Some(service) = config
+                        .services
+                        .iter()
+                        .find(|&service| service.model == *ota::OTA_MODEL_TRIPLET)
+                    {
+                        match ota::OtaService::from_config(
```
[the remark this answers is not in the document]
mattjperez: made a ticket, though may be a non-issue if we revert back to separate ota task after testing. [RSDK-9676](https://viam.atlassian.net/browse/RSDK-9676)

PR #370, opened 2025-01-02
[RSDK-9647] update 8mb partitions

mattjperez:
previous table left 1MB of space available, this recovers that and adds the coredump partition to micro-rdk-server

### DIFF
--- a/micro-rdk-server/esp32/ota_8mb_partitions.csv
+++ b/micro-rdk-server/esp32/ota_8mb_partitions.csv
@@ -1,8 +1,8 @@
 # Name,   Type, SubType, Offset,  Size, Flags
 # Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
-nvs, data, nvs, 0x9000, 0x6000,
-otadata, data, ota, 0xf000, 0x2000,
-phy_init, data, phy, 0x11000, 0x1000,
-ota_0,       app,   ota_0,    0x20000,  3968K,
-ota_1,       app,   ota_1,    0x400000, 3968K,
-coredump,    data,  coredump, 0x7e0000, 128K,
+nvs,	       data,	nvs,	  0x9000,	0x6000,
+otadata,       data,	ota,	  0xf000,	0x2000,
+phy_init,      data,	phy, 	  0x11000, 	0x1000,
+ota_0,         app,     ota_0,    0x20000,      3968K,
+ota_1,         app,     ota_1,    0x400000,     3968K,
+coredump,      data,    coredump, 0x7e0000,     128K,
--- a/micro-rdk-server/esp32/ota_8mb_partitions.csv
+++ b/micro-rdk-server/esp32/ota_8mb_partitions.csv
@@ -1,7 +1,8 @@
 # Name,   Type, SubType, Offset,  Size, Flags
 # Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
-nvs,	       data,	nvs,	  0x9000,	0x6000,
-otadata,       data,	ota,	  0xF000,	0x2000,
-phy_init,      data,	phy, 	  0x11000, 	0x1000,
-ota_0,	       0,	ota_0,	  ,		0x377000,
-ota_1,	       0,	ota_1,	  ,		0x377000,
+nvs,	       data,	nvs,	    0x9000,	  0x6000,
+otadata,     data,	ota,	    0xF000,	  0x2000,
+phy_init,    data,	phy, 	    0x11000, 	0x1000,
+ota_0,	     app,	  ota_0,	  0x20000,	3968K,
+ota_1,	     app,	  ota_1,	  0x400000,	3968K,
+coredump,    data,  coredump, 0x7e0000, 128K,
--- a/micro-rdk-server/esp32/ota_8mb_partitions.csv
+++ b/micro-rdk-server/esp32/ota_8mb_partitions.csv
@@ -1,8 +1,8 @@
 # Name,   Type, SubType, Offset,  Size, Flags
 # Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
-nvs,	       data,	nvs,	    0x9000,	  0x6000,
-otadata,     data,	ota,	    0xF000,	  0x2000,
-phy_init,    data,	phy, 	    0x11000, 	0x1000,
-ota_0,	     app,	  ota_0,	  0x20000,	3968K,
-ota_1,	     app,	  ota_1,	  0x400000,	3968K,
+nvs,         data,  nvs,      0x9000,   0x6000,
+otadata,     data,  ota,      0xF000,   0x2000,
+phy_init,    data,  phy,      0x11000,  0x1000,
+ota_0,       app,   ota_0,    0x20000,  3968K,
+ota_1,       app,   ota_1,    0x400000, 3968K,
 coredump,    data,  coredump, 0x7e0000, 128K,
--- a/micro-rdk-server/esp32/ota_8mb_partitions.csv
+++ b/micro-rdk-server/esp32/ota_8mb_partitions.csv
@@ -1,7 +1,7 @@
 # Name,   Type, SubType, Offset,  Size, Flags
 # Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
 nvs,         data,  nvs,      0x9000,   0x6000,
-otadata,     data,  ota,      0xF000,   0x2000,
+otadata,     data,  ota,      0xf000,   0x2000,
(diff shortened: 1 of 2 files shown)


micro-rdk-server/esp32/ota_8mb_partitions.csv
```diff
@@ -1,7 +1,8 @@
...
 # Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
-nvs,	       data,	nvs,	  0x9000,	0x6000,
-otadata,       data,	ota,	  0xF000,	0x2000,
-phy_init,      data,	phy, 	  0x11000, 	0x1000,
-ota_0,	       0,	ota_0,	  ,		0x377000,
-ota_1,	       0,	ota_1,	  ,		0x377000,
+nvs,         data,  nvs,      0x9000,   0x6000,
+otadata,     data,  ota,      0xF000,   0x2000,
+phy_init,    data,  phy,      0x11000,  0x1000,
+ota_0,       app,   ota_0,    0x20000,  3968K,
```
mattjperez, line 6: 56KB lost to get 0x10000 alignment requirement for app partitions

micro-rdk-server/esp32/ota_8mb_partitions.csv
```diff
@@ -1,7 +1,8 @@
 # Name,   Type, SubType, Offset,  Size, Flags
 # Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
-nvs,	       data,	nvs,	  0x9000,	0x6000,
-otadata,       data,	ota,	  0xF000,	0x2000,
-phy_init,      data,	phy, 	  0x11000, 	0x1000,
-ota_0,	       0,	ota_0,	  ,		0x377000,
-ota_1,	       0,	ota_1,	  ,		0x377000,
+nvs,         data,  nvs,      0x9000,   0x6000,
+otadata,     data,  ota,      0xF000,   0x2000,
```
acmorrow, line 4: nit: we should try to be consistent about upper vs lower for hex digits (see `0x7e`) below.

micro-rdk-server/esp32/ota_8mb_partitions.csv
```diff
@@ -1,7 +1,8 @@
...
+phy_init,    data,  phy,      0x11000,  0x1000,
+ota_0,       app,   ota_0,    0x20000,  3968K,
+ota_1,       app,   ota_1,    0x400000, 3968K,
```
acmorrow, line 7: I also think we should be consistent about whether we use hex or K/M for sizes. I sort of like keeping the offsets in hex, I think. They don't change that often, and things get logged in that format. But sizes are nice to be able to read quickly in human units. Is it weird to have offsets in hex and sizes in K/M?

micro-rdk-server/esp32/ota_8mb_partitions.csv
```diff
@@ -1,7 +1,8 @@
 # Name,   Type, SubType, Offset,  Size, Flags
 # Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
-nvs,	       data,	nvs,	  0x9000,	0x6000,
-otadata,       data,	ota,	  0xF000,	0x2000,
-phy_init,      data,	phy, 	  0x11000, 	0x1000,
-ota_0,	       0,	ota_0,	  ,		0x377000,
-ota_1,	       0,	ota_1,	  ,		0x377000,
+nvs,         data,  nvs,      0x9000,   0x6000,
```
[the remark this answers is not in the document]
mattjperez: When I format in my editor, the formatting looks all messed up on github. so I tried editing directly on github so it shows up properly. I ensured it was using all spaces, no tabs. so yeah, I can't tell which is the correct view (my editor, github, your editor).

templates/project/partitions.csv
```diff
@@ -1,7 +1,8 @@
 # Name,   Type, SubType, Offset,  Size, Flags
 # Note: if you have increased the bootloader size, make sure to update the offsets to avoid overlap
-nvs,      data, nvs, 0x9000    ,        0x8000,
-phy_init, data, phy, 0x11000    ,        0x1000,
-#reserved, data, undefined,0x12000,0x4e000
-factory,  app,  factory, 0x30000,       0x3D0000,
-coredump, data, coredump,,128K
+nvs,         data,  nvs,      0x9000,   0x6000,
```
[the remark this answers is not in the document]
mattjperez: ah, you're right. this was actually the main partition table I was testing changes with because these changes stemmed from improvements to the template workflow. I'll remove it

micro-rdk-server/esp32/ota_8mb_partitions.csv
```diff
@@ -1,7 +1,8 @@
...
+otadata,     data,  ota,      0xF000,   0x2000,
+phy_init,    data,  phy,      0x11000,  0x1000,
+ota_0,       app,   ota_0,    0x20000,  3968K,
+ota_1,       app,   ota_1,    0x400000, 3968K,
+coredump,    data,  coredump, 0x7e0000, 128K,
```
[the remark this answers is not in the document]
mattjperez: I was thinking of removing this file altogether once we make ota a default feature, but to have the correct partition table ready with this PR. I'm good with any ordering though.

so you want me to add just `coredump` to `micro-rdk-server/esp32/partitions.csv` or have it as an exact copy?

micro-rdk-server/esp32/ota_8mb_partitions.csv
mattjperez: I see your point. I like the offset in hex since it's easy to compare with address-space output in logs (I reference it a lot in ota).

For `otadata` at least, there is benefit in having it stay in hex because those are hard requirements according to the [docs](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/ota.html#ota-data-partition) (shown in hex) and would be easy to see if it's misconfigured.

I'm looking now if phy_init has similar requirements, but haven't come across anything yet.

mattjperez: we could also have all sizes in K/M and have a trailing comment for the hex on otadata

micro-rdk-server/esp32/ota_8mb_partitions.csv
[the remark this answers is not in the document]
mattjperez: I'll revert the unchanged lines for sake of diff'ing

micro-rdk-server/esp32/ota_8mb_partitions.csv
mattjperez: > Meanwhile, I'm not even entirely sure we are actually using the phy init, since I don't think we have `CONFIG_ESP_PHY_INIT_DATA_IN_PARTITION` set?

Oh, you're right. It's [off by default](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/kconfig.html?highlight=config_esp_phy_init_data_in_partition#config-esp-phy-init-data-in-partition). I don't think I've looked into `phy_init` before. it's always been in the examples.

We can look into it further. If we truly aren't using it we can:
a) allocate it more to NVS or app partitions
b) keep it as reserved 
c) keep it there so that if/when someone starts using it

micro-rdk-server/esp32/ota_8mb_partitions.csv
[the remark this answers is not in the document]
mattjperez: we can rename this file to `partitions.csv` when we make ota a default feature in the next PR, for sake of keeping the diffs clean

micro-rdk-server/esp32/ota_8mb_partitions.csv
mattjperez: per last meeting, leaving it as-is

npmenard:
> @npmenard - During review of this, I came up with the following alternative partition ordering which seems to yield a much larger NVS partition, eliminates all gaps, and gives us the same OTA sizes. I'd be interested in your thoughts:
> 
> ```
> phy_init,      data,	      phy,	     0x9000,	0x1000,
> coredump,      data,	      coredump,      0x10000, 	128K,
> ota_0,	       0,	      ota_0,	     0x30000,	0x3e0000,
> ota_1,	       0,	      ota_1,	     0x410000,	0x3e0000,
> otadata,       data,	      ota,	     0x7f0000,	0x2000,
> nvs,	       data,	      nvs,	     0x7f2000,  0xe000,
> ```
> 
> Also, I'd like us to decide what should happen to the regular `partitions.csv`, but I'm fine with that change coming as a subsequent PR in the interest of landing this one.

if it works as expected then (and i don't see why it wouldn't) then let's merge that and make it a default for the template too

acmorrow:
I'm taking this over in #382

PR #372, opened 2025-01-07
[RSDK-9365] fix flaky test

mattjperez:
the test has occasionally tripped, poisoning the `global_network_lock` (fixed). 
It's unclear how much time it would be needed to ensure it doesn't trip (or if other manageable factors are involved), but 1sec seems like a still-acceptable upper-bound on a single test. still faster than running the whole CI workflow again.

If other tests with similar timeouts have this issue it might make sense to make the timeout value global for tests.

### DIFF
--- a/micro-rdk/src/common/conn/viam.rs
+++ b/micro-rdk/src/common/conn/viam.rs
@@ -1332,7 +1332,7 @@ mod tests {
                 // await multiple times to give both servers opportunity to process request/response
                 let _ = Timer::after(Duration::from_millis(50)).await;
                 if !cloned_ram_storage.has_robot_credentials() {
-                    // server properly handled response from fake_server
+                    // viam_server properly handled response from fake_app_server
                     break;
                 }
             }
--- a/micro-rdk/src/common/conn/viam.rs
+++ b/micro-rdk/src/common/conn/viam.rs
@@ -1065,7 +1065,7 @@ mod tests {
         rc::Rc,
         sync::{
             atomic::{AtomicI32, Ordering},
-            Arc,
+            Arc, Condvar, Mutex,
         },
         time::Duration,
     };
@@ -1315,7 +1315,15 @@ mod tests {
 
         let mut app = AppServerInsecure::default();
 
+        // condvar to serialize fake-server-task auth and viam_server.run()
+        let pair = Arc::new((Mutex::new(false), Condvar::new()));
+        let pair2 = Arc::clone(&pair);
+
         app.auth_fn = Some(Rc::new(Box::new(move |req| {
+            let (lock, cvar) = &*pair2;
+            let mut started = lock.lock().unwrap();
+            *started = true;
+            cvar.notify_one();
             assert!(req.entity.contains("test-denied"));
             false
         })));
@@ -1324,9 +1332,21 @@ mod tests {
             let _fake_server_task =
                 cloned_exec.spawn(async move { run_fake_app_server(other_clone, app).await });
             let _task = cloned_exec.spawn(async move {
+                let (lock, cvar) = &*pair;
+                for _ in 0..3 {
+                    // lock within loop to drop mutex guard between (a)waits
+                    let mut started = lock.lock().unwrap();
+                    let result = cvar
+                        .wait_timeout(started, Duration::from_millis(10))
+                        .unwrap();
+                    started = result.0;
+                    if *started {
+                        break;
+                    }
+                }
                 viam_server.run().await;
             });
-            let _ = Timer::after(Duration::from_millis(1000)).await;
+            let _ = Timer::after(Duration::from_millis(500)).await;
             assert!(!cloned_ram_storage.has_robot_credentials())
         });
     }
--- a/micro-rdk/src/common/conn/viam.rs
+++ b/micro-rdk/src/common/conn/viam.rs
@@ -1326,7 +1326,7 @@ mod tests {
             let _task = cloned_exec.spawn(async move {
                 viam_server.run().await;
             });
-            let _ = Timer::after(Duration::from_millis(500)).await;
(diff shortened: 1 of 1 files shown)


micro-rdk/src/common/conn/viam.rs
```diff
@@ -1357,7 +1357,7 @@ mod tests {
             let _task = cloned_exec.spawn(async move {
                 viam_server.run().await;
             });
-            let _ = Timer::after(Duration::from_millis(500)).await;
+            let _ = Timer::after(Duration::from_millis(1000)).await;
```
acmorrow, line 1360: I don't see a way you can completely eliminate the possibility of a race between the `_fake_server_task` and the `viam_server.run()` task given what we have now, but you can I think significantly shrink the window. Put some sort of condvar like object into play, and have the `app.auth_fn` signal it when it. Then, wait on it before asserting on the state of `cloned_ram_storage`. That will at least guarantee that the auth endpoint on `run_fake_app_server` was invoked, meaning that the viam server has progressed far enough to have attempted it. You will still need a timeout, but I'll bet a much smaller one than 500 milliseconds will be effective in practice, rather than goign the other way and just doubling the timeout to one second.

A fully deterministic solution would probably require adding hooks into `ViamServer` to allow tests to wait for it to have progressed to certain will known states, and we don't have that now. We probably should at some point.

acmorrow (changes requested):

micro-rdk/src/common/conn/viam.rs
```diff
@@ -1355,6 +1363,18 @@ mod tests {
             let _fake_server_task =
                 cloned_exec.spawn(async move { run_fake_app_server(other_clone, app).await });
             let _task = cloned_exec.spawn(async move {
+                let (lock, cvar) = &*pair;
+                for _ in 0..3 {
+                    // lock within loop to drop mutex guard between (a)waits
```
mattjperez, line 1368: refer to [`On using std::sync::Mutex and tokio::sync::Mutex`](https://tokio.rs/tokio/tutorial/shared-state).

micro-rdk/src/common/conn/viam.rs
```diff
@@ -1346,7 +1346,15 @@ mod tests {
 
         let mut app = AppServerInsecure::default();
 
+        // condvar to serialize fake-server-task auth and viam_server.run()
```
mattjperez, line 1349: code taken from https://doc.rust-lang.org/std/sync/struct.Condvar.html#method.wait_timeout

micro-rdk/src/common/conn/viam.rs
```diff
@@ -1355,6 +1363,18 @@ mod tests {
             let _fake_server_task =
                 cloned_exec.spawn(async move { run_fake_app_server(other_clone, app).await });
             let _task = cloned_exec.spawn(async move {
+                let (lock, cvar) = &*pair;
+                for _ in 0..3 {
```
mattjperez, line 1367: limited to eventually fail if there are legitimate errors

micro-rdk/src/common/conn/viam.rs
mattjperez: it seems like we could, but would need to use `async-compat` since this is a `futures` executor and not a `tokio` one.

> https://docs.rs/async-compat/latest/async_compat/
> Tokio’s types cannot be used outside tokio context, so any attempt to use them will panic. 
>
> Solution: If you apply the [Compat](https://docs.rs/async-compat/latest/async_compat/struct.Compat.html) adapter to a future, the future will manually enter the context of a global tokio runtime. If a runtime is already available via tokio thread-locals, then it will be used. Otherwise, a new single-threaded runtime will be created on demand. That does not mean the future is polled by the tokio runtime - it only means the future sets a thread-local variable pointing to the global tokio runtime so that tokio’s types can be used inside it.

This would look something like
```rs
use async_compat::{Compat, CompatExt};
use tokio::sync::Notify;

fn futures_test() {
    futures::executor::block_on(Compat::new(async {
        notify2.notified().await;
        println!("received notification");
    }))
}
```

Still need to see if this would compile, but given the above info, still want me to try implementing it?

acmorrow: No, if it isn't a drop-in simplification, don't bother.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -1355,6 +1363,18 @@ mod tests {
...
+                    // lock within loop to drop mutex guard between (a)waits
+                    let mut started = lock.lock().unwrap();
+                    let result = cvar
+                        .wait_timeout(started, Duration::from_millis(10))
+                        .unwrap();
+                    started = result.0;
+                    if *started {
+                        break;
```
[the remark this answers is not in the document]
mattjperez: it appears to :man_shrugging: ? tests pass consistently. 
I haven't dove into how auth_fn and everything is _supposed_ to work, so maybe this is 'working' but not for the right reasons.

mattjperez:
from offline discussion:

as the executor is not multithreaded, the condvar isn't doing what we
think it's doing, and is in fact just blocking the whole executor.

in lieu of a hook/channel/flag to determine the state of a given
request/response, we are just using the fact that storage credentials
are erased as the flag that the request was processed
correctly.

the 0..10 loop of 50 milliseconds equates to the original timeout
duration but also gives more opportunities (via the timeout's .await)
for the server and fake_server to make progress

npmenard (approved):
LGTM! just add a comment about the loop to explain what's its doing and why it's written this way

acmorrow (approved):
LGTM. I think we will still want a way to do the sort of waiting that I originally suggested, but for the purposes of stabilizing this test, I'm fine with this approach for now.

PR #374, opened 2025-01-07
bump versions for 0.4.0-rc1

acmorrow:
Compare with https://github.com/viamrobotics/micro-rdk/commit/bb050e53179cd419ca7f4de670142d1a566da4df

- We don't want to update `micro-rdk-ffi/micrordklib-idf-component/CMakeLists.txt` because that's pinned to 0.3.0 per https://viam.atlassian.net/browse/RSDK-9493 until 0.4.0 GA.
- We don't want to update any templates, they should remain pinned to 0.3.3 for now, until 0.4.0 GA.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2704,7 +2704,7 @@ dependencies = [
 
 [[package]]
 name = "micro-rdk"
-version = "0.3.3"
+version = "0.4.0-rc1"
 dependencies = [
  "async-channel 2.3.1",
  "async-executor",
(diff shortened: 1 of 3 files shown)


acmorrow:
Also I created the following two tickets:
- https://viam.atlassian.net/browse/RSDK-9696, so that we stop putting the micro-rdk version into the version used in generated projects / modules. I think that's always been incorrect to do.
- https://viam.atlassian.net/browse/RSDK-9694, so we remember to update the templates as part of 0.4.0 GA.

npmenard (approved):

PR #376, opened 2025-01-08
Rsdk 9697 support multiple nvs

npmenard:
Implements the storage traits for any type we can iterate over, we can then easily pass an array of a vec as a storage to viam server. 
With this something like
`let storage = vec![
            NVSStorage::new("nvs").unwrap(),
            NVSStorage::new("nvs2").unwrap(),
 ];`
Will allow use to store the robot data over multiple NVS 

I didn't want to add an extra type to represent a collection of storage and this way the logic is driven by the underlying storage implementation.

### DIFF
--- a/examples/esp-idf-component/main/main.c
+++ b/examples/esp-idf-component/main/main.c
@@ -45,7 +45,7 @@ static void event_handler(void* arg, esp_event_base_t event_base,
     } else {
       xEventGroupSetBits(s_wifi_event_group, WIFI_FAIL_BIT);
     }
-    ESP_LOGI(TAG,"connect to the AP fail");
+    ESP_LOGI(TAG, "connect to the AP fail");
   } else if (event_base == IP_EVENT && event_id == IP_EVENT_STA_GOT_IP) {
     ip_event_got_ip_t* event = (ip_event_got_ip_t*) event_data;
     ESP_LOGI(TAG, "got ip:" IPSTR, IP2STR(&event->ip_info.ip));
@@ -274,7 +274,7 @@ void app_main(void)
     viam_server_register_c_generic_sensor(viam_ctx, "sensorA", config_A);
 
   if (ret != VIAM_OK) {
-    ESP_LOGE(TAG,"couldn't register sensorA model, error : %i", ret);
+    ESP_LOGE(TAG, "couldn't register sensorA model, error : %i", ret);
     return;
   }
 
@@ -286,47 +286,47 @@ void app_main(void)
   ret = viam_server_register_c_generic_sensor(viam_ctx, "sensorB", config_B);
 
   if (ret != VIAM_OK) {
-    ESP_LOGE(TAG,"couldn't register sensorB model, error : %i", ret);
+    ESP_LOGE(TAG, "couldn't register sensorB model, error : %i", ret);
     return;
   }
 
   ret = viam_server_set_provisioning_manufacturer(viam_ctx, "viam-example");
   if (ret != VIAM_OK) {
-    ESP_LOGE(TAG,"couldn't set manufacturer, error : %i", ret);
+    ESP_LOGE(TAG, "couldn't set manufacturer, error : %i", ret);
     return;
   }
 
   uint8_t mac[8];
   esp_err_t esp_err = esp_efuse_mac_get_default(mac);
   if (esp_err != ESP_OK){
-    ESP_LOGE(TAG,"couldn't get default mac, error : %i", esp_err);
+    ESP_LOGE(TAG, "couldn't get default mac, error : %i", esp_err);
     return;
   }
   char model[50];
   snprintf(model, 50, "esp32-%02X%02X", mac[6],mac[7]);
   ret = viam_server_set_provisioning_model(viam_ctx, model);
   if (ret != VIAM_OK) {
-    ESP_LOGE(TAG,"couldn't set model, error : %i", ret);
+    ESP_LOGE(TAG, "couldn't set model, error : %i", ret);
     return;
   }
 
   ret = viam_server_add_nvs_storage(viam_ctx, "nvs");
   if (ret != VIAM_OK) {
-    ESP_LOGE(TAG,"couldn't set add nvs partition, error : %i", ret);
+    ESP_LOGE(TAG, "couldn't set add nvs partition, error : %i", ret);
     return;
   }
 
   ret = viam_server_add_nvs_storage(viam_ctx, "nvs_other");
   if (ret != VIAM_OK) {
-    ESP_LOGE(TAG,"couldn't set add nvs partition, error : %i", ret);
+    ESP_LOGE(TAG, "couldn't set add nvs partition, error : %i", ret);
     return;
   }
 
-  ESP_LOGI(TAG,"starting viam server\r\n");
+  ESP_LOGI(TAG, "starting viam server\r\n");
 
   xTaskCreatePinnedToCore((void*)viam_server_start, "viam", CONFIG_MICRO_RDK_TASK_STACK_SIZE, viam_ctx, 6, NULL, CONFIG_MICRO_RDK_TASK_PINNED_TO_CORE_1);
 #else
-  ESP_LOGE(TAG,"enable MICRO_RDK_ENABLE_BUILD_LIBRARY ");
+  ESP_LOGE(TAG, "enable MICRO_RDK_ENABLE_BUILD_LIBRARY ");
 #endif
 
 }
--- a/examples/esp-idf-component/main/main.c
+++ b/examples/esp-idf-component/main/main.c
@@ -310,6 +310,18 @@ void app_main(void)
     return;
   }
 
+  ret = viam_server_add_nvs_storage(viam_ctx, "nvs");
+  if (ret != VIAM_OK) {
+    ESP_LOGE(TAG,"couldn't set add nvs partition, error : %i", ret);
+    return;
+  }
+
+  ret = viam_server_add_nvs_storage(viam_ctx, "nvs_other");
+  if (ret != VIAM_OK) {
+    ESP_LOGE(TAG,"couldn't set add nvs partition, error : %i", ret);
+    return;
+  }
+
   ESP_LOGI(TAG,"starting viam server\r\n");
 
   xTaskCreatePinnedToCore((void*)viam_server_start, "viam", CONFIG_MICRO_RDK_TASK_STACK_SIZE, viam_ctx, 6, NULL, CONFIG_MICRO_RDK_TASK_PINNED_TO_CORE_1);
--- a/micro-rdk-ffi/src/ffi/runtime.rs
+++ b/micro-rdk-ffi/src/ffi/runtime.rs
@@ -27,6 +27,7 @@ pub struct viam_server_context {
     registry: Box<ComponentRegistry>,
     provisioning_info: ProvisioningInfo,
     _marker: PhantomData<(*mut u8, PhantomPinned)>, // Non Send, Non Sync
+    storage: Vec<String>,
 }
 
 #[cfg(target_os = "espidf")]
@@ -49,6 +50,7 @@ pub extern "C" fn init_viam_server_context() -> *mut viam_server_context {
         registry,
         provisioning_info,
         _marker: Default::default(),
+        storage: Default::default(),
     }))
 }
 
@@ -188,6 +190,35 @@ pub unsafe extern "C" fn viam_server_register_c_generic_sensor(
     viam_code::VIAM_OK
 }
 
+/// Add an nvs partition to the storage collection
+///
+/// When the viam server starts each partitions (if they exists) will be added to a storage collection
+/// and made available to the server
+///
+/// # Safety
+/// `ctx`, `storage_name` must be valid pointers
+/// may panic if the storage_name cannot be pushed on the vector
+#[no_mangle]
+pub unsafe extern "C" fn viam_server_add_nvs_storage(
+    ctx: *mut viam_server_context,
+    storage_name: *const c_char,
+) -> viam_code {
+    if ctx.is_null() || storage_name.is_null() {
+        return viam_code::VIAM_INVALID_ARG;
+    }
+
+    let ctx = unsafe { &mut *ctx };
+    let name = if let Ok(s) = unsafe { CStr::from_ptr(storage_name) }.to_str() {
+        s
+    } else {
+        return viam_code::VIAM_INVALID_ARG;
+    };
+
+    ctx.storage.push(name.to_owned());
+
+    viam_code::VIAM_OK
+}
+
 #[allow(dead_code)]
 const ROBOT_ID: Option<&str> = option_env!("MICRO_RDK_ROBOT_ID");
 #[allow(dead_code)]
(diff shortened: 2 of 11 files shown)


micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -34,27 +35,38 @@ pub enum NVSStorageError {
...
+    #[error("nvs collection empty")]
+    NVSCollectionEmpty(#[from] EmptyStorageCollectionError),
 }
 
 #[derive(Clone)]
 pub struct NVSStorage {
     // esp-idf-svc partition driver ensures that only one handle of a type can be created
     // so inner mutability can be achieves safely with RefCell
     nvs: Rc<RefCell<EspCustomNvs>>,
-    partition_name: CString,
```
npmenard, line 44: store a copy of the partition rather than the partition name to avoid cloning a CString as we pass the Storage through different par to of the code

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -102,15 +103,15 @@ impl From<CertificateResponse> for TlsCertificate {
 pub trait WifiCredentialStorage {
     type Error: Error + Debug + Into<ServerError>;
     fn has_wifi_credentials(&self) -> bool;
-    fn store_wifi_credentials(&self, creds: WifiCredentials) -> Result<(), Self::Error>;
+    fn store_wifi_credentials(&self, creds: &WifiCredentials) -> Result<(), Self::Error>;
```
npmenard, line 106: I changed some functions of the storage traits to accept reference instead of taking ownership. Otherwise we would have to clone each time until we can find a storage where to store the object . It's not dramatic so happy to revert that change if we feel like it doesn't make sense

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -135,14 +136,18 @@ pub trait OtaMetadataStorage {
...
     fn reset_ota_metadata(&self) -> Result<(), Self::Error>;
 }
 
 pub trait StorageDiagnostic {
     fn log_space_diagnostic(&self);
 }
 
+#[derive(Error, Debug)]
+#[error("empty storage collection")]
+pub struct EmptyStorageCollectionError;
```
npmenard, line 149: this is a special error to handle the cases where a collection is empty

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -158,6 +163,22 @@ struct RAMCredentialStorageInner {
 #[derive(Default, Clone)]
 pub struct RAMStorage(Rc<Mutex<RAMCredentialStorageInner>>);
 
+#[derive(Error, Debug)]
+pub enum RAMStorageError {
```
npmenard, line 167: add an error type for RAMStorage to make it easier to test

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -174,42 +195,76 @@ impl RAMStorage {
...
+where
+    for<'a> &'a Iterable: IntoIterator<Item = &'a Storage>,
+    Storage::Error: From<EmptyStorageCollectionError>,
+{
+    type Error = Storage::Error;
+    fn has_ota_metadata(&self) -> bool {
+        self.into_iter().any(OtaMetadataStorage::has_ota_metadata)
+    }
+    fn get_ota_metadata(&self) -> Result<OtaMetadata, Self::Error> {
+        self.into_iter().fold(
```
npmenard, line 231: the idea here  is to use fold to call `get_ota_metadata` until Ok is returned. If all storage in a collection return an error only the last error is reported.
We don't have an iterator that behave like `try_fold` but actually short-circuit on the first Ok, so we will iterate through he whole collection even after a `get_ota_metadata` has succeeded.

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -174,42 +195,76 @@ impl RAMStorage {
...
 }
+#[cfg(feature = "ota")]
+impl<Iterable, Storage: OtaMetadataStorage> OtaMetadataStorage for Iterable
+where
+    for<'a> &'a Iterable: IntoIterator<Item = &'a Storage>,
+    Storage::Error: From<EmptyStorageCollectionError>,
+{
+    type Error = Storage::Error;
+    fn has_ota_metadata(&self) -> bool {
+        self.into_iter().any(OtaMetadataStorage::has_ota_metadata)
```
npmenard, line 228: short-circuit on the first true

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -174,42 +195,76 @@ impl RAMStorage {
...
     }
     fn reset_ota_metadata(&self) -> Result<(), Self::Error> {
         let _ = self.0.lock().unwrap().ota_metadata.take();
         Ok(())
     }
 }
+#[cfg(feature = "ota")]
+impl<Iterable, Storage: OtaMetadataStorage> OtaMetadataStorage for Iterable
+where
+    for<'a> &'a Iterable: IntoIterator<Item = &'a Storage>,
```
npmenard, line 223: will only work on collection that implement  `impl<'a, T,> IntoIterator for &'a  ...`

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -174,46 +195,82 @@ impl RAMStorage {
...
+    fn get_ota_metadata(&self) -> Result<OtaMetadata, Self::Error> {
+        self.into_iter().fold(
+            Err::<_, Self::Error>(EmptyStorageCollectionError.into()),
+            |val, s| val.or_else(|_| s.get_ota_metadata()),
+        )
+    }
+    fn reset_ota_metadata(&self) -> Result<(), Self::Error> {
+        self.into_iter().fold(
+            Err::<_, Self::Error>(EmptyStorageCollectionError.into()),
+            |val, s| val.or(s.reset_ota_metadata()),
```
npmenard, line 239: taking advantage of the eagerness of or might not be the best approach

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -174,46 +195,82 @@ impl RAMStorage {
...
+        let creds: RobotCredentials = cfg.clone().try_into()?;
         let mut inner_ref = self.0.lock().unwrap();
         let _ = inner_ref.robot_creds.insert(creds);
         Ok(())
     }
     fn get_robot_credentials(&self) -> Result<RobotCredentials, Self::Error> {
         let inner_ref = self.0.lock().unwrap();
-        let cfg = inner_ref.robot_creds.clone().unwrap_or_default().clone();
-        Ok(cfg)
+        log::info!("get robot creds");
```
acmorrow, line 264: Stray log, or `debug!`?

As an aside, I do miss some of the logging we used to have pre #316 that gave a little more context about progression through the startup sequence, whether we were using cached credentials or configs, etc. I'm usually for "silence is golden" but in this case that's very useful information to understand how things are coming up, especially in offline mode. Would you be OK with a ticket to re-introduce some of that logging? Or did you not propagate into the refactored world because you found it too chatty?

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -174,46 +195,82 @@ impl RAMStorage {
...
+        inner_ref
+            .robot_creds
+            .clone()
+            .ok_or(RAMStorageError::NotFound)
     }
     fn reset_robot_credentials(&self) -> Result<(), Self::Error> {
         let mut inner_ref = self.0.lock().unwrap();
         let _ = inner_ref.robot_creds.take();
+        log::info!("reset robot creds");
```
acmorrow, line 273: As above

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -304,12 +456,140 @@ impl WifiCredentialStorage for RAMStorage {
...
+    fn test_vec_storage() {
+        let ram1 = RAMStorage::new();
+        let ram2 = RAMStorage::new();
+        let ram3 = RAMStorage::new();
+        let v: Vec<RAMStorage> = vec![ram1.clone(), ram2.clone(), ram3.clone()];
+
+        assert!(!v.has_robot_credentials());
+        assert!(ram2
+            .store_robot_credentials(&CloudConfig {
+                app_address: "http://downloadramstorage.org".to_owned(),
```
acmorrow, line 542: maybe `.example.org`

micro-rdk/src/common/credentials_storage.rs
acmorrow: Can you elaborate a bit? I think I know what you are saying, but I'd rather not guess.

acmorrow (approved):
This is neat, LGTM, a few small comments. I'd like to think about it a bit tomorrow before it merges so I can make sure I totally understand how the various folds are working and think about whether there are any tricky edge cases.

acmorrow:
How do you envision configuring the space of available NVS, for, say, callers of `viam_server_start`?

micro-rdk/src/common/credentials_storage.rs
npmenard: As opposed to `or_else` `or` will always evaluate a function in argument if there is one (because it doesn't actually accept a closure) before considering replacing `val` hence the eagerness. We use to our advantage to make sure all copies of an object are erased through all storage. In theory we shouldn't need this since we use or_else elsewhere so only on copy should be in the storage array at anytime.

micro-rdk/src/common/credentials_storage.rs
npmenard: stray log i will remove it.
Yes let's make a ticket for some logging housekeeping

micro-rdk/src/common/credentials_storage.rs
npmenard: yes that was for the joke :P

npmenard:
> How do you envision configuring the space of available NVS, for, say, callers of `viam_server_start`?

What do you mean?

micro-rdk/src/common/credentials_storage.rs
acmorrow: I have one small concern here. I don't have a fully worked counterexample, but I wonder a bit about using something like a HashSet as the type backing IntoIterator, because the iteration order isn't deterministic.

micro-rdk/src/common/credentials_storage.rs
acmorrow: Done, in https://viam.atlassian.net/browse/RSDK-9715. Feel free to add more things you'd like to see added to a logging pass.

acmorrow:
> > How do you envision configuring the space of available NVS, for, say, callers of `viam_server_start`?
> 
> What do you mean?

As discussed in-person, I was interested in how you saw the ability to configure the selection of available partitions expressed in the FFI.

acmorrow (approved):
LGTM, though I'm interested in your thoughts about my iteration order worry.

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -231,8 +231,15 @@ impl RobotConfigurationStorage for NVSStorage {
 
     fn store_robot_credentials(&self, cfg: &CloudConfig) -> Result<(), Self::Error> {
         self.set_string(NVS_ROBOT_SECRET_KEY, &cfg.secret)?;
-        self.set_string(NVS_ROBOT_ID_KEY, &cfg.id)?;
-        self.set_string(NVS_ROBOT_APP_ADDRESS, &cfg.app_address)?;
+        self.set_string(NVS_ROBOT_ID_KEY, &cfg.id).or_else(|err| {
+            let _ = self.reset_robot_credentials();
```
acmorrow, line 235: - By using `reset_robot_credentials`, I think there is now a requirement that it erase keys in the same order that `store_robot_credentials` writes them, since a failed `erase_key` ends the method. Maybe it is better to explicitly just unset the keys as acted on so far in the `store` methods rather than relying on ordering in a different method? Accumulate some guards?
- ⬇️  - Is it a bug that `reset_robot_credentials` doesn't clear `NVS_ROBOT_APP_ADDRESS`?

micro-rdk-ffi/src/ffi/runtime.rs
```diff
@@ -284,7 +315,22 @@ pub unsafe extern "C" fn viam_server_start(ctx: *mut viam_server_context) -> via
         #[cfg(not(target_os = "espidf"))]
         let storage = micro_rdk::common::credentials_storage::RAMStorage::default();
         #[cfg(target_os = "espidf")]
-        let storage = micro_rdk::esp32::nvs_storage::NVSStorage::new("nvs").unwrap();
+        //let storage =  {micro_rdk::esp32::nvs_storage::NVSStorage::new("nvs").unwrap()};
```
acmorrow, line 318: Remove commented line

micro-rdk-ffi/src/ffi/runtime.rs
```diff
@@ -257,7 +288,7 @@ pub unsafe extern "C" fn viam_server_start(ctx: *mut viam_server_context) -> via
         use micro_rdk::common::credentials_storage::RAMStorage;
         use micro_rdk::common::credentials_storage::RobotConfigurationStorage;
         use micro_rdk::proto::provisioning::v1::CloudConfig;
-
+        log::info!("when the FFI was built a robot config was supplied, defaulting to this robot");
```
acmorrow, line 291: Maybe take this out for now but add a `TODO(RSDK-9715)` which is my init logging ticket so we can do this in coordination with other startup logging changes?

examples/esp-idf-component/main/main.c
```diff
@@ -310,6 +310,18 @@ void app_main(void)
     return;
   }
 
+  ret = viam_server_add_nvs_storage(viam_ctx, "nvs");
+  if (ret != VIAM_OK) {
+    ESP_LOGE(TAG,"couldn't set add nvs partition, error : %i", ret);
```
acmorrow, line 315: space after `,` here and below please

acmorrow (changes requested):

npmenard:
> > LGTM, though I'm interested in your thoughts about my iteration order worry.
> 
> I think the fix you've made here about ensuring that writes of multiple keys are all-or-nothing in a given partition is necessary. But I'm still a little concerned about iteration order determinism over the set of storages. Per my slack message:
> 
> > The situation I had in mind looked something like this: you have nvs partitions named A and B, but they are referenced via a HashSet, so the iteration order is indeterminate. If the firmware has some sort of unconditional write on startup (e.g. writing the robot config), then if you come up with A as first iterated once, and B first iterated once, then it will end up stored in both. If those values differ, and then you come up in a mode where you need to consume the value (e.g. offline), you may get either value depending on whether you iterate A or B first, and potentially one of them is out of date.
> > I haven’t totally convinced myself that this is possible or a real problem, but it has me a little hesitant.
> > Maybe it can easily be fixed by iterating in sorted order, based on partition name or address?
> > Or by requiring the partitions to be paired with “priorities”, so we can say “prefer the nvs partition, but spill to config if needed”. Then just iterate sorted on prio, and it wouldn’t matter if the container supporting iteration was non-deterministic. Setting prio though wouldn’t be 100% as you could accidentally set two with prio N and then you are back to non-determinism.

I think this can be an issue when storing config or certificate only since we don't explicitly erase these objects when storing them during viam server initialization. We implicitly expect the storage driver to overwrite. 
Everywhere else we either check if an object is in storage or erase it unilaterally 
We can add a ticket to change the initialization behavior to do an erase before storing but that would increase the wear on the flash.

PR #377, opened 2025-01-08
[CONSULT-469] add initial code for field parsing

gvaradarajan:


micro-rdk-nmea/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+version.workspace = true
+rust-version.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+micro-rdk = { path = "../micro-rdk", features = ["native"] }
```
gvaradarajan, line 14: To my great annoyance, we've gone back to a state where micro-rdk cannot be built as a feature-less library. I'll make a ticket for this and link it here

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -0,0 +1,356 @@
...
+where
+    T: Lookup,
+{
+    type FieldType = T;
+
+    fn read_from_data(
+        &self,
+        data: &[u8],
+        start_idx: usize,
+    ) -> Result<(usize, Self::FieldType), NmeaParseError> {
```
acmorrow, line 114: Same thoughts as above on signature here regarding `self` and the `usize` return.

acmorrow (changes requested):
Done with a first pass. I haven't taken too close a look at the implementation details of the bit bashing yet, as I'd like to fully understand how the types work together and I'm not quite there yet, but here are my notes so far.

micro-rdk/src/common/app_client.rs
```diff
@@ -63,7 +63,6 @@ pub enum AppClientError {
     AppConfigHeaderDateMissingError,
     #[error(transparent)]
     AppGrpcClientError(#[from] GrpcClientError),
-    #[cfg(feature = "data")]
```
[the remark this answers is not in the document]
gvaradarajan: Discussed offline, I found this issue when I tried to build featureless micro-RDK. The issue is that config monitor now uses this error, so trying to build without the data feature fails

micro-rdk-nmea/src/parse_helpers/enums.rs
```diff
@@ -0,0 +1,328 @@
+pub trait Lookup: Sized {
+    fn from_value(val: u32) -> Self;
```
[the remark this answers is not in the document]
gvaradarajan: So, there are lookup fields of various sizes in the NMEA 2000 CAN specification. Not wanting to handle an excessive amount of cases, I opted for a size that was sure to never be exceeded. I don't think the potential usefulness outweighs having to implement `from_value` for every numeric type.

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -0,0 +1,356 @@
...
+pub trait FieldReader {
+    type FieldType;
+
+    /// Takes a byte slice (`data`) and starting index. Returns a tuple consisting of
+    /// the next index to read from the byte sequence and the parsed value.
+    fn read_from_data(
+        &self,
+        data: &[u8],
+        start_idx: usize,
+    ) -> Result<(usize, Self::FieldType), NmeaParseError>;
```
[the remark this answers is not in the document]
gvaradarajan: - is there a reason we want it consumed?
- I'm not convinced this is easy to accomplish because something that implements the FieldReader trait is unaware of the data type of the next field

micro-rdk-nmea/src/parse_helpers/enums.rs
```diff
@@ -0,0 +1,329 @@
...
+    (0, Gps, "GPS"),
+    (1, Glonass, "GLONASS"),
+    (2, GpsGlonass, "GPS+GLONASS"),
+    (3, GpsSbasWaas, "GPS+SBAS/WAAS"),
+    (4, GpsSbasWaasGlonass, "GPS+SBAS/WAAS+GLONASS"),
+    (5, Chayka, "Chayka"),
+    (6, Integrated, "integrated"),
+    (7, Surveyed, "surveyed"),
+    (8, Galileo, "Galileo"),
+    CouldNotParse
```
npmenard, line 306: `CouldNotParse` feels like it should be an error vs an actual field. If we want to represent an Unknown LookupField then I would suggest naming it `UnknowLookupField` or alike

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -0,0 +1,362 @@
...
+                (data[start_idx] >> shift) as u32
+            }
+            x if x < 16 => {
+                let shift = 16 - x;
+                let raw_val = u16::from_le_bytes(data_slice.try_into()?);
+                (raw_val >> shift) as u32
+            }
+            x if x < 32 => {
+                let shift = 32 - x;
+                let raw_val = u32::from_le_bytes(data_slice.try_into()?);
```
npmenard, line 138: I think this will not work for `bit_size in [16;23]` as (self.bit_size / 8) == 3 and `bit_size ==32` as (self.bit_size / 8) == 5

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -0,0 +1,362 @@
...
+                raw_val >> shift
+            }
+            _ => unreachable!("lookup field raw value cannot be more than 32 bits"),
```
npmenard, line 141: should not instantiate LookupField if bit_size > 32

npmenard (changes requested):
Done with first pass;
I don't like having function signature be like `Result<(usize, Self::FieldType),...>` it means wee need logic to keep track where we are in the stream of bytes which i believe would be error prone.
Did you consider using a Reader at the bit level? perhaps https://docs.rs/bitreader/latest/bitreader/ can be an option

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -0,0 +1,362 @@
...
+/// T is a number type supported by `NumberField`.
+pub struct ArrayField<T, const N: usize>(PhantomData<NumberField<T>>);
+
+impl<T, const N: usize> ArrayField<T, N> {
+    pub fn new() -> Self {
+        Self(Default::default())
+    }
+}
+
+impl<T, const N: usize> FieldReader for ArrayField<T, N>
```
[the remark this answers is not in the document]
gvaradarajan: I don't believe that's true, I think it will only code gen for each actually referenced (T, N), but I can double-check

micro-rdk-nmea/src/parse_helpers/enums.rs
```diff
@@ -0,0 +1,329 @@
...
+/// (raw number value, name of enum instance, string representation)
+macro_rules! define_nmea_enum {
+    ( $name:ident, $(($value:expr, $var:ident, $label:expr)),*, $default:ident) => {
+        #[derive(Copy, Clone, Debug)]
+        pub enum $name {
+            $($var),*,
+            $default
+        }
+
+        impl From<u32> for $name {
```
acmorrow, line 14: Should this be `TryFrom`? Then you wouldn't need the `$default:ident` or associated enum member and could just return an error.

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -0,0 +1,404 @@
...
+use super::{
+    enums::NmeaEnumeratedField,
+    errors::{NmeaParseError, NumberFieldError},
+};
+
+/// Cursor over a byte slice of data. The cursor can be moved by either consuming
+/// child-instances of `DataRead` or calling the `advance` function
+pub struct DataCursor<'a> {
+    data: &'a [u8],
+    bit_position: Rc<AtomicUsize>,
```
acmorrow, line 19: I don't understand why atomics are making an appearance.

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -0,0 +1,404 @@
...
+
+impl<'a> DataCursor<'a> {
+    pub fn new(data: &'a [u8]) -> Self {
+        Self {
+            data,
+            bit_position: Rc::new(AtomicUsize::new(0)),
+        }
+    }
+
+    fn read(&self, bits: usize) -> Result<DataRead, NmeaParseError> {
```
acmorrow, line 42: If you call `read` twice without dropping the `DataRead` object that resulted from the first `read` before the second call to `read`, what happens?

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -0,0 +1,404 @@
...
+            let start_byte_position = bit_position / 8;
+            Ok(DataRead {
+                data: &self.data[start_byte_position..end_byte_position],
+                bit_position: self.bit_position.clone(),
+                bit_size: bits,
+            })
+        }
+    }
+
+    pub fn advance(&self, bits: usize) -> Result<(), NmeaParseError> {
```
acmorrow, line 57: Could `advance` be implemented by calling `read` and throwing away the result?

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -0,0 +1,404 @@
...
+    T: NmeaEnumeratedField,
+{
+    type FieldType = T;
+
+    fn read_from_cursor(&self, cursor: &DataCursor) -> Result<Self::FieldType, NmeaParseError> {
+        let data_read = cursor.read(self.bit_size)?;
+        let enum_value = match self.bit_size {
+            x if x <= 8 => Ok::<u32, NmeaParseError>(u8::try_from(data_read)? as u32),
+            x if x <= 16 => Ok::<u32, NmeaParseError>(u16::try_from(data_read)? as u32),
+            x if x <= 32 => Ok::<u32, NmeaParseError>(u16::try_from(data_read)? as u32),
```
acmorrow, line 189: Should this be `u32::try_from`?

acmorrow (commented):
Took a pass through. It mostly makes sense, but I'll need to spend some more time on `generate_number_field_readers` to really get what is going on there. Some questions and suggestions as well.

micro-rdk-nmea/src/parse_helpers/parsers.rs
gvaradarajan: I had a reason, but I have since forgotten it. I'll just remove it, it's not like the rest of the code is thread-safe anyway

micro-rdk-nmea/src/parse_helpers/parsers.rs
gvaradarajan: It will be very bad. They could both read from the same starting place but potentially different ending places. Even if they do read to the same end position, when they're both dropped the bit position will advance twice as far as it should. Should I protect against this?

micro-rdk-nmea/src/parse_helpers/parsers.rs
gvaradarajan: It could, is that preferred?

acmorrow: I think so, because then `advance` is very very simple, and the complexity lives only in `read`.

micro-rdk-nmea/src/parse_helpers/parsers.rs
acmorrow: Ideally, you would leverage the type system to make such a situation impossible. Just as a for instance, you could do something similar to how ScopeGuard works, where calling `read` on a `DataCursor` consumes the `DataCursor` and hands you a `DataRead` object that holds an inner `DataCursor`, which you must recover via `into_inner` or similar, which in turn consumes the `DataRead` and restores the `DataCursor`. That way, you go back and forth between them, but you can't end up with two `DataRead` objects obtained on the same `DataCursor` object. That might also let you get rid of the `Rc`? I'll bet @npmenard would have some even better ideas about how to structure it. This is just me thinking aloud.

micro-rdk-nmea/src/parse_helpers/parsers.rs
gvaradarajan: Ok I've remembered again. Rust will not let me mutate the content of an `Rc` in the drop for `DataRead` because it doesn't consider it safe to mutate it when the reference count isn't 0. I have to get around this with the interior mutability of `AtomicUsize`

micro-rdk-nmea/src/parse_helpers/enums.rs
gvaradarajan: So, I had this be the behavior because the gonmea library does not fail on an unrecognized lookup value

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -0,0 +1,411 @@
...
+
+                    let start_idx = value.bit_position.load(Ordering::SeqCst) % 8;
+                    let end_idx = start_idx + value.bit_size;
+                    let mut bit_vec = bits[start_idx..end_idx].to_bitvec();
+                    if bit_vec.len() != (max_size * 8) {
+                        // the last byte is incomplete, we reverse the remaining bits of the
+                        // incomplete byte so it can be padded with zeros, then we reverse the bits
+                        // of the now complete last byte for the proper bit order
+                        let last_bit_start = bit_vec.len() - (value.bit_size % 8);
+                        let _ = &bit_vec[last_bit_start..].reverse();
```
npmenard, line 130: what is the purpose of this operation? for example if I  want to read 12bit on a 16bit number then last_bit_start would be 8 so the reverse operation will affect bits [8..12]. I think this ends up doing something like shift the left most bits right overwriting on the way. 
Also different value of bit_size will change the amount of bits reversed can that affect reading numbers?

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -0,0 +1,411 @@
...
+
+/// Cursor over a byte slice of data. The cursor can be moved by either consuming
+/// child-instances of `DataRead` or calling the `advance` function
+pub struct DataCursor<'a> {
+    data: &'a [u8],
+    bit_position: Rc<AtomicUsize>,
+}
+
+struct DataRead<'a> {
+    data: &'a [u8],
```
npmenard, line 23: should own the buffer

npmenard (commented):
Done with first pass, i have some questions about the bit manipulations as I don't understand fully what's going on.

micro-rdk-nmea/src/parse_helpers/parsers.rs
gvaradarajan: That is what I'm trying to do, yes. The reversing is just to be able to pad the bits with 0 from the front. So, if the overflow bits are 1110, I want to make it so the last byte is 00001110. To do that I reverse just the last 4 bits in place (0111), add the padding (01110000), then reverse the last byte to restore the bit order (00001110)

gvaradarajan: are suggesting I could just right shift the last byte by however many bits are remaining?

npmenard: Not really. In my interpretation of the code (i think it's true for both version) if we have a data like this" `let data : [u8;2] = [0x0,0x10]` Then using a `NumberField<u16>` to read it will yield different result depending on whether we read 12,13,14,15 or 16 bits. This is because the amount of shifted bytes depend on the number of bits read, so the second byte of the array will be shifted differently like `0b010000 -> 0b000010` if bit read is 13 (shift 3) or like ` 0b010000 -> 0b001000` if bit read are 14 (shift 2). If this is how NMEA fields are parsed then we should just have an explanation so the code can be interpreted properly

gvaradarajan: This is how the gonmea library appears to do it, should I try double-checking the canboat C code?

npmenard (approved):
LGTM

PR #380, opened 2025-01-13
[RSDK-9497] [RSDK-9696] Release 0.4.0

acmorrow:


templates/module/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "{{project-name}}"
-version = "0.3.3"
+version = "0.0.1"
```
acmorrow, line 3: It didn't make much sense to me that we were generating projects and modules that had the same version as the micro-rdk from which they were generated. If you disagree I can drop this out easily enough.

I'd filed it as https://viam.atlassian.net/browse/RSDK-9696 but we hadn't got around to triaging it yet. Still, this seems as good a time as any, and it will mean one less version we need to hand-edit on each release (until we have a more automated release process).

micro-rdk-ffi/micrordklib-idf-component/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 idf_component_register(INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}")
 idf_build_set_property(INCLUDE_DIRECTORIES "${CMAKE_CURRENT_BINARY_DIR}/assets" APPEND)
 
-set(LIBMICRORDK_VERSION v0.3.0)
+set(LIBMICRORDK_VERSION v0.4.0)
```
acmorrow, line 4: This is https://viam.atlassian.net/browse/RSDK-9497, undoing the pin previously set in https://viam.atlassian.net/browse/RSDK-9493

templates/module/Cargo.toml
npmenard: I agree.

npmenard (approved):
LGTM

PR #382, opened 2025-01-14
[RSDK-9647] Standardize on 8MB OTA capable partition scheme

acmorrow:
Standardize on a new partition scheme using two OTA segments at 3968k each, 80k NVS, and 128k coredump.


npmenard (approved):
LGTM

PR #387, opened 2025-01-22
[RSDK-9774] add canary OTA update workflow,script to CI 

mattjperez:


.github/workflows/canary_ota.yml
```diff
@@ -0,0 +1,54 @@
+name: Update ESP32 Canary OtaServiceConfig
+
+on:
+  workflow_dispatch: #TODO(1): change to on:push:tags:'*'
+    pull_request:
+      branches:
+        - main
```
mattjperez, line 7: TODO: change this before merging so that it's accessible afterwards

canary/canary_ota.py
```diff
@@ -0,0 +1,74 @@
...
+        # assumes only one such service exists
+        if service["model"] == "ota_service":
+            service["attributes"]["url"] = url_target
+            service["attributes"]["version"] = tag_name            
+            service_updated = True
+            print(f"updating OtaServiceConfig to `{service}`")
+            break
+
+    if not service_updated:
+        viam_client.close()
```
gvaradarajan, line 56: I thought we were going to post via Slack if the update failed CI side, are we not doing that?

canary/canary_ota.py
```diff
@@ -0,0 +1,74 @@
...
+        sys.exit("failed to find or update service ota config")
```
gvaradarajan, line 57: should we raise an exception in order to mark the job as failed? Does `sys.exit` accomplish this as well?

canary/canary_ota.py
```diff
@@ -0,0 +1,74 @@
...
+        robot_part_id=robot_part.id,
+        name=robot_part.name,
+        robot_config=updated_config
+    )
+
+    # retrieve new config to verify
+    robot_part = await cloud.get_robot_part(robot_part_id=part_id)
+    for service in robot_part.robot_config["services"]:
+        if service["model"] == "ota_service":
+            print(f"OtaServiceConfig after updating: `{service}`")
```
gvaradarajan, line 69: I think we should re-check the tag name here and raise an error if there is a mismatch

gvaradarajan (changes requested):
looks mostly good, had some questions

canary/canary_ota.py
mattjperez: It will be when I do the followup slack work. Part of RSDK-9775 https://viam.atlassian.net/browse/RSDK-9775

.github/workflows/canary_ota.yml
```diff
@@ -0,0 +1,51 @@
+name: Update ESP32 Canary OtaServiceConfig
+
+on:
+  workflow_dispatch: #TODO(1): change to on:push:tags:'*'
```
acmorrow, line 4: Before merge? All tags? Just releases?

acmorrow (approved):
LGTM

canary/canary_ota.py
mattjperez: also, to re-iterate, this is not the final state of these files. need to get them into `main` first to make sure the flow even works to validate before moving to the next steps

canary/canary_ota.py
mattjperez: I believe so, but can't confirm until it's actually in main and runs. 

I can:
a. a keep it as-is and we find out eventually if it works properly (can verify easily by removing the ota service from app before running), change it to exception if not
b. change it to exception 

up to you. let me know which you prefer

canary/canary_ota.py
mattjperez: added

canary/canary_ota.py
mattjperez: added

gvaradarajan (approved):
LGTM

PR #390, opened 2025-01-27
[RSDK-9734] Release 0.4.1-rc1 and reduce number of release touchpoints

acmorrow:
This is the commit I would like to tag as 0.4.0-rc1.

Along with the usual update of version numbers, there are some other simplifications here: 

- It starts using the tag name rather than a git hash in the template configs. This will mean we won't need a post release commit to update the templates with the now-known SHA1 of the release commit.
- It moves both templates to use the same style of declaration for the micro-rdk dependency.
- The `micro-rdk-ffi/micrordklib-idf-component/CMakeLists.txt` infrastructure now *derives* the version information from the version metadata in `micro-rdk-ffi/micrordklib-idf-component/idf_component.yml` by way of the `COMPONENT_VERSION` ESP-IDF CMake variable. This eliminates a duplicative touchpoint for releases.
- The version number in `micro-rdk-ffi/micrordklib-idf-component/idf_component.yml` is being advanced to reflect the RC. Downstream consumers of `micrordklib-idf-component` will need to explicitly specify a version when adding `micro-rdk` to their`idf_components.yml` to avoid pulling the RC. But they should be doing that anyway.

### DIFF
--- a/micro-rdk-ffi/micrordklib-idf-component/CMakeLists.txt
+++ b/micro-rdk-ffi/micrordklib-idf-component/CMakeLists.txt
@@ -1,15 +1,14 @@
 idf_component_register(INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}")
 idf_build_set_property(INCLUDE_DIRECTORIES "${CMAKE_CURRENT_BINARY_DIR}/assets" APPEND)
 
-set(LIBMICRORDK_VERSION v0.4.0)
-set(LIBMICRORDK_URL https://github.com/viamrobotics/micro-rdk/releases/download/${LIBMICRORDK_VERSION}/micro-rdk-lib.zip)
+set(LIBMICRORDK_URL https://github.com/viamrobotics/micro-rdk/releases/download/${COMPONENT_VERSION}/micro-rdk-lib.zip)
 set(LIBMICRORDK_PATH ${CMAKE_BINARY_DIR}/import/micro-rdk-lib.zip)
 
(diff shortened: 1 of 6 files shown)


templates/module/Cargo.toml
```diff
@@ -6,8 +6,13 @@ authors = ["{{authors}}"]
 resolver = "2"
 rust-version = "1.83"
 
-[dependencies]
-micro-rdk = {git = "https://github.com/viamrobotics/micro-rdk.git", features = ["{{mcu}}"], version = "0.4.0", rev = "380a098"}
+[dependencies.micro-rdk]
+git = "https://github.com/viamrobotics/micro-rdk.git"
+version = "0.4.1-rc1"
+rev = "v0.4.1-rc1"
```
mattjperez, line 12: will these rev fields still be necessary? would be great if we can move away from this to just reference a specific version. I don't think I've ever seen another rust project that does this for their templates.

micro-rdk-ffi/micrordklib-idf-component/CMakeLists.txt
```diff
@@ -1,15 +1,14 @@
 idf_component_register(INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}")
 idf_build_set_property(INCLUDE_DIRECTORIES "${CMAKE_CURRENT_BINARY_DIR}/assets" APPEND)
 
-set(LIBMICRORDK_VERSION v0.4.0)
-set(LIBMICRORDK_URL https://github.com/viamrobotics/micro-rdk/releases/download/${LIBMICRORDK_VERSION}/micro-rdk-lib.zip)
+set(LIBMICRORDK_URL https://github.com/viamrobotics/micro-rdk/releases/download/${COMPONENT_VERSION}/micro-rdk-lib.zip)
```
mattjperez, line 4: nice

mattjperez (approved):
LGTM, with a question

templates/module/Cargo.toml
acmorrow: I'd like that too, but it doesn't seem to be possible right now. This at least eliminates the need for the post-release BUMP commit, so I think it is a net win.

gvaradarajan: I am interested in the answer to this question

gvaradarajan (approved):

PR #391, opened 2025-01-27
Make 0.4.1-rc2 due to Cargo.lock issue post-merge for rc1

acmorrow:
@mattjperez and @gvaradarajan - Sorry, we had a race condition on -rc1. I had to merge above #386, but that had a stale reference to 0.4.0 in the lock file:

- https://github.com/viamrobotics/micro-rdk/actions/runs/12997311308/job/36248023226
- https://github.com/viamrobotics/micro-rdk/blob/main/Cargo.lock#L2843

So we need an -rc2, and this is that change, along with a run of `cargo update` to make sure that we have all dependencies correct.

Meanwhile, I noticed that we have a `Cargo.lock` in the project template. @mattjperez and I discussed, and this doesn't make sense to me. We shouldn't be pinning the versions of things in a generated project - that's up to the project maintainer to manage their `Cargo.lock`, and we aren't well positioned to create a good one for them. The one in the project template was also stale, referring to Micro-RDK 0.3.3: https://github.com/viamrobotics/micro-rdk/blob/main/templates/project/Cargo.lock#L1414.

I've gone ahead and removed it here.


gvaradarajan (approved):
LGTM, I like the removal of Cargo.lock from the project template

mattjperez (approved):
LGTM, ran a test check in case, all passing

PR #392, opened 2025-01-27
[CONSULT-469] - macro support for array and repeating fields in NMEA

gvaradarajan:
Also contains the following fixes 
- adjustment for number field parsing (it turns out I was somewhat mistaken on the format)
- incorrect advancement of the cursor for when a field is annotated with the `offset` property


micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -330,11 +332,10 @@ mod tests {
         let data_vec: Vec<u8> = vec![154, 6, 125, 179, 152, 113];
         let mut cursor = DataCursor::new(data_vec);
 
-        // 154 is 10011010, reading the first 3 bits should yield 100 = 4
+        // 154 is 10011010, reading the last 3 bits should yield 010 = 4
```
acmorrow, line 335: Comment is stale

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -306,8 +308,8 @@ mod tests {
         assert!(cursor.read(16).is_ok());
         let res = reader.read_from_cursor(&mut cursor);
         assert!(res.is_ok());
-        // 125 = 01111101, first four bits as byte => 00000111 = 7
-        assert_eq!(res.unwrap(), 7);
+        // 125 = 01111101, last four bits as byte => 00001101 = 7
```
acmorrow, line 311: Comment is stale

micro-rdk-nmea-macros/src/composition.rs
```diff
@@ -141,12 +157,55 @@ impl PgnComposition {
...
+            }
+        }
+    }
+}
+
+fn get_read_statement(name: &Ident, purpose: CodeGenPurpose) -> TokenStream2 {
+    match purpose {
+        CodeGenPurpose::Message => quote! {let #name = reader.read_from_cursor(&mut cursor)?;},
+        CodeGenPurpose::Fieldset => quote! {let #name = reader.read_from_cursor(cursor)?;},
+    }
```
acmorrow, line 201: I'm a little confused here, why does one form need a &mut and not the other?
Also don't love that this method basically assumes that `cursor` is a valid symbol in the context of the expansion, but maybe that's generaly going on and it is just more noticeable here.

micro-rdk-nmea-macros/src/composition.rs
```diff
@@ -141,12 +157,55 @@ impl PgnComposition {
...
+        let error_ident = quote! {#crate_ident::parse_helpers::errors::NmeaParseError};
+        let field_set_ident = quote! {#crate_ident::parse_helpers::parsers::FieldSet};
+
+        quote! {
+            impl #impl_generics #name #src_generics #src_where_clause {
+                #(#attribute_getters)*
+            }
+
+            impl #impl_generics #field_set_ident for #name #src_generics #src_where_clause {
+                fn from_data(cursor: &mut DataCursor) -> Result<Self, #error_ident> {
```
acmorrow, line 179: It bothers me a bit that `into_token_stream` ends up with ` from_bytes` over a Vec, but `into_fieldset_token_stream` ends up with a `from_data` over a `&mut DataCursor`. Why do they differ that way?

acmorrow (changes requested):
Overall, I'm fairly happy with this. I like the way the field set declarations work. A few questions. It might be worth sitting down together for 30 to talk through a few things, mostly for my understanding.

micro-rdk-nmea-macros/src/composition.rs
gvaradarajan: I guess they don't need to, I can change `from_bytes` to `from_cursor` and have it take a `DataCursor` argument instead

micro-rdk-nmea-macros/src/composition.rs
gvaradarajan: A fieldset is only going to be parsed in the context of an overall message so it's `from_cursor` needs to take a reference to the cursor rather than consuming it as the one for message would.

The assumption of cursor as a valid symbol is generally happening everywhere, but I think it's acceptable because the only exposed code is the proc_macro. Every other function is meant to only be used as part of the implementation of the macro

micro-rdk-nmea-macros/src/composition.rs
```diff
@@ -51,16 +57,23 @@ impl PgnComposition {
 
             let macro_attrs = MacroAttributes::from_field(field)?;
             if macro_attrs.offset != 0 {
-                let offset = macro_attrs.offset / 8;
+                let offset = macro_attrs.offset;
```
acmorrow, line 60: Was the `/ 8` a bug? Or did the API of `cursor.read` change? Meaning of `#offset`?

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -11,40 +11,42 @@ use super::{
...
+        DataCursor {
+            data,
+            bit_offset: 0,
+        }
     }
 
     pub fn read(&mut self, bit_size: usize) -> Result<Vec<u8>, NmeaParseError> {
-        if bit_size / 8 > self.data.len() {
+        let bits_to_read = bit_size + self.bit_offset;
+        if bits_to_read / 8 > self.data.len() {
```
acmorrow, line 29: I think this won't work, because if you had one byte left, but `bits_to_read` was 15, this check would pass even though you only had 8 bits of data left.

Maybe `if bits_to_read > self.data.len() * 8` is clearer?

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -11,40 +11,42 @@ use super::{
...
-        // For example, let's say the data is [179, 152], and we want to read a u16 from a bit size of 12
-        //
-        // [179, 152] is 39091 in u16, reading the first 12 bits and ignoring the last 4
-        // should yield [10110011 10011000] => [10110011 00001001] => 2483 (in Little-Endian)
-        if bit_size % 8 != 0 {
-            res.push(self.data[0] >> (8 - (bit_size % 8)));
-            let mask: u8 = 255 >> (bit_size % 8);
-            if let Some(first_byte) = self.data.get_mut(0) {
-                *first_byte |= mask;
+        res.extend(self.data.drain(..(bits_to_read / 8)));
```
acmorrow, line 33: OK, just confirming, that here we are intentionally "under" reading. If `bits_to_read` is 15, we actually only want to read one byte permanently, then manually push a munged value we compute out of the byte we didn't permanently read, so that we can set the cursors `bit_offset` and leave it pointing to the partially read byte. Do I have that right?

acmorrow (approved):
I think I get it. There is what looks one small bug which should be fixed before push, and a thing I want you to confirm back, but other than that it LGTM.

micro-rdk-nmea-macros/src/composition.rs
gvaradarajan: it was a bug

micro-rdk-nmea/src/parse_helpers/parsers.rs
gvaradarajan: that is correct

PR #397, opened 2025-01-28
add gcp auth, use updated sdk action

mattjperez:
added a the full json credentials. according to the google auth actions doc, will not work for gsutil, so using gcloud instead.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -201,14 +201,15 @@ jobs:
         allowUpdates: true
         name: ${{ github.ref_name }}
         body: ${{ steps.changelog.outputs.changelog }}
+    
+    - name: Authenticate GCP
+      uses: 'google-github-actions/auth@v2'
+      with:
+        credentials_json: '${{ secrets.GCP_CREDENTIALS_JSON }}'
 
     - name: Setup Google Cloud SDK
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -199,14 +199,15 @@ jobs:
...
-      with:
-        version: 'latest'
-        project_id: ${{ secrets.GCP_PROJECT_ID }}
-        service_account_key: ${{ secrets.GCP_SA_KEY }}
+      uses: google-github-actions/setup-gcloud@v2
 
     - name: Upload to GCP
       run: |
-        gsutil cp micro-rdk-server-esp32-ota.bin gs://${{ secrets.GCP_BUCKET_NAME }}/${{ github.ref_name }}/
+        gcloud storage cp micro-rdk-server-esp32-ota.bin gs://${{ secrets.GCP_BUCKET_NAME }}/${{ github.ref_name }}/
```
acmorrow, line 213: The RDK seems to use something else to do this, via `google-github-actions/upload-cloud-storage`: https://github.com/viamrobotics/rdk/blob/main/.github/workflows/staticbuild.yml#L127

mattjperez:
confirmed worked with new `workflow_dispatch` button

PR #401, opened 2025-01-29
[RSDK-9855] Repair project template cargo config and remove cruft

acmorrow:
I'll leave some comments to help clarify these changes.


.cargo/config.toml
```diff
@@ -1,24 +1,17 @@
 [target.xtensa-esp32-espidf]
 linker = "ldproxy"
 runner = "espflash --monitor"
-rustflags = ["-C", "default-linker-libraries", "-Clink-args=-Wl,-Map=./esp-build.map", "--cfg", "espidf_time32"]
-
+rustflags = ["--cfg", "espidf_time32"]
```
acmorrow, line 4: `"-C", "default-linker-libraries"` was never intended for application on xtensa targets, it was only to work around https://github.com/esp-rs/embuild/issues/16 on riscv, which we don't currently support. If and when we *do* support it, the upstream issue may be resolved and we won't need it.

Also, the esp-build.map thing doesn't seem to get produced, at least for me. I'm going to double check that in a few different environments before committing to removing it, but if I can't get the file to be emitted I don't think we should have the flag.

.cargo/config.toml
```diff
@@ -1,24 +1,17 @@
...
 runner = "espflash --monitor"
-rustflags = ["-C", "default-linker-libraries", "--cfg", "espidf_time32"]
+rustflags = ["--cfg", "espidf_time32"]
 
 [target.xtensa-esp32s3-espidf]
 linker = "ldproxy"
 runner = "espflash --monitor"
-rustflags = ["-C", "default-linker-libraries", "--cfg", "espidf_time32"]
-
-[target.riscv32imc-esp-espidf]
```
acmorrow, line 17: We don't support riscv right now, so it doesn't make sense to me to carry around the block for configuring it.

micro-rdk-server/.cargo/config.toml
```diff
@@ -1,20 +0,0 @@
-[target.xtensa-esp32-espidf]
```
acmorrow, line 1: I removed this file because as far as I know, cargo doesn't consult `.cargo/config.toml` files from subdirectories of a workspace.

micro-rdk/.cargo/config.toml
```diff
@@ -1,19 +0,0 @@
-[target.xtensa-esp32-espidf]
```
acmorrow, line 1: Same as above removal.

templates/project/.cargo/config.toml
```diff
@@ -3,30 +3,21 @@
 {% if mcu != "esp32" %}#{% endif %}target = "xtensa-esp32-espidf"
 {% if mcu != "esp32s2" %}#{% endif %}target = "xtensa-esp32s2-espidf"
 {% if mcu != "esp32s3" %}#{% endif %}target = "xtensa-esp32s3-espidf"
-{% if mcu != "esp32c3" %}#{% endif %}target = "riscv32imc-esp-espidf"
 
 [target.xtensa-esp32-espidf]
 linker = "ldproxy"
 runner = "espflash --monitor"
+rustflags = ["--cfg", "espidf_time32"]
```
acmorrow, line 10: This espidf_time32 thing was needed on all platforms as long as we are on ESP-IDF 4, but in this file it was only being applied for the riscv target. It should be applied everywhere.

templates/project/.cargo/config.toml
```diff
@@ -3,30 +3,21 @@
...
 [target.xtensa-esp32s2-espidf]
 linker = "ldproxy"
 runner = "espflash --monitor"
+rustflags = ["--cfg", "espidf_time32"]
 
 [target.xtensa-esp32s3-espidf]
 linker = "ldproxy"
 runner = "espflash --monitor"
-
-[target.riscv32imc-esp-espidf]
```
acmorrow, line 20: Removed because we don't support this target

templates/project/.cargo/config.toml
```diff
@@ -3,30 +3,21 @@
...
-# Future - necessary for the experimental "native build" of esp-idf-sys with ESP32C3
-# See also https://github.com/ivmarkov/embuild/issues/16 [build]
-
-rustflags = ["-C", "default-linker-libraries", "--cfg", "espidf_time32"]
+rustflags = ["--cfg", "espidf_time32"]
 
 [unstable]
-
 build-std = ["std", "panic_abort"]
-#build-std-features = ["panic_immediate_abort"] # Required for older ESP-IDF versions without a realpath implementation
```
acmorrow, line 32: Stale

.cargo/config.toml
mattjperez: doesn't get produced for me either

mattjperez (approved):
LGTM

PR #405, opened 2025-01-29
update CI, python version and deps

mattjperez:
CI python dependencies weren't frozen, additionally we were using python 3.10. 

need to test if this works as expected using manual deploy before merging


canary/requirements.txt
```diff
@@ -1,5 +1,15 @@
-python-dateutil
-viam-sdk
-pymongo
-numpy
-slack-sdk
+dnspython==2.7.0
```
acmorrow, line 1: I think the better way to do this, rather than explicitly freezing including all transitive deps, is to set just the packages we directly rely on with "compatible" version markers. Given what is in this `freeze` result, I think that'd look like:
```
numpy ~= 2.2
pymongo ~= 4.11
python-dateutil ~= 2.2
slack-sdk ~= 3.34
viam-sdk ~= 0.38
```

.github/workflows/canary_ota.yml
```diff
@@ -17,8 +17,8 @@ on:
 env:
   ESP32_CANARY_ROBOT: ${{ secrets.ESP32_CANARY_ROBOT }}
   ESP32_CANARY_ROBOT_PART_ID: ${{ secrets.ESP32_CANARY_ROBOT_PART_ID }}
-  ESP32_CANARY_API_KEY_2: ${{ secrets.ESP32_CANARY_API_KEY_2 }}
-  ESP32_CANARY_API_KEY_ID_2: ${{ secrets.ESP32_CANARY_API_KEY_ID_2 }}
+  ESP32_CANARY_API_KEY: ${{ secrets.ESP32_CANARY_API_KEY }}
```
acmorrow, line 20: If we no longer need `KEY_*_2`, please remember to remove them from the repository settings after this merges.

acmorrow (changes requested):

canary/requirements.txt
mattjperez: changed

acmorrow (approved):

PR #408, opened 2025-01-30
[CONSULT-469] - introduce macro for PGN registration and NmeaMessage type

gvaradarajan:


micro-rdk-nmea-macros/src/composition.rs
```diff
@@ -138,17 +129,22 @@ impl PgnComposition {
         let error_ident = quote! {#crate_ident::parse_helpers::errors::NmeaParseError};
         let mrdk_crate = crate::utils::get_micro_rdk_crate_ident();
         quote! {
+
             impl #impl_generics #name #src_generics #src_where_clause {
-                pub fn from_cursor(mut cursor: #crate_ident::parse_helpers::parsers::DataCursor, source_id: u8) -> Result<Self, #error_ident> {
+                #(#attribute_getters)*
```
acmorrow, line 134: This block is now *so close* to what is in `into_fieldset_token_stream`. Is there a way to unify it?

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -106,3 +115,96 @@ pub struct GnssSatsInView {
...
+            }
+        }
+    };
+}
+
+define_pgns!(
+    (WaterDepth, Pgn128267, 128267),
+    (TemperatureExtendedRange, Pgn130316, 130316),
+    (GnssSatsInView, Pgn129540, 129540)
+);
```
acmorrow, line 159: - I'm not keen on these being separate from the associated `struct`s. I feel like somehow the 128267 should be attached to the `struct WaterDepth` declaration. Ideally, there'd be a way, statically, to obtain the PGN for a given struct as well.

- Also not keen on needing to re-iterate the PGN number here, once as `PgnX` and then as `X`.

- Maybe this is asking for too much magic from the macro system, but could you somehow derive programatically all the applications of the `PgnMessageDerive` macro and then you wouldn't even need to re-enumerate them here?

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -106,3 +115,96 @@ pub struct GnssSatsInView {
...
+
+pub struct NmeaMessage {
+    metadata: NmeaMessageMetadata,
+    data: MessageData,
+}
+
+impl NmeaMessage {
+    pub fn new(mut bytes: Vec<u8>) -> Result<Self, NmeaParseError> {
+        let msg_data = bytes.split_off(32);
```
acmorrow, line 168: This `32` should be a constant somewhere, and replaced globally

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -106,3 +115,96 @@ pub struct GnssSatsInView {
...
+                        Self::$enum($pgndef::from_cursor(cursor)?)
+                    }),*,
+                    x => Self::Unsupported(UnparsedMessageData::from_bytes(bytes, x)?)
+                })
+            }
+
+            pub fn to_readings(self) -> Result<GenericReadingsResult, crate::parse_helpers::errors::NmeaParseError> {
+                match self {
+                    $(Self::$enum(msg) => msg.to_readings()),*,
+                    Self::Unsupported(msg) => msg.to_readings()
```
acmorrow, line 148: Is it meaningful to return an unsupported as a reading?

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -106,3 +115,96 @@ pub struct GnssSatsInView {
...
+                    $(Self::$enum(_) => $pgn),*,
+                    Self::Unsupported(unparsed) => unparsed.pgn()
+                }
+            }
+
+            pub fn from_bytes(pgn: u32, bytes: Vec<u8>) -> Result<Self, crate::parse_helpers::errors::NmeaParseError> {
+                Ok(match pgn {
+                    $($pgn => {
+                        let cursor = DataCursor::new(bytes);
+                        Self::$enum($pgndef::from_cursor(cursor)?)
```
acmorrow, line 139: If each struct as an associated PGN, is there a double check you can do here that the runtime value of `pgn` matches the one statically associated with the struct?

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -106,3 +115,96 @@ pub struct GnssSatsInView {
...
+    (GnssSatsInView, Pgn129540, 129540)
+);
+
+pub struct NmeaMessage {
+    metadata: NmeaMessageMetadata,
+    data: MessageData,
+}
+
+impl NmeaMessage {
+    pub fn new(mut bytes: Vec<u8>) -> Result<Self, NmeaParseError> {
```
acmorrow, line 167: I'm curious why this takes the `Vec` by value rather than just viewing the data, since ultimately, everything gets parsed out, shouldn't this be able to take something like a byte slice, all the way down?

Should this implement `TryInto` at the top level?

micro-rdk-nmea/src/lib.rs
```diff
@@ -26,11 +29,10 @@ mod tests {
         let mut data = Vec::<u8>::new();
         let res = general_purpose::STANDARD.decode_vec(water_depth_str, &mut data);
         assert!(res.is_ok());
-        let cursor = DataCursor::new(data[33..].to_vec());
-        let message = WaterDepth::from_cursor(cursor, 13);
+        let cursor = DataCursor::new(data[32..].to_vec());
```
acmorrow, line 32: So, before, what was happening by starting at byte 33?

Also, is there a reason in this test and the others not to interpret the metadata in the 32 bytes at the front? It seems worthwhile to test that it parses correctly.

Finally, it looks like `WaterDepth` still does have a `source_id`, so, I'm curious why the value is no longer being validated.

acmorrow (changes requested):
Lots of thoughts and questions on this one

micro-rdk-nmea/src/messages/pgns.rs
gvaradarajan: I tried to do all the things you're asking for here previously and couldn't find a reasonable way to bend the macro system to achieve them. Since this file will be auto-generated in the future (except for the declarative macro), I don't think it's a big deal. Auto-generated files are rarely elegant or nice to look at anyway

micro-rdk-nmea/src/messages/pgns.rs
gvaradarajan: Yeah I think so. If there's a PGN implementation missing this will give you data to test with immediately

micro-rdk-nmea-macros/src/composition.rs
gvaradarajan: The common statements in the meat of the functions are already pretty unified, I'm not sure what further abstraction is valuable?

micro-rdk-nmea/src/lib.rs
gvaradarajan: Byte 32 was the source_id for those messages that have one, but not every message has a source_id field. I can add checks in the tests for the ones that do

micro-rdk-nmea/src/messages/pgns.rs
gvaradarajan: No because the runtime value of the PGN is what determines which struct will be used (see `NmeaMessage::new`)

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -251,6 +256,7 @@ impl TryFrom<Vec<u8>> for NmeaMessageMetadata {
         if value.len() < 32 {
             return Err(NmeaParseError::NotEnoughData);
         }
+        let pgn = u32::from_le_bytes(value[0..4].try_into()?);
```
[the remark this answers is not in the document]
gvaradarajan: I couldn't find any explanation either in the canboat page or our gonmea library, I'm guessing they're "reserved" in much the way a sequence of bytes often is in the middle of the data for an actual message

micro-rdk-nmea/src/messages/pgns.rs
acmorrow: I do see what you are saying about the virality of lifetime when dealing with byte slices, but, on the other hand, `Vec` is saying that this method needs a uniquely owned buffer of data in order to convert it into an `NmeaMessage`. That seems like it shouldn't be needed. If I have a bunch of bytes that I read into a buffer off the wire or bus that is, say, already in a Vec, I'd expect the parser to just be able to consume that data by read-only reference, rather than needing to copy it into a Vec just to call this method. On the other hand, I don't want to delay this work further. Would you be OK with filing a follow-up ticket?

micro-rdk-nmea/src/messages/pgns.rs
acmorrow: I meant it as more sort of a debug assertion. Once you have the type, the statically associated ID on the type better be equal to the runtime type or somethign went terribly wrong. But I'm mostly OK with your argument that that can't happen.

micro-rdk-nmea/src/lib.rs
acmorrow: I think it is worth checking the source_id is decoded correctly for those types that have them.

I also think it is worth parsing the metadata here and validating the decoded values.

micro-rdk-nmea-macros/src/composition.rs
acmorrow: Nevermind, I took a closer look and there is more variation than I'd realized between the two.

acmorrow (changes requested):

micro-rdk-nmea/src/parse_helpers/parsers.rs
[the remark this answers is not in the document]
acmorrow: I guessed it was somethign like that. I'd have a slight preference for this method to use something cursor-ier to walk the `value` bytes, rather than just hardcoding the ranges with a skip, but I'm not going to push for it.

micro-rdk-nmea/src/messages/pgns.rs
gvaradarajan: I don't understand how you would use these traits in the defined enum. It's not like you could iterate through the possible enum variants to look for which variant type has the matching PGN right?

micro-rdk-nmea/src/lib.rs
acmorrow: Thanks for adding the check of the `source_id`. Did you disagree with the idea to validate the metadata too?

micro-rdk-nmea/src/messages/pgns.rs
acmorrow: Ping?

micro-rdk-nmea/src/messages/pgns.rs
acmorrow: It isn't for using in the defined enum. It is to create a constant association between the struct defining the PGN, and the PGN number.

Lets say i have a `Vec<u8>`, and for some reason I'm explicitly trying to decode it to a `WaterDepth` and associated metadata. If the PGN code is associated with `WaterDepth`, then that decoding logic can validate that the PGN byte it sees matches the constant associated with the type, and if it doesn't match it can throw an error.

Or, imagine if later we wanted to actually serialize back to on-the-wire bytes for some reason. So, given a `WaterDepth` struct, how do I know what PGN number to write in the metadata? I don't, because it is locked up in the `MessageData` enum machinery / `define_pgns!`. But if each type has an associated constant, then I don't have that issue. I can know exactly what PGN code to write.

micro-rdk-nmea/src/lib.rs
acmorrow: I don't feel too strongly about it. I was suggesting it mostly because you already have static test data here, but the first 32 bytes get thrown away, when it could be a perfectly good test of decoding the metadata. If you do a separate test then you would just need to obtain the metadata bytes from somewhere anyway. Up to you.

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -106,3 +115,96 @@ pub struct GnssSatsInView {
...
+                    $(Self::$enum(msg) => msg.to_readings()),*,
+                    Self::Unsupported(msg) => msg.to_readings()
+                }
+            }
+        }
+    };
+}
+
+define_pgns!(
+    (WaterDepth, Pgn128267, 128267),
```
acmorrow, line 156: So, if I understand correctly, the generated enum will look like:
```
enum MessageData {
    Pgn128267(WaterDepth),
    Pgn130316(TemperatureExtendedRange),
    ...
}
```

So, do you actually need the second item in the tuple? 

```
define_pgns!(
    (WaterDepth, 127267),
    (TemperatureExtendedRange, 130316),
    ...
```
would seem to have all of the information required to generate the enum, and you wouldn't need to write the PGN value twice, once as part of a type name and the other as an integer.

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -106,3 +115,96 @@ pub struct GnssSatsInView {
     #[length_field = "sats_in_view"]
     satellites: Vec<Satellite>,
 }
+
+macro_rules! define_pgns {
+    ( $(($pgndef:ident, $enum:ident, $pgn:expr)),* ) => {
+        #[derive(Clone, Debug)]
+        pub enum MessageData {
```
acmorrow, line 122: Maybe `NmeaMessage` instead of `MessageData`, which is pretty generic.

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -106,3 +115,96 @@ pub struct GnssSatsInView {
...
+                }
+            }
+        }
+    };
+}
+
+define_pgns!(
+    (WaterDepth, Pgn128267, 128267),
+    (TemperatureExtendedRange, Pgn130316, 130316),
+    (GnssSatsInView, Pgn129540, 129540)
```
acmorrow, line 158: With the enum variants looking like `Pgn130316` or `Pgn129540`, this enum is somewhat illegible to a user, since they need to know the PGN code to match, rather than the human readable name like `GnssSatsInView`. Is it necessary to have the PGN number in the enum variants? Or could they be named much like the wrapped value?

micro-rdk-nmea/src/messages/pgns.rs
acmorrow: Ah, you have `NmeaMessage` later, which is this + metadata. Maybe this should be `NmeaMessageBody` or something.

micro-rdk-nmea/src/lib.rs
gvaradarajan: wait I do have a unit test for this actually, does this really matter for every message type?

acmorrow: No. Maybe just do one of them, if you want to.

micro-rdk-nmea/src/messages/pgns.rs
gvaradarajan: I previously had `DataCursor` work this way, but then @npmenard said it should own the data. If you can convince him this is better when he gets back, I'll make the ticket.

My opinion on this is not strong but I don't really see a use-case outside of debugging to want a read-only reference to the data meant to be converted. I can see a use case for maybe supporting any type that implements Read. I could make a ticket for that if you agree

micro-rdk-nmea/src/messages/pgns.rs
gvaradarajan: it's hard to interpolate tokens, I can try to see if the paste crate helps out here

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -1,12 +1,23 @@
...
-    parsers::{DataCursor, FieldSet},
+    errors::NmeaParseError,
+    parsers::{DataCursor, FieldSet, NmeaMessageMetadata},
 };
 
-#[derive(PgnMessageDerive, Debug)]
+#[derive(PgnMessageDerive, Clone, Debug)]
 pub struct WaterDepth {
+    #[pgn = 128267]
+    _pgn: PhantomData<u32>,
```
acmorrow, line 20: You need the PhantomData so there is something for the `pgn` macro to act on, is that right?

I'm pretty OK with it given that these are ultimately going to be generated. But, if you wanted to eliminate it, maybe it would work to just let the #pgn macro adhere to whatever was the first field and then dig it out as you iterate the fields?

micro-rdk-nmea-macros/src/attributes.rs
```diff
@@ -144,6 +146,32 @@ impl MacroAttributes {
...
+                                    } else {
+                                        return Err(error_tokens("pgn parameter must be int"));
+                                    }
+                                } else {
+                                    return Err(error_tokens("pgn parameter must be int"));
+                                }
+                            }
+                            _ => {
+                                return Err(error_tokens(
+                                    "bits received unexpected attribute value",
```
acmorrow, line 170: pgn instead of bits, I think.

micro-rdk-nmea/src/messages/pgns.rs
acmorrow: Sure, that sounds promising. But see my other comment just below about the naming of these enum members anyway. I'm not convinced that names with the PGN code in them are what is desired

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -106,3 +124,101 @@ pub struct GnssSatsInView {
...
+    metadata: NmeaMessageMetadata,
+    data: MessageData,
+}
+
+impl TryFrom<Vec<u8>> for NmeaMessage {
+    type Error = NmeaParseError;
+    fn try_from(mut value: Vec<u8>) -> Result<Self, Self::Error> {
+        let msg_data = value.split_off(MESSAGE_HEADER_OFFSET);
+        let metadata = NmeaMessageMetadata::try_from(value)?;
+        let data = MessageData::from_bytes(metadata.pgn(), msg_data)?;
```
acmorrow, line 182: I feel like both these types (`NmeaMessageMetadata` and `MessageData`) should either start with `Nmea`, or not. Given that there is also NmeaMessage, maybe that breaks the tie in favor of `NmeaMessageData`?

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -106,3 +124,101 @@ pub struct GnssSatsInView {
     #[length_field = "sats_in_view"]
     satellites: Vec<Satellite>,
 }
+
+macro_rules! define_pgns {
+    ( $(($pgndef:ident, $enum:ident, $pgn:expr)),* ) => {
```
acmorrow, line 129: Do you still need $pgn?

micro-rdk-nmea-macros/src/attributes.rs
```diff
@@ -46,6 +46,7 @@ pub(crate) struct MacroAttributes {
     pub(crate) is_lookup: bool,
     pub(crate) is_fieldset: bool,
     pub(crate) length_field: Option<Ident>,
+    pub(crate) pgn: Option<u32>,
```
acmorrow, line 49: It isn't really optional though, is it? Is there a check somewhere that ensures that this gets populated into `MacroAttributes`? I mean, I see why you did it this way, so you can make `MacroAttributes` and then populate it step-by-step, so I don't mind that much. I guess you could work around it with a builder that had an `Option` and then a built type that didn't. But that feels like overkill to just remove on `Option`

micro-rdk-nmea-macros/src/composition.rs
```diff
@@ -31,31 +32,49 @@ impl PgnComposition {
...
+                    statements.set_pgn_declaration(quote! {
+                        const PGN: u32 = #pgn;
+                    });
+                    statements
+                        .struct_initialization
+                        .push(quote! { _pgn: std::marker::PhantomData, });
+                    return Ok(statements);
+                } else {
+                    let err_msg = format!(
+                        "pgn field must define pgn attribute, macro attributes: {:?}",
```
acmorrow, line 71: Ah, this is where the `Option` being non-`None` gets enforced.

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -106,3 +124,101 @@ pub struct GnssSatsInView {
...
+                match self {
+                    $(Self::$enum(msg) => msg.to_readings()),*,
+                    Self::Unsupported(msg) => msg.to_readings()
+                }
+            }
+        }
+    };
+}
+
+pub const MESSAGE_HEADER_OFFSET: usize = 32;
```
acmorrow, line 164: Header, or MetaData? Or Data? I think it is the offset of the data?

acmorrow (changes requested):
I think this is very close to ready to approve. This pass is mostly only small things or questions/suggestions, and some naming nits.

micro-rdk-nmea/src/messages/pgns.rs
gvaradarajan: Yes that's right. I considered using whatever the first field is, but ultimately decided against it for the sake of clarity

micro-rdk-nmea-macros/src/attributes.rs
gvaradarajan: An instance of `MacroAttributes` pertains to a particular field, so it should, in fact, be optional

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -145,41 +145,41 @@ macro_rules! define_pgns {
...
 }
 
-pub const MESSAGE_HEADER_OFFSET: usize = 32;
+pub const MESSAGE_DATA_OFFSET: usize = 32;
 
 define_pgns!(
-    (WaterDepth, Pgn128267, 128267),
-    (TemperatureExtendedRange, Pgn130316, 130316),
-    (GnssSatsInView, Pgn129540, 129540)
+    (WaterDepth, 128267),
```
acmorrow, line 167: Last question, I promise: Do you even need the PGN code as the second argument anymore?

acmorrow (approved):
LGTM with one last annoying question.

micro-rdk-nmea/src/messages/pgns.rs
gvaradarajan: yes I cannot use information only available at runtime to generate the content of the match statement for the `from_bytes` function

PR #411, opened 2025-02-04
PGN testing

acmorrow:
- Run all tests in the workspace and do not fail fast
- Avoid need to reiterate PGN numbers in the PGN enum macro.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ upload: cargo-ver
 	cargo +esp espflash flash --package micro-rdk-server --monitor --partition-table micro-rdk-server/esp32/partitions.csv --baud 460800 -f 80mhz --bin micro-rdk-server-esp32 --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort
 
 test:
-	cargo test -p micro-rdk --lib --features native,ota
+	cargo test --workspace --tests --no-fail-fast --features native,ota
 
 clippy-native:
 	cargo clippy -p micro-rdk --no-deps --features native,ota  -- -Dwarnings
--- a/micro-rdk-nmea/src/lib.rs
+++ b/micro-rdk-nmea/src/lib.rs
@@ -9,8 +9,7 @@ mod tests {
         messages::{
             message::Message,
             pgns::{
-                GnssPositionData, GnssSatsInView, TemperatureExtendedRange, WaterDepth,
-                MESSAGE_DATA_OFFSET,
+                GnssPositionData, GnssSatsInView, NmeaMessage, NmeaMessageBody, TemperatureExtendedRange, WaterDepth, MESSAGE_DATA_OFFSET
             },
         },
         parse_helpers::{
@@ -157,10 +156,17 @@ mod tests {
         let res = general_purpose::STANDARD.decode_vec(msg_str, &mut data);
         assert!(res.is_ok());
 
-        let cursor = DataCursor::new(data[MESSAGE_DATA_OFFSET..].to_vec());
-        let message = GnssSatsInView::from_cursor(cursor);
-        assert!(message.is_ok());
+        let nmea_message = NmeaMessage::try_from(data);
+        assert!(nmea_message.is_ok());
+
+        let nmea_message = nmea_message.unwrap();
+        let message = match nmea_message.data {
+            NmeaMessageBody::GnssSatsInView(val) => Some(val),
+            _ => None,
+        };
+        assert!(message.is_some());
         let message = message.unwrap();
+
         println!("message: {:?}", message);
 
         let source_id = message.source_id();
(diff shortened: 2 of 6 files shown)


micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -126,7 +126,7 @@ pub struct GnssSatsInView {
 }
 
 macro_rules! define_pgns {
-    ( $(($pgndef:ident, $pgn:expr)),* ) => {
+    ( $($pgndef:ident),* ) => {
```
acmorrow, line 129: I didn't push more for this change in #408 because I wasn't completely sure there wasn't a reason it didn't work, but it sure seems to.

gvaradarajan: huh, ok. Sounds good to me

Makefile
```diff
@@ -68,7 +68,7 @@ upload: cargo-ver
 	cargo +esp espflash flash --package micro-rdk-server --monitor --partition-table micro-rdk-server/esp32/partitions.csv --baud 460800 -f 80mhz --bin micro-rdk-server-esp32 --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort
 
 test:
-	cargo test -p micro-rdk --lib --features native,ota
+	cargo test --workspace --tests --no-fail-fast --features native,ota
```
gvaradarajan, line 71: what is `--no-fail-fast`?

gvaradarajan (approved):
LGTM

micro-rdk-nmea/src/messages/pgns.rs
acmorrow: I guess the one thing to check is whether this `NmeaMessageBody` type actually has any tests? If you want, I'll add one.

Makefile
acmorrow: It prevents a failing test from stopping the execution of other tests. I'd much prefer to learn in one go all the tests I broke, rather than it stopping on the first failure.

micro-rdk-nmea/src/lib.rs
```diff
@@ -157,10 +156,17 @@ mod tests {
...
+        assert!(nmea_message.is_ok());
+
+        let nmea_message = nmea_message.unwrap();
+        let message = match nmea_message.data {
+            NmeaMessageBody::GnssSatsInView(val) => Some(val),
+            _ => None,
+        };
+        assert!(message.is_some());
         let message = message.unwrap();
+
```
acmorrow, line 169: @gvaradarajan - New since last review, this now ensures that the NmeaMessage enum gets tested at least once.

micro-rdk-nmea/src/messages/pgns.rs
```diff
@@ -164,14 +164,14 @@ macro_rules! define_pgns {
...
-    (GnssSatsInView, 129540)
+    WaterDepth,
+    TemperatureExtendedRange,
+    GnssSatsInView
 );
 
 pub struct NmeaMessage {
-    metadata: NmeaMessageMetadata,
-    data: NmeaMessageBody,
+    pub metadata: NmeaMessageMetadata,
```
acmorrow, line 173: Needed for testing, let me know if you would prefer accessors, but I find them less interesting in a language where immutability is the default.

acmorrow (commented):
Updated, PTAL. I also fixed some new warnings that popped up with the expanded testing.

micro-rdk-nmea/src/lib.rs
gvaradarajan: thank you for adding this

micro-rdk-nmea/src/messages/pgns.rs
gvaradarajan: if it's just for testing, would `pub(crate)` work?

micro-rdk/src/common/robot.rs
```diff
@@ -504,6 +504,7 @@ impl LocalRobot {
     }
 
     pub fn get_periodic_app_client_tasks(&mut self) -> Vec<Box<dyn PeriodicAppClientTask>> {
+        #[allow(unused_mut)]
```
gvaradarajan, line 507: ah yes, finally

gvaradarajan (approved):
LGTM

micro-rdk-nmea/src/messages/pgns.rs
acmorrow: Ah, yes, I think it would. I always forget that that is an option. Too accustomed to C++ having only public/private. If it works I'll make that change before push.

acmorrow:
@gvaradarajan - Oops. I thought I'd committed and pushed the change to `pub(crate)` but it turns out I'd only staged it. However, having thought about it more, `NmeaMessage` is `pub`, but none of its fields are, which means the only thing you can do with it is create it, or serialize it. That feels pretty restrictive. Anyway, happy to make a follow-up commit if you disagree.

PR #416, opened 2025-02-18
[RSDK-10022] - affix target OS for installer build in CI to specific version

gvaradarajan:


.github/workflows/publish.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         job:
-          - { target: x86_64-unknown-linux-gnu, file_suffix: amd64-linux, os: ubuntu-latest}
+          - { target: x86_64-unknown-linux-gnu, file_suffix: amd64-linux, os: ubuntu-22.04}
```
acmorrow, line 18: - Do you want to pin the macos and windows builds too?
- Are we *actually* building x86_64-apple-darwin? Or is it really arm

gvaradarajan: maybe? what do you propose pinning them to?

And yes it is actually arm-apple-darwin

acmorrow: I think, per https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories, that they should be `windows-2022` and `macos-14` respectively. And we should probably fix up the target name for `x86_64-apple-darwin` since it is inaccurate. I think it is `aarch64-apple-darwin` but please check me on that.

gvaradarajan: no wait, that target is correct, it's the Cargo triplet

mattjperez:
I pushed a copy of this branch to the repo to test the changes so far. [job link](https://github.com/viamrobotics/micro-rdk/actions/runs/13416256436)

.github/workflows/publish.yml
acmorrow: Yes, almost all recent macs are arm based, and the macos-14 builder is too, and `macos-latest` is `macos-14` so these were already arm builds.

acmorrow: Yes, and Rosetta was helpfully running the x86 binary for me.

acmorrow (approved):
LGTM

PR #419, opened 2025-02-24
[RSDK-10067] - add release action workflow to project template

gvaradarajan:
This will significantly reduce the friction for those developing firmware using the project template that want to make use of OTA by preparing a compiled binary as a release asset whenever a new version tag is pushed to the project. In theory, the download link to the release asset can be used in the configuration of the OTA service for a machine hosting the firmware.

However, this download link will only work when we have fixed the OTA service to properly handle redirects from the link location of the binary (likely to be addressed by [RSDK-10054](https://viam.atlassian.net/browse/RSDK-10054)).

[RSDK-10054]: https://viam.atlassian.net/browse/RSDK-10054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/templates/project/.github/workflows/publish.yml
+++ b/templates/project/.github/workflows/publish.yml
@@ -1,0 +1,60 @@
+name: publish
+
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  build-project:
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/viamrobotics/micro-rdk-dev-env:amd64
+    steps:
+    - name : Checkout main branch code
+      uses: actions/checkout@v3
+      with:
+        fetch-depth: 2
+    - name: Build esp32 OTA binary
+      run: |
+        bash -c '. /home/testbot/.bash_profile ; git config --global --add safe.directory "$ESP_ROOT"/esp-idf && make build-esp32-ota'
+        cp target/xtensa-esp32-espidf/project-ota.bin project-ota.bin
+    - name: Upload release Lib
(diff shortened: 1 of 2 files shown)


gvaradarajan:
@JessamyT I've marked that there should be doc changes in the ticket, but also wanted to let you know about this here. On the version release after this is merged, we'll have to inform users of the project template that they'll have to give "Read and write permissions" in the Workflow Settings of their repository to properly use this.

templates/project/.github/workflows/publish.yml
```diff
@@ -0,0 +1,60 @@
...
+          project-ota.bin
+
+  publish-release:
+    needs: [build-project]
+    runs-on: ubuntu-latest
+    steps:
+    - name: Check release type
+      id: check-tag
+      run: |
+        if echo ${{ github.event.ref }} | grep -Eq '^refs/tags/v.*rc[0-9]{1}$'; then
```
mattjperez, line 37: should we add a project README where we can add a line about this GH actions workflow?

mattjperez (approved):
one question, but otherwise LGTM

templates/project/.github/workflows/publish.yml
gvaradarajan: I think not, we don't want that README to end up in the generated template because owners of the projects will probably want their own READMEs. If anything, details about how to use the project template should be in DEVELOPMENT.md. But I don't want to include it until the github links are usable in OTA and we can successfully document a complete release process.

PR #421, opened 2025-02-24
[RSDK-9887] try multiple networks when connecting

mattjperez:
This PR creeped a little, but it covers the following:
1. preps `WifiCredentials[Storage]` for change to `NetworkSetting[sStorage]`, but maintaining compatibility with installer and provisioning, and current project template. Final transition and updating templates left for later work. [RSDK-10105]
2. add logic for trying multiple network credentials (with priority [RSDK-9889])
3. moves the temporary logic for checking agent config to the config monitor [RSDK-9888]

[RSDK-9889]: https://viam.atlassian.net/browse/RSDK-9889?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RSDK-9888]: https://viam.atlassian.net/browse/RSDK-9888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[RSDK-10105]: https://viam.atlassian.net/browse/RSDK-10105?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


micro-rdk/src/common/config_monitor.rs
```diff
@@ -116,13 +117,35 @@ where
...
+                            reboot = true;
+                        }
+                    }
                 }
             }
 
+            if reboot {
+                log::info!("rebooting from config monitor...");
+                // TODO(RSDK-9464): flush logs to app.viam before restarting
+                self.restart();
```
acmorrow, line 146: I think the reason we reboot for a robot configuration change is because we don't implement reconfiguration, so starting over is the only way to get a working robot again. But I don't think the same is necessarily true for wifi settings. On the other hand, implementing it without a reboot seems sort of complicated, since the updated settings might invalidate a network that is currently in use.

micro-rdk/src/common/config_monitor.rs
```diff
@@ -116,13 +117,35 @@ where
                         reboot = true;
                     }
                 }
+            }
+
+            if let Ok(device_agent_config) = app_client.get_agent_config().await {
```
acmorrow, line 122: Should we be doing this at all in the case of an externally managed network?

micro-rdk/src/common/conn/viam.rs
```diff
@@ -489,14 +489,21 @@ where
...
+            log::debug!("networks to try: {:?}", networks);
+            for network in networks.iter().cycle() {
+                log::info!("attempting to connect to network `{}`", network.ssid);
+                if let Err(err) = wifi.set_sta_mode(network.clone()).await {
+                    log::error!("failed to connect to network: {:?}", err);
+                    let _ = Timer::after(Duration::from_secs(1)).await;
+                } else {
+                    log::info!("successfully connected to network `{}`", network.ssid);
+                    break;
+                }
```
acmorrow, line 506: If we have a thing called `WifiManager`, I'd expect to see this sort of logic encapsulated there, rather than explicitly written out here.

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -70,6 +58,21 @@ impl From<RobotCredentials> for CloudConfig {
     }
 }
 
+#[derive(Clone, Debug, Default)]
```
acmorrow, line 61: Was there a particular reason to move this?

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -88,32 +91,22 @@ impl WifiCredentials {
...
     fn has_wifi_credentials(&self) -> bool;
     fn store_wifi_credentials(&self, creds: &WifiCredentials) -> Result<(), Self::Error>;
-    fn get_wifi_credentials(&self) -> Result<WifiCredentials, Self::Error>;
-    fn reset_wifi_credentials(&self) -> Result<(), Self::Error>;
+
+    fn has_default_network(&self) -> bool;
+    fn store_default_network(&self, ssid: &str, password: &str) -> Result<(), Self::Error>;
+    fn get_default_network(&self) -> Result<NetworkSetting, Self::Error>;
+    fn reset_default_network(&self) -> Result<(), Self::Error>;
+    fn get_all_networks(&self) -> Result<Vec<NetworkSetting>, Self::Error>;
```
acmorrow, line 109: - I get that the old name isn't ideal anymore, but is this renaming more trouble than it is worth right now? Even if you fix the template, it will mean old projects can't upgrade to the next build of micro-rdk without changes.

- If you feel strongly that we should do the renaming, I think you need forwarding functions for all four methods, not just two.

- The forwarding methods should probably be annotated with `deprecated`: https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-deprecated-attribute

micro-rdk/src/common/credentials_storage.rs
```diff
@@ -88,32 +91,22 @@ impl WifiCredentials {
...
-    fn from(resp: CertificateResponse) -> Self {
-        Self {
-            certificate: resp.tls_certificate.into_bytes(),
-            private_key: resp.tls_private_key.into_bytes(),
-        }
-    }
-}
-
-pub trait NetworkSettingsStorage {
+pub trait WifiCredentialStorage {
```
acmorrow, line 94: This is restoring the old name, correct?

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -335,6 +338,10 @@ impl NetworkSettingsStorage for NVSStorage {
         network_settings: &[NetworkSetting],
     ) -> Result<(), Self::Error> {
         let bytes: Vec<u8> = postcard::to_allocvec(network_settings)?;
+        log::info!(
```
acmorrow, line 341: Write deduplication effectiveness depends on deterministic values here. Should the values in `network_settings` be sorted somehow before invoking `postcard`?

acmorrow (changes requested):

micro-rdk/src/common/config_monitor.rs
mattjperez: yeah I wasn't sure about this part either. My reasoning for this was what if we get a network that has priority over the current connection and it doesn't reboot? For example, if a device is currently using a metered/costly network and we update to prioritize another one which isn't. Unless the metered network goes down or the device resets, the device won't switch to the non-metered network.

micro-rdk/src/common/credentials_storage.rs
mattjperez: correct

micro-rdk/src/esp32/nvs_storage.rs
mattjperez: I can have it sort it before writing. will enforce it here as opposed to needing the caller to pre-sort it.

micro-rdk/src/common/credentials_storage.rs
mattjperez: I had removed the other two as they weren't actively used anywhere, but you're right that they were still available and someone could have been using it in their own project. I will write forwarding functions as well. 

I like the idea of adding the deprecated tag since that will ensure both compatibility and make users aware that they should change it.

micro-rdk/src/common/conn/viam.rs
acmorrow: Another reason to like WifiManager for this is that it has (I think) the ability to scan for visible wifi networks. That'd let you filter out any in the list that aren't even visible, making for a faster connection experience overall.

micro-rdk/src/common/credentials_storage.rs
mattjperez: to group `WifiCredentials` and `WifiCredentialsStorage` closer together. though looking at everything it might be better to move all wifi credentials stuff to the beneath `RobotConfigurationStorage`. can do it before merging.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -489,14 +489,12 @@ where
...
         if let Some(wifi) = self.wifi_manager.as_ref().as_ref() {
-            while let Err(err) = wifi
-                .set_sta_mode(self.storage.get_wifi_credentials().unwrap())
-                .await
-            {
-                log::error!("couldn't connect to wifi reason {:?}", err);
-                let _ = Timer::after(Duration::from_secs(2)).await;
+            let networks = self.storage.get_all_networks().unwrap();
+            log::info!("attempting to configure wifi according to priority...");
+            while let Err(_) = wifi.try_connect_with_priority(networks.clone()).await {
```
mattjperez, line 497: I'm not a fan of this `clone()`. We could remove it, however I'm not sure how we can ensure the inputs are sorted otherwise. same issue exists with `store_network_settings`.

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -299,6 +297,36 @@ impl Esp32WifiNetwork {
...
+    async fn try_connect_with_priority(
+        &self,
+        mut networks: Vec<NetworkSetting>,
+    ) -> Result<(), WifiManagerError> {
+        let access_points: Vec<String> = self
+            .scan_networks_inner()
+            .await
+            .inspect_err(|e| log::error!("failed to scan available networks: {}", e))?
+            .iter()
+            .map(|ap| ap.ssid.to_string())
```
mattjperez, line 309: could probably add more filtering on some minimum signal strength

npmenard (changes requested):
done with first path

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -343,29 +353,47 @@ impl NetworkSettingsStorage for NVSStorage {
...
-        let ssid = self.get_string(NVS_WIFI_SSID_KEY)?;
-        let pwd = self.get_string(NVS_WIFI_PASSWORD_KEY)?;
-        Ok(WifiCredentials { ssid, pwd })
+    fn get_default_network(&self) -> Result<NetworkSetting, Self::Error> {
+        let ssid = self.get_string(NVS_DEFAULT_SSID_KEY)?;
+        let password = self.get_string(NVS_DEFAULT_PASSWORD_KEY)?;
+        Ok(NetworkSetting {
+            ssid,
+            password,
+            priority: 0,
```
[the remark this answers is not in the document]
mattjperez: Since priorities can be negative, we chose 0 as the default. This way users can have control over the order of networks tried relative to the flashed network.

Higher values have higher priority.

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -343,29 +353,47 @@ impl NetworkSettingsStorage for NVSStorage {
...
+            Vec::new()
+        };
+        networks.push(default);
+        Ok(networks)
     }
 
-    fn store_wifi_credentials(&self, creds: &WifiCredentials) -> Result<(), Self::Error> {
-        self.set_string(NVS_WIFI_SSID_KEY, &creds.ssid)?;
-        self.set_string(NVS_WIFI_PASSWORD_KEY, &creds.pwd)
+    fn store_default_network(&self, ssid: &str, password: &str) -> Result<(), Self::Error> {
```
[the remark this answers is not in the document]
mattjperez: agreed. we would need to make similar changes in the installer for writing network credentials to NVS.

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -350,6 +378,13 @@ impl WifiManager for Esp32WifiNetwork {
     {
         Box::pin(async { self.set_ap_sta_mode(conifg_ap).await })
     }
+    fn try_connect_with_priority(
```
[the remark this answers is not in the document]
mattjperez: `try_connect_by_priority`? open to suggestions

micro-rdk/src/esp32/nvs_storage.rs
acmorrow: I don't think it should sort by priority though. I think network name would be a better sort key.

micro-rdk/src/common/config_monitor.rs
acmorrow: I think that's a reasonable take. If/when we implement network roaming we will need to revisit this anyway.

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -299,6 +297,36 @@ impl Esp32WifiNetwork {
         log::info!("connection successful");
         Ok(())
     }
+    async fn try_connect_with_priority(
+        &self,
+        mut networks: Vec<NetworkSetting>,
```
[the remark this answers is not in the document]
acmorrow: I'm in favor of sorting here. Otherwise there is a caller requirement that we don't have a way to enforce.

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -299,6 +297,36 @@ impl Esp32WifiNetwork {
...
+    ) -> Result<(), WifiManagerError> {
+        let access_points: Vec<String> = self
+            .scan_networks_inner()
+            .await
+            .inspect_err(|e| log::error!("failed to scan available networks: {}", e))?
+            .iter()
+            .map(|ap| ap.ssid.to_string())
+            .collect();
+        let access_points: HashSet<String> = HashSet::from_iter(access_points);
+        networks.sort_by(|a, b| b.priority.cmp(&a.priority));
```
acmorrow, line 314: Are there allowed to be multiple networks with the same priority? If so, how should they sort? Should there be a secondary sort on the network name?

mattjperez: added `impl Ord for NetworkSetting` that sorts by priority first, then name

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -343,29 +353,47 @@ impl NetworkSettingsStorage for NVSStorage {
...
+        let ssid = self.get_string(NVS_DEFAULT_SSID_KEY)?;
+        let password = self.get_string(NVS_DEFAULT_PASSWORD_KEY)?;
+        Ok(NetworkSetting {
+            ssid,
+            password,
+            priority: 0,
+        })
+    }
+
+    fn get_all_networks(&self) -> Result<Vec<NetworkSetting>, Self::Error> {
```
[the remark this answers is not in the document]
mattjperez: changed it to still return the default network even if the other networks are unavailable (such as error when reading/deserializing from NVS).

It will still error though if the default network is unable to be retrieved

npmenard (changes requested):
LGTM but there is a comment to be addressed before it can be merged

micro-rdk/src/esp32/nvs_storage.rs
```diff
@@ -334,7 +337,14 @@ impl NetworkSettingsStorage for NVSStorage {
         &self,
         network_settings: &[NetworkSetting],
     ) -> Result<(), Self::Error> {
-        let bytes: Vec<u8> = postcard::to_allocvec(network_settings)?;
+        let mut settings = network_settings.to_vec();
+        // enforce sorting by ssid before serialization for write deduplication
```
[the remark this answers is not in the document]
mattjperez: removed the sorting

npmenard (approved):
LGTM

PR #422, opened 2025-02-28
RSDK-9715 RSDK-9816 logging and app client safety

acmorrow:
I had plans to go further with more logging, but I stopped here because while working on the logging I got curious about the `AppClientTaskRunner` infrastructure, and I think I've found a way to get rid of the `unsafe` stuff. Expect follow-ups PRs under both tickets.

Some of the logging changes are opinionated. Some may be too chatty. Feel free to push back where you see fit.

As usual, I'll take a pass through and leave comments on anything I think needs further description, but I expect most of it should be straightforward.


micro-rdk-server/esp32/main.rs
```diff
@@ -70,30 +72,38 @@ mod esp32 {
...
                         PASS.unwrap().to_string(),
                     ))
-                    .expect("Failed to store WiFi credentials to NVS");
+                    .expect("failed to store WiFi credentials to storage");
             }
         }
 
-        if !storage.has_robot_configuration() {
+        if !storage.has_robot_credentials() {
+            log::warn!("no machine credentials were found in storage");
```
acmorrow, line 100: I've tried to consistently replace `robot` with `machine` in the logs I've updated.

micro-rdk-server/esp32/main.rs
```diff
@@ -70,30 +72,38 @@ mod esp32 {
...
         // At runtime, if the program does not detect credentials or configs in storage,
         // it will try to load statically compiled values.
 
         if !storage.has_wifi_credentials() {
+            log::warn!("no wifi credentials were found in storage");
+
             // check if any were statically compiled
             if SSID.is_some() && PASS.is_some() {
-                log::info!("Storing static values from build time wifi configuration to NVS");
+                log::info!("storing static values from build time wifi configuration to storage");
```
acmorrow, line 89: I've tried to consistently move to lowercase initial word in log messages I added or were nearby my changes.

micro-rdk-server/native/main.rs
```diff
@@ -41,10 +43,14 @@ mod native {
         // At runtime, if the program does not detect credentials or configs in storage,
         // it will try to load statically compiled values.
 
-        if !storage.has_robot_configuration() {
+        if !storage.has_robot_credentials() {
```
acmorrow, line 46: I believe that this was a bug. The other two mainlines here check `credentials`, but `native` was checking `configuration`.

micro-rdk/src/common/app_client.rs
```diff
@@ -316,6 +317,12 @@ impl AppClient {
                             current_dt,
                             err
                         );
+                    })
+                    .inspect(|_| {
```
acmorrow, line 321: The `inspect` functions are great.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -707,24 +738,33 @@ where
...
-        let grpc_client = GrpcClient::new(app_client_io, self.executor.clone(), app_uri)
-            .await
-            .map_err(AppClientError::AppGrpcClientError)?;
 
-        AppClientBuilder::new(Box::new(grpc_client), robot_creds.clone())
-            .build()
-            .await
+        let robot_creds = self.storage.get_robot_credentials().unwrap();
+
+        {
```
acmorrow, line 748: Moved the statements with `?` into a block so I can `inspect` any failure before returning, but otherwise this code didn't change except for logs.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -734,35 +774,58 @@ where
...
+                                    Ok(Some(next_duration)) => {
+                                        duration = next_duration;
+                                    }
+                                    Ok(None) => {}
+                                    e @ Err(_) => {
+                                        break e;
+                                    }
+                                }
+                            }
+                        }
```
acmorrow, line 798: This (I think) replicates the functionality of the `TaskRunnerState` and `AppClientTaskRunner` machinery, but without the need for pin projection and unsafe code. I've verified that a failure in any one task causes the `FuturesUnordered::next` call to return an error, and that all the tasks stop/drop with the `FuturesUnordered`, so I think I'm getting the intended behavior here.

There is an extra scope in there (at one point I was initializing some variables before creating the future, but then they went away.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -734,35 +774,58 @@ where
...
+                        log::error!(
+                            "an app client task returned an error {:?} - dropping app client",
+                            err
+                        );
                         break;
                     }
                 }
             }
+
+            // Explicitly release the app_client resources before sleeping, and before
```
acmorrow, line 812: I don't know if this is actually necessary. I need to better familiarize myself with the lifetime rules in rust, but the theory is that the defunct `AppClient` wouldn't get destroyed until after the timeout and the `self.connect_to_app().await` had completed and returned an object to overwrite it. But there is no point holding on to the dead `AppClient` while we build the new one. If I'm wrong about that and `Rust` will drop it before calling `connect_to_app`, then this can go away.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -859,22 +926,32 @@ where
...
 
-        self.executor.spawn(async move {
-            #[allow(unused_mut)]
-            let mut srv = GrpcServer::new(robot, GrpcBody::new());
-            #[cfg(feature = "local-signaling")]
-            if let Some(ss) = ss {
-                srv.register_signaling_server(ss);
+        log::info!("spawning task for new HTTP2 connection");
+        self.executor.spawn(
+            async move {
```
acmorrow, line 931: I'm not sure why `cargo fmt` decided to push this down, but it did. Perhaps it is related to the `inspect` calls at the end?

Anyway, the only interesting changes in here are new logs.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -909,10 +986,21 @@ where
                     let (answer, prio) = api.answer(0).await?;
                     let robot = self.robot.clone();
 
-                    let task = self.executor.spawn(async move {
-                        let mut conn = api.connect(answer, robot).await?;
-                        conn.run().await
-                    });
+                    log::info!("spawning task for new WebRTC connection");
+                    let task = self.executor.spawn(
+                        async move {
```
acmorrow, line 991: As above for HTTP2 but now for webrtc, only interesting changes are logging and `inspect` calls.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -985,92 +1073,15 @@ where
...
             let incoming = Box::pin(futures_lite::future::or(h2_conn, webrtc_conn)).await;
             if let Err(e) = self.serve_incoming_connection(incoming).await {
-                log::error!("failed to server incoming connection reason {:?}", e)
+                log::error!("failed to serve incoming connection: {:?}", e)
             }
         }
     }
 }
 
-pin_project_lite::pin_project! {
```
acmorrow, line 996: This is now all dead code.

templates/project/src/main.rs
```diff
@@ -55,6 +55,8 @@ fn main() {
     esp_idf_svc::sys::link_patches();
     initialize_logger::<EspLogger>();
 
+    log::info!("{} started (esp32)", env!("CARGO_PKG_NAME"));
```
acmorrow, line 58: I believe this gets evaluated at build time, so this will nicely report the name of the project generated from the template.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -734,35 +774,58 @@ where
...
+
+            log::info!("attempting to reestablish a connection to app for app client tasks");
             app_client = self
                 .connect_to_app()
                 .await
                 .inspect_err(|error| {
-                    if error.is_permission_denied() || error.is_unauthenticated() {
+                    log::error!("couldn't connect to app server, reason {:?}", error);
+                     if error.is_permission_denied() || error.is_unauthenticated() {
+                        log::error!("credentials were positively rejected by app - cached credentials and configuration will be erased");
```
mattjperez, line 828: same as comment above on similar line. these checks around here and above where I commented might be a candidate for DRY

micro-rdk/src/common/conn/viam.rs
```diff
@@ -734,35 +774,58 @@ where
...
+                                        duration = next_duration;
+                                    }
+                                    Ok(None) => {}
+                                    e @ Err(_) => {
+                                        break e;
+                                    }
+                                }
+                            }
+                        }
                     });
```
mattjperez, line 799: noting here from offline discussion,

- `self.app_client_tasks` are set early in startup and don't change
- below where we iterate through the `app_client_tasks`, any error from app_client is an issue for all futures in the Vec. 
  -  (maybe should rename to `task_queue` or something to distinguish from `self.app_client_tasks`)
- we then attempt to create a new app client connection and insert it into the `mut app_client` parameter that was passed in.

mattjperez (approved):
LGTM, left a couple suggestions and a note for my own understanding if we need to revisit.

As for the logs, we can evaluate in practice if they appear to be too chatty.

micro-rdk/src/common/conn/viam.rs
npmenard: So ok to explicitly drop it here although drop will be repeatably called on None until a connection is established.

npmenard (approved):
LGTM with a comment on the change regarding the task runner. I don't think we should do it and if so let's separate it in another ticket

acmorrow:
I've backed out the app client task changes and updated the log statements. I'm planning to merge this if the workflows are clean.

PR #424, opened 2025-03-03
[RSDK-10125] handle basic redirection in OTA

mattjperez:


micro-rdk/src/common/ota.rs
```diff
@@ -334,71 +327,108 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+                    &uri,
+                    CONN_RETRY_SECS
+                );
+                Timer::after(Duration::from_secs(CONN_RETRY_SECS)).await;
+            };
 
-        let conn = Box::new(self.exec.spawn(async move {
-            if let Err(err) = conn.await {
-                log::error!("connection failed: {:?}", err);
+            let conn = Box::new(self.exec.spawn(async move {
```
mattjperez, line 386: From here to L430, all methods propagate errors in lieu of a `break` statement for the outer loop.

micro-rdk/src/common/ota.rs
```diff
@@ -334,71 +327,108 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+                .body(Empty::<Bytes>::new())
+                .map_err(|e| OtaError::Other(e.to_string()))?;
+            response = sender
+                .send_request(request)
+                .await
+                .map_err(|e| OtaError::Other(e.to_string()))?;
+
+            // TODO(RSDK-9346): handle other success codes
+            let status = response.status();
+            match (status.is_success(), status.is_redirection()) {
```
mattjperez, line 405: can switch this for a `if-else` chain if preferable

micro-rdk/src/common/ota.rs
```diff
@@ -304,15 +304,8 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
-    /// Attempts to perform an OTA update.
-    /// On success, returns an `Ok(true)` or `Ok(false)` indicating if a reboot is necessary.
-    pub(crate) async fn update(&mut self) -> Result<bool, OtaError<S>> {
-        if !(self.needs_update()) {
-            return Ok(false);
-        }
-
-        let mut uri = self
-            .url
+    fn parse_uri(&self, url: &str) -> Result<hyper::Uri, OtaError<S>> {
```
mattjperez, line 307: I considered using `fn parse_uri(&mut self)` and setting the `url` within the `OtaService` struct itself, but figured keeping track of the original url may still be useful for debugging later. Though the original can also be found in the robot config. happy to change it if there's a preference

micro-rdk/src/common/ota.rs
```diff
@@ -334,71 +327,108 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+        if !(self.needs_update()) {
+            return Ok(false);
+        }
+
+        let mut uri = self.parse_uri(&self.url)?;
+        let mut response: hyper::Response<hyper::body::Incoming>;
+
+        let conn = loop {
+            let (mut sender, conn) = loop {
+                let mut num_tries = 0;
```
acmorrow, line 345: Before, the initialization of `num_tries` was outside this loop. Did you mean to move it inside?

micro-rdk/src/common/ota.rs
```diff
@@ -334,71 +327,108 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+                            "`location` not found in redirection response header".to_string(),
+                        ));
+                    }
+                    uri = self.parse_uri(headers["location"].to_str().map_err(|e| {
+                        OtaError::Other(format!(
+                            "invalid redirection `location` in header: {} - {:?}",
+                            e, headers
+                        ))
+                    })?)?;
+                    continue;
```
acmorrow, line 421: Will this ever give up if there is a circular redirect? It doesn't look to me like `num_tries` bounds it.

micro-rdk/src/common/ota.rs
```diff
@@ -334,71 +327,108 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+    pub(crate) async fn update(&mut self) -> Result<bool, OtaError<S>> {
+        if !(self.needs_update()) {
+            return Ok(false);
+        }
+
+        let mut uri = self.parse_uri(&self.url)?;
+        let mut response: hyper::Response<hyper::body::Incoming>;
+
+        let conn = loop {
+            let (mut sender, conn) = loop {
```
acmorrow, line 344: Having looked through the whole thing, I wonder if you really need this double loop and could just let `num_tries` drive the whole thing, especially as a way out from a redirect loop.

micro-rdk/src/common/ota.rs
```diff
@@ -334,71 +327,108 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+                let mut num_tries = 0;
+                num_tries += 1;
+                if num_tries == NUM_RETRY_CONN + 1 {
```
acmorrow, line 347: Pushing this down an indent level has made it hard to see what has really changed. I suspect it is possible to either 1) remove the outer loop and restore the indent level, or 2) hoist the interior loop into a function in such a way that there is no real code motion.

micro-rdk/src/common/ota.rs
```diff
@@ -334,71 +327,108 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+                .map_err(|e| OtaError::Other(e.to_string()))?;
+
+            // TODO(RSDK-9346): handle other success codes
+            let status = response.status();
+            match (status.is_success(), status.is_redirection()) {
+                (true, false) => break conn,
+                (false, true) => {
+                    log::debug!("target url has been redirected...");
+                    let headers = response.headers();
+                    if !headers.contains_key("location") {
```
acmorrow, line 410: Is there no constant for this key in hyper?

micro-rdk/src/common/ota.rs
```diff
@@ -334,71 +327,108 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+                        return Err(OtaError::Other(
+                            "`location` not found in redirection response header".to_string(),
```
acmorrow, line 412: Maybe log the actual response code here so we can understand a little better what went wrong (maybe).

acmorrow (changes requested):

micro-rdk/src/common/ota.rs
mattjperez: the re-exported http crate does, just need to confirm that it's exposed. will add if so

micro-rdk/src/common/ota.rs
mattjperez: I could have `num_tries` drive it. or a combination of `num_tries` and `num_redirects` if we wanted to go finer, but I'm ok with `num_tries`

micro-rdk/src/common/ota.rs
mattjperez: I hadn't considered the circular redirect, removing the outer loop as suggested will fix that

micro-rdk/src/common/ota.rs
mattjperez: I purposefully put it in the inner loop. I had used it as number of tries to a specific URL, not the whole process

mattjperez: ah, no, even then it's still wrong :sweat_smile:. will fix it

mattjperez:
was able to simplify to one loop, however needed to add some optional variables to satisfy linter. 
testing now that it still functions as expected

micro-rdk/src/common/ota.rs
```diff
@@ -352,53 +363,97 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
-        // TODO(RSDK-9346): handle other success codes
-        if response.status() != 200 {
-            return Err(OtaError::Other(
-                format!("Bad Request - Status:{}", response.status()).to_string(),
-            ));
+            conn = Some(Box::new(self.exec.spawn(async move {
+                if let Err(err) = inner_conn.await {
+                    log::error!("connection failed: {:?}", err);
+                }
+            })));
```
acmorrow, line 411: [question] I've read this block a few times now, I'm still not sure what it is doing. I know it isn't new, but I'd still like to understand it. Maybe the best action to take here is a comment for this block.

micro-rdk/src/common/ota.rs
```diff
@@ -352,53 +363,97 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+                    }
+                    uri = self.parse_uri(headers[hyper::header::LOCATION].to_str().map_err(
+                        |e| {
+                            OtaError::Other(format!(
+                                "invalid redirection `location` in header: {} - {:?}",
+                                e, headers
+                            ))
+                        },
+                    )?)?;
+                    drop(conn.take());
```
acmorrow, line 445: - [question] Do you need the explicit `drop`? Wouldn't `let _  = conn.take()` do the trick?
- [minor] Log the redirected location at info here.

micro-rdk/src/common/ota.rs
```diff
@@ -352,53 +363,97 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+            response = sender
+                .send_request(request)
+                .await
+                .map_err(|e| OtaError::Other(e.to_string()))?;
+
+            let status = response.status();
+            match (status.is_success(), status.is_redirection()) {
+                (true, false) => break,
+                (false, true) => {
+                    log::debug!("target url has been redirected...");
```
acmorrow, line 428: [optional] I'd make this `info` level, Say `OTA` in it, and give the original URL

micro-rdk/src/common/ota.rs
```diff
@@ -334,14 +327,32 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
-        let (mut sender, conn) = loop {
+        loop {
             num_tries += 1;
             if num_tries == NUM_RETRY_CONN + 1 {
                 return Err(OtaError::Other(
                     "failed to establish connection".to_string(),
                 ));
             }
+
+            let mut sender = None;
```
acmorrow, line 353: [optional] - Might not be a bad idea to `log::info` here that we are on the N attempt to do an OTA download to the current URL.

micro-rdk/src/common/ota.rs
acmorrow: I think the `match` is fine.

micro-rdk/src/common/ota.rs
acmorrow: I think the function is fine as is.

micro-rdk/src/common/ota.rs
```diff
@@ -352,53 +363,97 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
+            if sender.is_none() || inner_conn.is_none() {
+                return Err(OtaError::Other(
+                    "invalid state, error path not handled".to_string(),
+                ));
+            }
+            let mut sender = sender.unwrap();
+            let inner_conn = inner_conn.unwrap();
+
+            if retry {
+                log::debug!(
```
acmorrow, line 398: [optional] - I'd consider having this at `info` level, if not `warn`, with a somewhat more polished message.

acmorrow (changes requested):
Definitely looking better with the single loop. Mostly some optional logging suggestions tagged with `[optional]`, though I do have two less-optional questions about flow of control, tagged with `[question]`.

micro-rdk/src/common/ota.rs
```diff
@@ -352,53 +363,97 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
-        };
-
-        let conn = Box::new(self.exec.spawn(async move {
-            if let Err(err) = conn.await {
-                log::error!("connection failed: {:?}", err);
+            if sender.is_none() || inner_conn.is_none() {
+                return Err(OtaError::Other(
+                    "invalid state, error path not handled".to_string(),
+                ));
+            }
```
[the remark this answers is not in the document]
mattjperez: ah no you're right, it was supposed to come after the retry check. this is a bug.

micro-rdk/src/common/ota.rs
mattjperez: `conn` is the underlying connection and acts like an engine that keeps IO moving, https://docs.rs/hyper/latest/hyper/client/conn/http2/struct.Connection.html.

This is why we maintain a handle to it and drop it after we've finished sending/receiving all the bytes of a given request/response

micro-rdk/src/common/ota.rs
mattjperez: It *should* but I know there are some cases where `let _ = {}` can get optimized out by the compiler. This was me being overly cautious if anything. can switch to that

micro-rdk/src/common/ota.rs
mattjperez: added inline comment

micro-rdk/src/common/ota.rs
acmorrow: It's fine with `drop`.

micro-rdk/src/common/ota.rs
```diff
@@ -334,14 +327,34 @@ impl<S: OtaMetadataStorage> OtaService<S> {
...
                 return Err(OtaError::Other(
                     "failed to establish connection".to_string(),
                 ));
             }
+
+            log::info!("OTA connection attempt {}: `{}` ", num_tries, uri);
+
+            let mut sender = None;
+            let mut inner_conn = None;
+            let retry;
```
acmorrow, line 357: Last thing: I'm not convinced you need the `retry` flag at all. If `sender` and `inner_conn` are set, then you can proceed, otherwise it is an error. Why not just check for `if sender.is_none() || inner_conn.is_none` to decide to retry?

acmorrow (approved):
LGTM with one question about whether `retry` is actually needed.

PR #425, opened 2025-03-05
rename template Cargo.toml to Cargo.toml.liquid

mattjperez:
By having a `cargo-generate` template embedded in the same repo as our `micro-rdk` library, when `cargo` resolves dependencies, the template's `Cargo.toml` is also _seen_ during compilation. Projects using micro-rdk as a library (such as generated projects) then see the following warning when compiling with `make build-esp32-bin`:

```
error: invalid character `{` in package name: `{{project-name}}`, the first character must be a Unicode XID start character (most letters or `_`)
 --> ../../.cargo/git/checkouts/micro-rdk-5c871d18509c9d7e/c015e36/templates/module/Cargo.toml:2:8
  |
2 | name = "{{project-name}}"
  |        ^^^^^^^^^^^^^^^^^^
  |
error: invalid inline table
expected `}`
  --> ../../.cargo/git/checkouts/micro-rdk-5c871d18509c9d7e/c015e36/templates/project/Cargo.toml:23:4
   |
23 |   {% if camera == "true" %}"camera",{% endif %}
   |    ^
   |
   Compiling compiler_builtins v0.1.133
```

This change attempts to fix this, however it's possible that existing projects won't benefit until a new release is made.

### Related
- https://cargo-generate.github.io/cargo-generate/index.html
  > Due to the use of Shopify's Liquid, cargo-generate special cases files with the file-ending .liquid, by simply removing the file-ending when processing the files. If you, as a template author, truly want the .liquid file-ending, you need to repeat it twice!

- https://github.com/cargo-generate/cargo-generate/issues/1333


acmorrow:
Thanks for fixing this! It had bothered me but I hadn't had time to chase it down. LGTM.

acmorrow (approved):

PR #429, opened 2025-03-13
Update CI flow

npmenard:
This is an attempt to speed up CI for micro rdk. I have split the CI between what needs to be run in the micro-rdk docker container and what doesn't. 
I have also enabled caching for build artifact so speed up the process further. 

Caching work well in the native case where I see speed reduction from 6:30 to 2:20, for esp32 builds I saw reductions from 9mins to 6mins. 
For the later we have two bottleneck, first rust-cache doesn't cache build-std crates (see https://github.com/Swatinem/rust-cache/issues/161) second calling make build-esp32-bin forces a partial build of the crates which takes up to 2 minutes I am not exactly sure why this is happening. I am tempted to just run clippy on the esp32 micro-rdk-server we would however loose the ability the confirm the binary doesn't exceed the partition table side.

I didn't change the release workflow so I don't expect that to break


.github/workflows/ci.yml
```diff
@@ -0,0 +1,65 @@
...
+            - 'micro-rdk/src/esp32/**'
+          src-common:
+            - 'micro-rdk/src/common/**'
+            - 'micro-rdk/src/gen/**'
+            - 'micro-rdk-ffi/**'
+            - 'micro-rdk-server/**'
+
+  format:
+    name: Check code formatting
+    uses: ./.github/workflows/format.yml
```
npmenard, line 53: if format fails CI doesn't continue before it has been fixed

.github/workflows/esp32.yml
```diff
@@ -0,0 +1,56 @@
...
+        cache-directories: "/host/.micro-rdk-docker-caches/cargo-registry/"
+        cache-all-crates: true
+    - name: ESP32 Clippy
+      run: |
+        git config --global --add safe.directory /opt/esp/esp-idf
+        cargo clippy -p micro-rdk --lib --features esp32 --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort --release --no-deps
+    - name: ESP32 Server  Clippy & Build
+      run: |
+        cargo clippy -p micro-rdk-server --bin micro-rdk-server-esp32 --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort --release --no-deps
+        make build-esp32-bin
```
npmenard, line 52: tempted to remove this line

.github/workflows/native.yml
```diff
@@ -0,0 +1,46 @@
...
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 2
+    - uses: dtolnay/rust-toolchain@1.83
+      with:
+        components: clippy
+    - name: Cache Dependencies
+      uses: Swatinem/rust-cache@v2
+    - name: install libudev
+      run: sudo apt update && sudo apt install libudev-dev -y
```
npmenard, line 34: needed to build the test suite

Makefile
```diff
@@ -99,18 +97,14 @@ doc:
...
 
-size:
-	find . -name "esp-build.map" -exec ${IDF_PATH}/tools/idf_size.py {} \;
-
-build-esp32-bin: build-esp32-ota
+build-esp32-bin:
 	cargo +esp espflash save-image \
 		--skip-update-check \
 		--package=micro-rdk-server \
-		--features=ota \
```
npmenard, line 109: ota is now a default on micro-rdk do we still need to have a ota feature in the micro-rdk-server crate or is it safe to remove?

.github/workflows/esp32.yml
```diff
@@ -0,0 +1,56 @@
...
+      uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 2
+    # workaround, the environment definition is under the testbot user but github action runs everything as root
+    - name: Move profile
+      run: cp /home/testbot/.bash_profile /etc/profile
+    - name : Set default toolchain to esp
+      run: rustup default esp
+    - name: Cache Dependencies
```
npmenard, line 40: we might want to key the cache and only store it when the CI execute from the main branch. Otherwise I am not sure if a cache will be stored for each PR. Since both cache have a combined size of 1.4 GB we can have and handful of PR opens before evicting older cache

.github/workflows/ci.yml
```diff
@@ -0,0 +1,67 @@
...
+    branches:
+      - '**'
+    tags-ignore:
+      - '**'
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+
+jobs:
```
acmorrow, line 18: Personally, I'm not fond of this. I realize the savings of not doing a build at all are pretty big, but the risk of the filter becoming subtly incorrect outweigh it in my opinion. I'd rather iteratively get caching good enough that we don't care if we do an esp32 build, even if we didn't really change anything esp32 specific.

.github/workflows/esp32.yml
```diff
@@ -0,0 +1,56 @@
+name: CI for ESP32
+
+on:
+  workflow_call:
+
+env:
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: "-D warnings --cfg espidf_time32 "
```
acmorrow, line 8: - I wouldn't expect the `time32` thing to be needed, as it is already in the top-level `.cargo/config.toml`.
- What is the `-D warnings` doing here?

.github/workflows/esp32.yml
```diff
@@ -0,0 +1,56 @@
+name: CI for ESP32
+
+on:
+  workflow_call:
+
+env:
+  CARGO_INCREMENTAL: 0
```
acmorrow, line 7: This comment suggests that `Swatinem/rust-cache` does this already: https://github.com/dtolnay/rust-toolchain/issues/26

Makefile
acmorrow: Safe to remove, I think, unless there is a use-case for disabling it, which I can't think of.

.github/workflows/native.yml
acmorrow: Maybe it belongs in the canon image definition then?

.github/workflows/esp32.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: Cache Dependencies
+      uses: Swatinem/rust-cache@v2
+      with:
+        cache-directories: "/host/.micro-rdk-docker-caches/cargo-registry/"
+        cache-all-crates: true
+    - name: ESP32 Clippy
+      run: |
+        git config --global --add safe.directory /opt/esp/esp-idf
+        cargo clippy -p micro-rdk --lib --features esp32 --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort --release --no-deps
+    - name: ESP32 Server  Clippy & Build
```
acmorrow, line 49: Why have two clippys rather than just running with `--workspace` like we do for native? Or does that not work because of things like the installer?

.github/workflows/format.yml
```diff
@@ -0,0 +1,32 @@
...
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 2
+    - name: Check out PR branch code
+      if: github.event_name == 'pull_request_target'
+      uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 2
+    - uses: dtolnay/rust-toolchain@stable
```
acmorrow, line 27: This should probably be 1.83

Makefile
```diff
@@ -74,10 +72,10 @@ clippy-nmea:
 	cargo clippy -p micro-rdk-nmea --no-deps --features native -- -Dwarnings
 
 clippy-native:
-	cargo clippy -p micro-rdk --no-deps --features native,ota  -- -Dwarnings
+	cargo clippy -p micro-rdk --no-deps --features native,ota --release  -- -Dwarnings
```
acmorrow, line 75: Just FYI but we have some overlapping/conflicting changes between this and #430.

.github/workflows/esp32.yml
acmorrow: Let's keep it for now. Is there a reason you don't want to have it in a separate step rather than combined with clippy?

.github/workflows/native.yml
```diff
@@ -0,0 +1,38 @@
...
+    - name: Clippy Micro-RDK workspace
+      run: cargo clippy --workspace --no-deps --locked --all-targets --features native,ota
+    - name: Tests
+      run: cargo test --workspace --tests --no-fail-fast --features native,ota
```
acmorrow, line 38: I sort of wish every `cargo ...` command we had in the workflows was actually a `make some-target` instead. I'd like there to be one source of truth for what I can build locally that will validate what is going to be built in CI. Without the makefile targets and the CI actions being 1:1, that's hard to do.

.github/workflows/esp32.yml
acmorrow: Isn't the point though to improve PR performance? I don't have a great understanding of how this cache works exactly, and the description on the repo doesn't provide a lot of context.

acmorrow (approved):
LGTM with various questions and opinionated takes.

.github/workflows/ci.yml
npmenard: i agree I considered it but you convinced me to remove it

.github/workflows/esp32.yml
npmenard: yes i saw that, honestly I want to leave it there in case it changes downstream

.github/workflows/esp32.yml
npmenard: me neither, i guess my main concern is that if we have an handful of PRs open they will evict each other's cache. But i would like to test it this way and see the outcomes

.github/workflows/native.yml
npmenard: the idea is to avoid pulling the docker image since it's big

.github/workflows/esp32.yml
```diff
@@ -0,0 +1,51 @@
...
+      if: github.event_name == 'pull_request_target'
+      uses: actions/checkout@v4
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+        fetch-depth: 2
+    # workaround, the environment definition is under the testbot user but github action runs everything as root
+    - name: Move profile
+      run: cp /home/testbot/.bash_profile /etc/profile
+    - name : Set default toolchain to esp
+      run: rustup default esp
```
acmorrow, line 38: With everything as makefile targets having `cargo +esp`, I think you can remove this now.

.github/workflows/esp32.yml
acmorrow: Fine with me. I think land this and see how it goes, then we can iterate.

.github/workflows/format.yml
```diff
@@ -0,0 +1,32 @@
...
+      with:
+        components: rustfmt
+    - name: Format
+      run: cargo fmt --all --check
```
acmorrow, line 31: There is a makefile target that does this. If you want `-D warnings` on it, maybe put that in the makefile and take the `RUSTFLAGS` out above.

.github/workflows/native.yml
acmorrow: I'd missed that we aren't in the docker container here, this is fine.

.github/workflows/format.yml
acmorrow: Ping - I think this needs to be 1.83.

acmorrow:
I'm going to pause further work on #430 until after this merges since there are already conflicts.

PR #430, opened 2025-03-13
[RSDK-10228] Publish more artifacts in generated projects workflow

acmorrow:
This brings the micro-rdk-server and projects generated from the project template up to parity with the work done in the temp sensor project, where we publish not just the OTA artifact on tags, but also the base flashable image and the associated ELF binary.

I'll leave comments along the way.


Makefile
```diff
@@ -111,10 +111,11 @@ build-esp32-bin: build-esp32-ota
 		--bin=micro-rdk-server-esp32 \
 		--partition-table=micro-rdk-server/esp32/partitions.csv \
 		--target=xtensa-esp32-espidf \
-		-Zbuild-std=std,panic_abort --release \
+		-Zbuild-std=std,panic_abort \
+		--release \
 		--flash-size=8mb \
 		--merge \
-		target/xtensa-esp32-espidf/micro-rdk-server-esp32.bin
+		target/xtensa-esp32-espidf/release/micro-rdk-server-esp32.bin
```
acmorrow, line 118: I dislike that we need to name the path to the output directory explicitly here, but I couldn't find a way around it. In any case, the outputs belong in the same directory as the intermediates.

Makefile
```diff
@@ -125,15 +126,16 @@ build-esp32-ota:
 		--bin=micro-rdk-server-esp32 \
 		--partition-table=micro-rdk-server/esp32/partitions.csv \
 		--target=xtensa-esp32-espidf \
-		-Zbuild-std=std,panic_abort --release \
-		target/xtensa-esp32-espidf/micro-rdk-server-esp32-ota.bin
+		-Zbuild-std=std,panic_abort \
+		--release \
+		target/xtensa-esp32-espidf/release/micro-rdk-server-esp32-ota.bin
```
acmorrow, line 131: Ditto, moving into `release`.

templates/project/.github/workflows/publish.yml
```diff
@@ -18,13 +18,17 @@ jobs:
         fetch-depth: 2
     - name: Build esp32 OTA binary
       run: |
-        bash -c '. /home/testbot/.bash_profile ; git config --global --add safe.directory "$ESP_ROOT"/esp-idf && make build-esp32-ota'
-        cp target/xtensa-esp32-espidf/project-ota.bin project-ota.bin
+        bash -c '. /home/testbot/.bash_profile ; git config --global --add safe.directory "$ESP_ROOT"/esp-idf && make build-esp32-bin build-esp32-ota'
+        cp target/xtensa-esp32-espidf/release/{{project-name}} project.elf
+        cp target/xtensa-esp32-espidf/release/{{project-name}}.bin project.bin
```
acmorrow, line 23: We need to know the project name here, since the ELF binary will have that name (we can control the names for the `.bin` files, but as far as I know the ELF file is auto-determined by cargo. So, now this file needs substitution. See below for the sad sad effects.

templates/project/.github/workflows/publish.yml
```diff
@@ -34,27 +38,31 @@ jobs:
     - name: Check release type
       id: check-tag
       run: |
-        if echo ${{ github.event.ref }} | grep -Eq '^refs/tags/v.*rc[0-9]{1}$'; then
+        if echo {% raw %} ${{ github.event.ref }} {% endraw %} | grep -Eq '^refs/tags/v.*rc[0-9]{1}$'; then
```
acmorrow, line 41: The syntax for liquid and github actions conflicts: https://github.com/cargo-generate/cargo-generate/discussions/843. So, any github thing in `${{ ... }` needs to be escaped.

The other way to do this would be to make this file explicitly not part of liquid substitution, and dig the project name out with `cargo metadata` and `jq`:

```
$ cargo metadata --no-deps --format-version=1 | jq '.packages[0].name' -r
awesome
```

We could put the result of that in `GITHUB_ENV` and then use it later.

I didn't do this because the docker image doesn't have `jq` in it, but we could add it of course. LMK if there is preference for that over this admittedly horrible escaping which will plague us indefinitely as we iterate the template workflow.

templates/project/Makefile
```diff
@@ -33,12 +33,14 @@ list-partitions:
 
 build-esp32-bin:
 	cargo espflash save-image \
-		--merge \
-		--chip esp32 \
-		--partition-table partitions.csv \
-		--target xtensa-esp32-espidf \
-		--flash-size ${FLASHSIZE} \
+		--skip-update-check \
```
acmorrow, line 36: flags here had drifted out of sync with standard flags used for builds.

templates/project/.github/workflows/publish.yml
npmenard: does that mean we have to leave the asset name as project.elf?

templates/project/.github/workflows/publish.yml
npmenard: why do you think this will hurt us in the long run?

npmenard (approved):
LGTM with two minor comments:
 1 I think published assets should reuse the project name and not be plainly named project
 2 Without debug symbols I am not sure what is the upside of keeping the ELF in the assets

templates/project/.github/workflows/publish.yml
acmorrow: No, we can call this whatever we want. I can easily interpolate the project name into the files here. If we are doing that, I think is probably worthwhile to put the tag in as well, so:

`project-x.y.z.elf`
`project-x.y.z.bin`
`project-ota-x.y.z.bin`

templates/project/.github/workflows/publish.yml
acmorrow: It will be easy to forget to wrap all github expansions in the `{% raw %}` and `{% endraw %}` tags, and if one is forgotten, it appears that `cargo generate` will just blithely continue on, producing an invalid workflow file, that won't be discovered until a tag is created.

acmorrow:
> LGTM with two minor comments: 1 I think published assets should reuse the project name and not be plainly named project 2 Without debug symbols I am not sure what is the upside of keeping the ELF in the assets

Re 1, I agree, and was just going for minimal changes. I'm happy you think it is worthwhile to do a better job here, and I'll post an update that does so.

Re 2, we can get the elf published now, and when we light up debug symbols it will become useful. Or, I can do that work in this review. Which way do you want to go?

templates/project/.github/workflows/publish.yml
gvaradarajan: Can we isolate this issue by declaring this a variable for the workforce or something? Does such a thing exist?

templates/project/.github/workflows/publish.yml
gvaradarajan: I think this is worth doing right now

gvaradarajan (approved):
LGTM, had input on a couple of conversations

templates/project/.github/workflows/publish.yml
acmorrow: I've decided to go the `jq` path instead, and explicitly exclude the github workflow from liquid interpolation.

templates/project/.github/workflows/publish.yml
acmorrow: I've put the project name in, but left the version out, for now. We can revisit later.

acmorrow:
@npmenard and @gvaradarajan - Requesting a re-review because I just pushed a bunch of additional changes. These amount to:
- Eliminating the use of the GCP bucket everywhere, and now OTA from GH release assets.
- Moved all builds to produce debug symbols and build with opt=`z`, which removed the need for the `release-ffi` build type.
- Changed the template github workflow to use the project name, as derived by `cargo manifest ... | jq ...`
- Prohibited the ill-fated combination of liquid syntax and GH workflow syntax by excluding the template workflow from cargo-generate.

I haven't tested the generated GH worfklows yet but will be doing so this afternoon. Once it all works I'll manually update the temp sensors worfklows to match.

Cargo.toml
```diff
@@ -29,17 +29,13 @@ version = "0.4.1-rc10"
...
-opt-level = "s"
+debug = true
+opt-level = "z"
 
 [profile.dev]
-debug = true # Symbols are nice and they don't increase the size on Flash
+debug = true
 opt-level = "z"
 
-[profile.release-ffi]
```
npmenard, line 38: cool

templates/project/.github/workflows/publish.yml
```diff
@@ -16,20 +21,26 @@ jobs:
       uses: actions/checkout@v3
       with:
         fetch-depth: 2
+    - name: Set Project Name
+      id: set_project_name
+      run: |
+          echo "PROJECT_NAME=`cargo metadata --no-deps --format-version=1 | jq '.packages[0].name' -r`" >> $GITHUB_OUTPUT
```
gvaradarajan, line 27: oh neat

gvaradarajan (approved):
LGTM

templates/project/.github/workflows/publish.yml
acmorrow: Unfortunately, we don't have `jq` in our canon images so I'm going to need to do that before we can merge this.

PR #431, opened 2025-03-18
RSDK-10268 Pin docker image in gh actions

acmorrow:
This way, I can do another PR to update the docker image as I need and push it without it taking immediate effect, and then make another PR to upgrade to that new docker image along with any associated changes in the action definitions. Right now I can't do that.

Also valuable so we can re-run actions from old commits, and lookup what docker image was used to build a given commit, should we ever need to do that archaeology.

This does mean we will need to (remember to) not overwrite tags when pushing docker images, but at least now we can meaningfully do that. I intend to push the next docker images as `1.83.0-1-a{rm,md}64`


npmenard (approved):
LGTM

PR #432, opened 2025-03-18
[RSDK-10228] New canon image and changes to take advantage of it

acmorrow:
This is split in two commits: the first makes the changes to the canon image and the versioning scheme it uses. If/when this PR is approved, I will rebuild the images from that commit and publish them.

The second commit adopts the new image and makes adjustments per the changes to the images.

### DIFF
--- a/.github/workflows/esp32.yml
+++ b/.github/workflows/esp32.yml
@@ -7,18 +7,17 @@ env:
   CARGO_INCREMENTAL: 0
   RUST_BACKTRACE: short
 
-# by default github use sh as shell
-defaults:
-  run:
-    # will use bash to run each command, bash will source /etc/profile which will give us the environment to build for esp32
-    shell: bash --norc  -leo pipefail {0}
-        
 jobs:
   all:
     name: "Build, clippy and test for esp32"
     runs-on: ubuntu-latest
+    # by default github use sh as shell
+    defaults:
+      run:
+        # will use bash to run each command, bash will source /etc/profile which will give us the environment to build for esp32
+        shell: bash --norc  -leo pipefail {0}
     container:
-      image: ghcr.io/viamrobotics/micro-rdk-dev-env:1.83.0-amd64
+      image: ghcr.io/viamrobotics/micro-rdk-dev-env:1.83.0-v2-amd64
     steps:
     - name : Checkout main branch code
       if: github.event_name != 'pull_request_target'
@@ -31,9 +30,6 @@ jobs:
       with:
         ref: ${{ github.event.pull_request.head.sha }}
         fetch-depth: 2
-    # workaround, the environment definition is under the testbot user but github action runs everything as root
-    - name: Move profile
-      run: cp /home/testbot/.bash_profile /etc/profile
     - name : Set default toolchain to esp
       run: rustup default esp
     - name: Cache Dependencies
@@ -48,4 +44,3 @@ jobs:
     - name: ESP32 Server Build
       run: |
         make build-esp32-bin
-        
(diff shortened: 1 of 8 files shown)


.github/workflows/esp32.yml
```diff
@@ -7,18 +7,17 @@ env:
...
-defaults:
-  run:
-    # will use bash to run each command, bash will source /etc/profile which will give us the environment to build for esp32
-    shell: bash --norc  -leo pipefail {0}
-        
 jobs:
   all:
     name: "Build, clippy and test for esp32"
     runs-on: ubuntu-latest
+    # by default github use sh as shell
```
acmorrow, line 14: I moved these down a scope, into jobs that have a `container` set. I didn't see much need to override the shell for any other jobs that happened to run outside the container.

npmenard: Fair

.github/workflows/esp32.yml
```diff
@@ -7,18 +7,17 @@ env:
...
+    defaults:
+      run:
+        # will use bash to run each command, bash will source /etc/profile which will give us the environment to build for esp32
+        shell: bash --norc  -leo pipefail {0}
     container:
-      image: ghcr.io/viamrobotics/micro-rdk-dev-env:1.83.0-amd64
+      image: ghcr.io/viamrobotics/micro-rdk-dev-env:1.83.0-v2-amd64
```
npmenard, line 20: is the v2 temporary? not sure that I like it

etc/docker/docker.make
```diff
@@ -1,32 +1,55 @@
...
+
+# The compatibility version for these images. Bump this when making
+# incompatible changes to the images (e.g. changes for which there
+# need to be compensating changes in the Micro-RDK, its templates,
+# github actions, or generated projects, or when delivering a new
+# feature in the image that will be relied on by micro-rdk. After
+# pushing the images with `micro-rdk-upload`, adjust the pins in the
+# Micro-RDK repository accordingly. The pins should not specify the
+# minor version (e.g. use `1.83.0-v2`, not `1.83.0-v2.1`).
+BUILD_TAG_MAJOR_VER=2
```
npmenard, line 17: I see.
 I don't think we want to keep major an minor version of the micro-rdk docker container, we just don't make enough changes to justify that IMO.

npmenard: But i imagine the concern stems from user using the docker image to do cloud builds

npmenard (approved):
I am uneasy with the way we are tagging image now. I think this is justified to support cloud builds but it feels too early

LGTM anyhow  happy to talk about it tomorrow

etc/docker/docker.make
acmorrow: There are two concerns, really. The first is that we want to be able to advance the container without breaking existing generated projects, which would happen today if we pushed again to the `1.83.0:a{rm,md}64` tags, and the second is that we want to be able to deterministically rebuild old commits/tags (e.g. for bisecting). A third, I suppose, would come if/when we have multiple stable release streams and need to advance the container on `main` without altering what the stable branches use.

Having considered it some more overnight, I also don't like the minor version, since it undermines the primary aim here, which is that every micro-rdk commit has a determined and immutable container version against which it runs. If the minor version can change out from under, that defeats that aim. So, I see a few things we could do instead:

- Drop minor, and just have an incrementing build number (e.g. `v1`, `v2`, etc). In which case, do we even care about having the rust version in there too?
- Use a datestamp instead (e.g. `ghcr.io/viamrobotics/micro-rdk-dev-env-20250319:arm`)
- Have references in micro-rdk use an image digest instead of a tag (I particularly dislike this)

So, basically, I'm definitely open to other ideas that achieve the build reproducibility thing, but I do think it is fairly important and that we can get there without too much effort from where we are now. Also, it is important to note that image tags are not immutable since you can overwrite them, and there doesn't seem to be a way to disable that on ghcr, unfortunately. Still, I think versioning these somehow is a net advance over our current state.

acmorrow: Per discussion, we've decided against starting versioning right now, but will revisit as we work on ESP-IDF 5.

PR #433, opened 2025-03-18
[RSDK-10120] Projects from templates can use provisioning

acmorrow:
A small but important one. Without this, flashable (rather than OTA) firmware cannot be generated on GH for projects produced from the template - an important part of our cloud build plan.


npmenard (approved):
LGTM

PR #434, opened 2025-03-19
[RSDK-10228] Fixup passing of project name between jobs in template

acmorrow:
I'd erroneously extrapolated the syntax for accessing outputs from one job in another, this fixes it. Also a small fix where it was missing `-ota`.

I've verified that a project built from this template does the right thing (in fact, that's how I discovered it was wrong in the first place).


mattjperez (approved):
LGTM

PR #437, opened 2025-03-20
[RSDK-10193] [RSDK-10203] Upgrade embedded-svc for compat with esp-idf-svc

acmorrow:
As a side effect, fixes the ambiguities in the wifi code. It is a little strange to me that `esp-idf-svc` doesn't re-export the `embedded-svc` crate (as it does for `-sys` and `-hal`) if the two need to be kept in lockstep, because there doesn't seem to be a way to express in our `Cargo.toml` that we want to depend on the same version of `embedded-svc` that `esp-idf-svc` does.

This keeps the `~` pin on the version, since I think we don't want `embedded-svc` advancing a minor version on us without `esp-idf-svc` advancing, and vice versa. It also now applies the pin in the template, where I think it was missing before.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1404,25 +1404,6 @@ dependencies = [
 
 [[package]]
 name = "embedded-svc"
-version = "0.27.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


gvaradarajan (approved):
LGTM

npmenard:
@acmorrow closing this in preference of #441 (it also fix the wifi configuration issue)

PR #439, opened 2025-03-21
[RSDK-10195] Update watchdog logic for ESP-IDF 5

gvaradarajan:


npmenard (changes requested):

micro-rdk/src/common/conn/viam.rs
```diff
@@ -431,10 +431,17 @@ where
     pub fn run_forever(&mut self) -> ! {
         #[cfg(feature = "esp32")]
         {
+            let wdt_cfg = crate::esp32::esp_idf_svc::sys::esp_task_wdt_config_t {
+                timeout_ms: (180 * 10_u32.pow(6)), // 180 seconds in milliseconds
```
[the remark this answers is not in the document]
gvaradarajan: oops, thought I put 3

micro-rdk/src/common/conn/viam.rs
```diff
@@ -431,10 +431,16 @@ where
     pub fn run_forever(&mut self) -> ! {
         #[cfg(feature = "esp32")]
         {
+            let wdt_cfg = crate::esp32::esp_idf_svc::sys::esp_task_wdt_config_t {
+                timeout_ms: (180 * 10_u32.pow(3)), // 180 seconds in milliseconds
+                trigger_panic: true,
+                // TWDT wants to know the bitmask for which core's idle task it should subscribe to
+                idle_core_mask: 1 << unsafe { crate::esp32::esp_idf_svc::hal::cpu::core().into() },
```
npmenard, line 438: is their a reason why we subscribe the current core only here?

gvaradarajan: It seems to me the only task handle we add to the watchdog is the one right below and so it's only the current task's cpu matters. Have I misunderstood?

npmenard: i think `idle_core_mask` tells the TWDT whic idle core tasks needs to be watched

gvaradarajan: Sorry, I think I still don't understand. I thought the implication [here](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/system/wdts.html#_CPPv4N21esp_task_wdt_config_t14idle_core_maskE) that if I supply `1 << 0` then it subscribes to the idle task of core 0 and `1 << 1` would have it subscribe to core 1. How do I indicate both cores in this scenario?

npmenard: (1 << CONFIG_FREERTOS_NUMBER_OF_CORES) - 1 should work

PR #441, opened 2025-03-21
[RSDK-10193] upgrade crates

npmenard:
upgrade embedded-svc & esp-idf-svc so types are compatible.
Since esp-idf-svc exposes the scan sort methods I went ahead and removed our custom code

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -62,12 +62,11 @@ dns-message-parser = { version = "0.7", default-features = false}
 either = "1.13.0"
 # TODO(RSDK-8991): Upgrade to latest `embedded-hal`
 embedded-hal = { version = "~0.2", features = ["unproven"] }
-# TODO(RSDK-10203): Upgrade `embedded-svc` to latest release (in sync with esp-idf-svc?)
-embedded-svc = "0.28"
+embedded-svc = "~0.28"
 embuild = "0.32.0"
 env_logger = "0.11.5"
 esp-idf-part = "0.5.0"
-esp-idf-svc = { version = "0.51", default-features = false }
+esp-idf-svc = { version = "~0.51", default-features = false }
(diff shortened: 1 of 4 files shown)


micro-rdk/src/esp32/conn/network.rs
```diff
@@ -116,6 +112,8 @@ impl Esp32WifiNetwork {
             auth_method: AuthMethod::None,
             password: "".try_into().unwrap(),
             channel: None,
+            scan_method: ScanMethod::CompleteScan(ScanSortMethod::Signal),
```
gvaradarajan, line 115: oh sick, that's a fun new thing

gvaradarajan (approved):
LGTM

Cargo.toml
```diff
@@ -47,7 +47,7 @@ atomic-waker = "1.1.2"
 axum = "0.7.9"
 axum-extra = "0.9.6"
 base64 = "0.22.1"
-bincode = "2.0.0-rc.3"
+bincode = "=2.0.0-rc.3"
```
acmorrow, line 50: Please make a new ticket to upgrade `bincode` to 2.0 GA, and `// TODO(RSDK-...): ...` it here. That ticket can depend on https://viam.atlassian.net/browse/RSDK-9716 which will move us to Rust 1.85+ at which point we won't need the pin anymore, so it would be nice to let it advance naturally from then on.

acmorrow (approved):
LGTM with a request for a ticket and associated `TODO` on the `bincode` pin in `Cargo.toml`.

Cargo.toml
npmenard: Yes I don't know what this crate is doing in our tree. It seems to be reference by ota for native build but I don't think we support it? @mattjperez  can confirm whether the native code for OTA update the binary

PR #443, opened 2025-03-21
[RSDK-10190] remove lease

npmenard:
Dhcp lease type are not exposed by bindgen anymore (before they were via the tcpip_adapter_types.h) To re-expose would require adding components/esp_netif/lwip/esp_netif_lwip_internal.h to the list of headers.

### DIFF
--- a/micro-rdk/src/esp32/conn/network.rs
+++ b/micro-rdk/src/esp32/conn/network.rs
@@ -1,4 +1,3 @@
-use core::ffi::c_void;
 use std::{
     cell::RefCell,
     ffi::CString,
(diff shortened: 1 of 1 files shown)


gvaradarajan (commented):
Do we use these temporary IPs for any important purpose?

acmorrow (approved):
LGTM, though I must admit the intention/effects of the code being removed is opaque to me. For posterity, would you mind leaving a quick note in this review about what this code was doing and why, and why we no longer need to do that, or have decided it isn't worth it vs grabbing the symbols from LWIP?

PR #444, opened 2025-03-21
[RSDK-8998] remove symbols

npmenard:
Some symbols are not defined anymore in ESP-IDF5

### DIFF
--- a/micro-rdk-ffi/src/ffi/runtime.rs
+++ b/micro-rdk-ffi/src/ffi/runtime.rs
@@ -30,11 +30,6 @@ pub struct viam_server_context {
     storage: Vec<String>,
 }
 
-#[cfg(target_os = "espidf")]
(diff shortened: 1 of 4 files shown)


acmorrow:
I'd like a few quick clarifications
- Is it the `g_spiram_ok` symbol that went away? Or also the `g_wifi_feature_caps`? `CONFIG_FEATURE_CACHE_TX_BUF_BIT`?
- Is the intention here that we basically no longer care about boards without SPIRAM, such that it isn't worth finding another way to achieve this?

acmorrow (approved):
LGTM as it is purely removal, but I left some questions on a review level comment. Also, I'd prefer to see this merge with a commit message referencing a ticket. If the symbols really just got removed as part of going to ESP-IDF 5, I think you could put it under RSDK-8998.

PR #446, opened 2025-03-24
[RSDK-10191] use newer EspFirmwareLoad api

mattjperez:


micro-rdk/src/common/ota.rs
```diff
@@ -525,18 +525,25 @@ impl<S: OtaMetadataStorage> OtaService<S> {
                             #[cfg(feature = "esp32")]
                             {
                                 log::info!("verifying new ota firmware");
-                                let mut loader = EspFirmwareInfoLoader::new();
-                                loader
-                                    .load(&data)
+                                let mut new_fw = FirmwareInfo {
```
mattjperez, line 528: note: can't use `..Default::default()` because `FirmwareInfo` does not  `impl Default`

PR #447, opened 2025-03-24
fix qemu build

npmenard:


Makefile
```diff
@@ -37,7 +37,7 @@ native:
 	cargo run -p micro-rdk-server --bin micro-rdk-server-native
 
 build-qemu:
-	cargo +esp build -p micro-rdk-server --bin micro-rdk-server-esp32  --features qemu --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort && cargo +esp espflash save-image --package micro-rdk-server --features qemu --merge --chip esp32 target/xtensa-esp32-espidf/debug/debug.bin -T micro-rdk-server/esp32/partitions.csv -s 4mb  --bin micro-rdk-server-esp32 --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort
+	cargo +esp build -p micro-rdk-server --bin micro-rdk-server-esp32  --features qemu --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort --release && cargo +esp espflash save-image --release --package micro-rdk-server --features qemu --merge --chip esp32 target/xtensa-esp32-espidf/release/qemu.bin -T micro-rdk-server/esp32/partitions.csv -s 8mb  --bin micro-rdk-server-esp32 --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort
```
acmorrow, line 40: - I'm curious: do you really need the two step build here? Would just the `cargo espflash save-image ... --features qemu` do the trick?
- What would you think of moving this target down, closer to `build-esp32-bin` and `build-esp32-ota`, maybe calling it `build-qemu-bin` for consistency, and re-ordering and linebreaking the arguments to make it consistent with the `save-image` calls for the other two builds? I suspect it would be less likely to drift out of alignment if it was positioned and written that way.

acmorrow (approved):
LGTM, with a small question and an optional suggestion.

PR #448, opened 2025-03-25
[RSDK-10191] FirmwareInfo usage comment

mattjperez:


micro-rdk/src/common/ota.rs
```diff
@@ -543,6 +546,11 @@ impl<S: OtaMetadataStorage> OtaService<S> {
                                     );
                                     log::debug!("new firmware app description: {:?}", new_fw);
                                     got_info = true;
+                                } else {
+                                    log::warn!(
```
acmorrow, line 550: If this log gets hit, then we pull another chunk, I think. And it tries to look for the header again. So we will 1) keep logging this, and 2) if the header is never found will it just keep downloading forever? The loop is very complex so it is hard to see the flow of control. Could you just confirm for me what the sequence of events will be if you point it to a very large file that never contains the firmware header?

acmorrow (approved):

PR #449, opened 2025-03-25
[RSDK-10188] esp-idf v5 ADC APIs

mattjperez:
The main differences:
1. `AdcChannelDriver` is initialized with an `AdcDriver`, [example](https://docs.esp-rs.org/esp-idf-svc/esp_idf_svc/hal/adc/oneshot/index.html)
2. The `attenuation setting is passed as an struct input to the configuration instead of as a type parameter when initializing the channel driver.
3. Removed the `Arc<Mutex<>>` from the struct entirely


micro-rdk/src/esp32/board.rs
```diff
@@ -84,104 +88,98 @@ impl EspBoard {
...
                         .iter()
                         .map(|v| {
-                            let adc1 = Arc::new(Mutex::new(AdcDriver::new(
-                                unsafe { ADC1::new() },
-                                &Config::new().calibration(true),
-                            )?));
+                            let adc1 = AdcDriver::new(unsafe { ADC1::new() })?;
+                            let config = AdcChannelConfig {
+                                attenuation: DB_11,
+                                calibration: Calibration::Line,
```
npmenard, line 94: What does that do and why do we set it to line?

micro-rdk/src/esp32/analog.rs
```diff
@@ -1,42 +1,33 @@
...
+use std::borrow::Borrow;
 
-// TODO(RSDK-10188): Update to ESP-IDF 5 ADC API
-use crate::esp32::esp_idf_svc::hal::gpio::ADCPin;
-use std::sync::{Arc, Mutex};
-
-pub struct Esp32AnalogReader<'a, const A: u32, T: ADCPin> {
-    channel: AdcChannelDriver<'a, A, T>,
-    driver: Arc<Mutex<AdcDriver<'a, T::Adc>>>,
+pub struct Esp32AnalogReader<'a, T: ADCPin, M: Borrow<AdcDriver<'a, T::Adc>>> {
```
npmenard, line 9: I don't like the increase of complexity of types parameters, is there a way we can make it more readable. Like storing only the driver and building the channel on each read? Is building a channel expensive?

npmenard (changes requested):
Done with first pass

micro-rdk/src/esp32/board.rs
mattjperez: previously we did `&Config::new().calibration(true)`. The `Calibration` enum has three variants: `None`, `Line`, `Curve` according to the [source code](https://docs.esp-rs.org/esp-idf-hal/src/esp_idf_hal/adc.rs.html#532-553). However, only `None` and `Line` are available to our build. So I used `Line` to in place of the previous `calibration(true)`.

npmenard: make sense :ok_hand:

micro-rdk/src/esp32/board.rs
```diff
@@ -84,104 +88,98 @@ impl EspBoard {
                     let analogs: Result<Vec<AnalogReaderType<u16>>, BoardError> = analogs
                         .iter()
                         .map(|v| {
-                            let adc1 = Arc::new(Mutex::new(AdcDriver::new(
-                                unsafe { ADC1::new() },
-                                &Config::new().calibration(true),
-                            )?));
+                            let adc1 = AdcDriver::new(unsafe { ADC1::new() })?;
```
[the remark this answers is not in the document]
mattjperez: this was porting the same logic from the previous implementation, however we determined a few things:
1. the previous implementation was wrong, initializing ADC1 multiple times
2. the previous implementation still allowed for multiple adc pins to be configured without user-observable 'error', 
3. the new ADC1 throws a recoverable error when initializing ADC1 more than once

I've since moved the ADC1 init to outside of the `map` and wrapped in an `Arc`. Confirmed that multiple adc pins come up when configured.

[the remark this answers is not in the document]
mattjperez: as a follow-up, we should probably either:
a. store the board [`Peripherals`](https://docs.esp-rs.org/esp-idf-hal/esp_idf_hal/peripherals/struct.Peripherals.html) in our `Esp32Board` struct
b. store initialized `Arc<Driver>`s in our board

micro-rdk/src/esp32/analog.rs
mattjperez: I think that may actually make the type even more complex. 

At the very least the `new` input parameters would either need to be:
1. `(name: String, driver: AdcDriver<'d, T::Adc>, pin: impl Peripheral<P = T> + 'd,
    config: AdcChannelConfig)` if we want to create the `AdcChannelDriver` on-demand
2. `(name: String, adc: impl Adc, pin: impl Peripheral<P = T> + 'd, config: AdcChannelConfig)` if we want to handle both the `AdcDriver` and `AdcChannelDriver` initialization in the `Esp32AnalogReader`.

PR #450, opened 2025-03-26
[RSDK-10306] initialize empty networks settings in NVS on startup

mattjperez:
Not sure if there's a better place for this to live. Here felt like a more natural place than in the config monitor logic.


micro-rdk/src/common/conn/viam.rs
```diff
@@ -502,6 +502,13 @@ where
         // if wifi manager is configured loop forever until wifi is connected via
         // a the provisioned network or one from previously stored agent config
         if let Some(wifi) = self.wifi_manager.as_ref().as_ref() {
+            if !self.storage.has_network_settings() {
+                log::info!("initializing additional networks storage...");
```
acmorrow, line 506: - Remove the ellipsis
- Maybe say "setting additional networks list to empty in storage since it is not present" or something like that. `Initializing` says what we are doing but not why.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -502,6 +502,13 @@ where
...
+                if let Err(e) = self.storage.store_network_settings(Default::default()) {
+                    log::error!("failed to initialize new network settings storage: {}", e);
```
acmorrow: Make this error text align with whatever you pick above.

acmorrow (approved):
I think this is a reasonable place to do this. LGTM

micro-rdk/src/common/conn/viam.rs
mattjperez: yup, just noticed and made that change

acmorrow (approved):

PR #452, opened 2025-03-27
[RSDK-10288] sta disconnect events log

mattjperez:
I took this approach with the enum because:
1. all of the error codes are constants at the top level of `esp_idf_sys`
2. they are defined as `c_uint` which are typically `u32` however they can be `u16` depending on the platform
3. because enum discriminants are typically `i32`s, I wasn't confident that setting the variants directly would be consistent between the bindings and the enum (`enum WifiError { Roaming = esp_idf_sys::wifi_err_ROAMING ... }`).
4. The `wifi_event_sta_disconnected_t`'s `reason` field is defined as a `u8` in the bindings

Most of the time was spent hand-rolling the enum a few times as I realized the above.

 still needs to be tested to see that the error messages are present and useful


micro-rdk/src/esp32/conn/network.rs
```diff
@@ -609,6 +609,12 @@ impl Esp32ExternallyManagedNetwork {
             && ev_id == esp_idf_svc::sys::wifi_event_t_WIFI_EVENT_STA_DISCONNECTED
         {
             data.connected.store(false, Ordering::Release);
+            if !ev_data.is_null() {
+                let disconn: *const esp_idf_svc::sys::wifi_event_sta_disconnected_t =
+                    unsafe { core::mem::transmute(ev_data) };
+                let reason: super::wifi_error::WifiErrReason = (*disconn).reason.into();
+                log::error!("sta disconnected: {:?}", reason);
```
acmorrow, line 616: I believe that the intention of this ticket (which was admittedly light on details) was to actually inspect the internal structure of the `StaConnected` and `StaDisconnected` events up around line 200 and report interesting context around these events (probably disconnection is generally more interesting).

mattjperez: that's probably a better place than where I found originally, yeah.
I shifted the logging to there if you'd like to take a look.

So does that mean we want to look deeper into `StaConnected` as well for this ticket? It seems we would be able to log the ssid but not much else, https://docs.esp-rs.org/esp-idf-svc/esp_idf_svc/wifi/struct.StaConnectedRef.html

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -202,25 +202,30 @@ impl Esp32WifiNetwork {
...
-                                .get_configuration()
-                                .map_or("<no_ssid>".to_owned(), |c| {
-                                    c.as_client_conf_ref().unwrap().ssid.to_string()
-                                });
-                            log::error!("could not connect to WiFi {} cause : {:?}", ssid, err);
+        let subscription =
+            sl_stack.subscribe::<WifiEvent, _>(move |event: WifiEvent| match event {
+                WifiEvent::StaDisconnected(disconnected) => {
+                    let reason: super::wifi_error::WifiErrReason = disconnected.reason().into();
+                    log::info!("StaDisconnected event received: {:?}", reason);
```
acmorrow, line 209: I wouldn't bother writing `StaDisconnected event` here - it isn't really a meaningful bit of information for a consumer of the logs.

Instead, let's focus the log on user facing info: let's log the SSID, the reason, and, the RSSI. Something like:

```
log::info!("received a WiFi disconnection event for SSID {} (RSSI {}) with reason {}"
```

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -202,25 +202,30 @@ impl Esp32WifiNetwork {
...
+                            }
                         }
                     }
                 }
-            } else if matches!(event, WifiEvent::StaConnected(_)) {
-                log::info!("wifi connected event received");
-            }
-        })?;
+                WifiEvent::StaConnected(_) => {
+                    log::info!("wifi connected event received");
```
acmorrow, line 225: Similar thoughts. I think it is valuable to at least log the SSID here:

```
log::info!("received a WiFi connection event for SSID {}"
```

micro-rdk/src/esp32/conn/network.rs
acmorrow, line 209: From your slack message, I can see that logging the reason with `{:?}` is going to look like `GroupKeyUpdateTimeout(16)`. However, the 16 isn't really helpful there, maybe. It makes it look like the 16 is somehow parameterizing the `GroupKeyUpdateTimeout`, but it is really just the underlying code. Maybe `Debug` on `WifiErrReason` isnt' the right way to log it.

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -202,25 +202,30 @@ impl Esp32WifiNetwork {
...
-                            let ssid = wifi_mut
-                                .get_configuration()
-                                .map_or("<no_ssid>".to_owned(), |c| {
-                                    c.as_client_conf_ref().unwrap().ssid.to_string()
-                                });
-                            log::error!("could not connect to WiFi {} cause : {:?}", ssid, err);
+        let subscription =
+            sl_stack.subscribe::<WifiEvent, _>(move |event: WifiEvent| match event {
+                WifiEvent::StaDisconnected(disconnected) => {
+                    let reason: super::wifi_error::WifiErrReason = disconnected.reason().into();
```
acmorrow, line 208: Maybe import it, since it will also get used in the connected event below per my comment?

acmorrow (changes requested):

micro-rdk/src/esp32/conn/wifi_error.rs
```diff
@@ -0,0 +1,171 @@
+#![allow(non_snake_case)]
+use esp_idf_svc::sys::*;
+use core::ffi::c_uint;
+
+impl From<u16> for WifiErrReason {
```
acmorrow, line 5: I think I'd rather get a compilation error if we try to build on a platform where C's `unsigned int` isn't 32 bits. That's a really rare platform.

micro-rdk/src/esp32/conn/wifi_error.rs
```diff
@@ -0,0 +1,171 @@
...
+pub enum WifiErrReason {
+    Unspecified(c_uint),
+    AuthExpire(c_uint),
+    AuthLeave(c_uint),
+    AssocExpire(c_uint),
+    AssocTooMany(c_uint),
+    NotAuthed(c_uint),
+    NotAssoced(c_uint),
+    AssocLeave(c_uint),
+    AssocNotAuthed(c_uint),
```
acmorrow, line 118: I have to wonder if there isn't some macro crate out there that can emit this kind of boilerplate for us given a C enum.

Also surprised that esp-idf-{svc,sys} doesn't do this for us.

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -202,25 +202,54 @@ impl Esp32WifiNetwork {
...
-                            let ssid = wifi_mut
-                                .get_configuration()
-                                .map_or("<no_ssid>".to_owned(), |c| {
-                                    c.as_client_conf_ref().unwrap().ssid.to_string()
-                                });
-                            log::error!("could not connect to WiFi {} cause : {:?}", ssid, err);
+        let subscription =
+            sl_stack.subscribe::<WifiEvent, _>(move |event: WifiEvent| match event {
+                WifiEvent::StaDisconnected(disconnected) => {
+                    let ssid: String = CString::new(disconnected.ssid())
```
acmorrow, line 208: I believe there is no restriction against having null characters in an SSID, so this would truncate. What other options are there for logging a `&[u8]` that don't require first converting into a CString?

acmorrow (approved):
LGTM though I have a question/concern about moving the SSID through a CString representation on its way to the log.

PR #453, opened 2025-03-28
[RSDK-10281] Fix deprecation warnings in project template

acmorrow:
Also, cleanup logging terminology and harmonize between viam-micro-server and template projects.


gvaradarajan (approved):
LGTM

PR #454, opened 2025-03-31
[RSDK-9841] Use COMPONENT_VERSION in FFI component with ESP-IDF 5

acmorrow:
We tried to do this here https://github.com/viamrobotics/micro-rdk/commit/0ccc9fc29133b1baa5b96651d98226f39240ee2d#diff-ed1421c19afa77da895d13381af64b9e0d604a7deda389355470f832609962d7L4 but we couldn't yet because `COMPONENT_VERSION` doesn't exist before ESP-IDF 5.

This just gets rid of one more touchpoint for updating versions during release.

I don't plan to merge this until we have a working ESP-IDF 5 compile so I can test it out, but I had it in my tree waiting, so it seemed worth getting it up for review.


mattjperez (approved):
nice

PR #456, opened 2025-03-31
[RSDK-9066] common certs bundle

mattjperez:
should we include these changes in the ffi sdkconfig as well? or up to the downstream user? @npmenard 

```sh
# before
App/part. size:    3,560,608/4,063,232 bytes, 87.63%

# after
App/part. size:    3,507,712/4,063,232 bytes, 86.33%

# ~50KB savings
```


micro-rdk-server/sdkconfig.defaults
```diff
@@ -14,10 +14,8 @@ CONFIG_ESP32_DEFAULT_CPU_FREQ_240=y
 CONFIG_ESPTOOLPY_FLASHFREQ_80M=y
 CONFIG_ESPTOOLPY_FLASHMODE_DIO=y
 
-
 CONFIG_MBEDTLS_CERTIFICATE_BUNDLE=y
-// When moving to IDF5 we should be able to use CONFIG_MBEDTLS_CERTIFICATE_BUNDLE_DEFAULT_CMN only
-CONFIG_MBEDTLS_CERTIFICATE_BUNDLE_DEFAULT_FULL=y
+CONFIG_MBEDTLS_CERTIFICATE_BUNDLE_DEFAULT_CMN=y
```
mattjperez, line 18: the `CONFIG_MBEDTLS_CERTIFICATE_BUNDLE` is on by default. not sure why we make it explicit

templates/project/sdkconfig.defaults
```diff
@@ -17,6 +17,8 @@ CONFIG_ESPTOOLPY_FLASHMODE_DIO=n
 
 CONFIG_ESP_TLS_SERVER=y
 CONFIG_MBEDTLS_SSL_PROTO_DTLS=y
+CONFIG_MBEDTLS_CERTIFICATE_BUNDLE=y
+CONFIG_MBEDTLS_CERTIFICATE_BUNDLE_DEFAULT_CMN=y
```
mattjperez, line 21: I didn't see any `MBEDTLS_CERTIFICATE` values in the template. not sure if they were left out on purpose or not

acmorrow: Definitely add them.

Would you mind filing a follow-up ticket to examine and resolve the differences between the sdkconfig for the viam-micro-server and what is in the template?

acmorrow (approved):
LGTM.

Were you able to measure any change in the image size with this? I think it is worth double checking that we achieve the savings we were looking for.

templates/project/sdkconfig.defaults
mattjperez: [RSDK-10363](https://viam.atlassian.net/browse/RSDK-10363?atlOrigin=eyJpIjoiMTNlYzIzYWYxNDMyNDIwZjhjMDczM2M5OTBhMjJkM2IiLCJwIjoiaiJ9)

PR #458, opened 2025-04-01
[RSDk-8991] update embedded-hal to v1.0.0

mattjperez:
still testing, but it compiles


acmorrow (commented):
That was easy! LGTM if your testing is OK.

acmorrow (approved):

PR #460, opened 2025-04-01
Remove builder and building qemu

npmenard:
quick change to bypass building qemu and download pre-compiled binaries instead

### DIFF
--- a/etc/docker/Dockerfile
+++ b/etc/docker/Dockerfile
@@ -1,25 +1,3 @@
-FROM rust:1.83-bookworm AS builder
-RUN apt update && apt-get install -y --no-install-recommends git libudev-dev make bzip2 gcc libgcrypt20 libgcrypt20-dev  ninja-build libudev-dev sudo wget git flex bison gperf ccache libffi-dev libssl-dev dfu-util libusb-1.0-0 cmake pkg-config libglib2.0-dev libpixman-1-dev software-properties-common openssh-client libslirp-dev python3-venv
-
-RUN export TARGET=$(rustc -vV | sed -En 's/^host: //p') && \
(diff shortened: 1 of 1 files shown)


etc/docker/Dockerfile
```diff
@@ -36,13 +14,23 @@ ENV RUSTUP_HOME=/opt/rust/rustup \
...
-
-RUN apt-get install -y --no-install-recommends git wget flex bison gperf python3 python3-pip python3-virtualenv cmake ninja-build ccache libffi-dev libssl-dev libudev-dev dfu-util libusb-1.0-0 openssh-client software-properties-common python3-venv sudo curl && \
+RUN apt-get install -y --no-install-recommends make build-essential pkg-config git wget flex bison gperf python3 python3-pip python3-virtualenv cmake ninja-build ccache libffi-dev libssl-dev libudev-dev dfu-util libusb-1.0-0 openssh-client software-properties-common python3-venv sudo curl libpixman-1-0 libslirp0 libsdl2-2.0-0 xz-utils && \
     apt-get install -y --no-install-recommends cmake && \
     apt-get install -y --no-install-recommends jq && \
     update-alternatives --install /usr/bin/python python /usr/bin/python3 10
 
+RUN set -eux; \
+    dpkgArch="$(dpkg --print-architecture)"; \
+    qemuSha="https://github.com/espressif/qemu/releases/download/esp-develop-9.0.0-20240606/qemu-esp_develop_9.0.0_20240606-checksum.sha256"; \
```
mattjperez, line 24: sanity check on your heiroglyphics, so you're initializing `qemuSha` then updating it depending on the `dpkgArch`?

you're then downloading the binary, extracting the tar to  `/usr` then removing the original file?

mattjperez (approved):
LGTM

etc/docker/Dockerfile
npmenard: yes was wrong :P 

yes indeed

PR #457, opened 2025-04-01
[RSDK-10360] Upgrade to Rust 1.85

acmorrow:
- `s/1.83/1.85g`
- Various clippy fixes that show up on new toolchain
- Explicitly select ESP-IDF v5.4.1 everywhere.
- Commit the `components_esp32.lock` file, which is now meaningful since we are using the mdns component

Once this is ready to go, I'll push the docker containers and restart the jobs so we can validate it works.

PR #461, opened 2025-04-03
fix linting errors, rm ota from native

mattjperez:
these are just small changes to 
~1. address an unused import warning~ 
2. rm ota from native compilation (only useful for specific testing)
3. silence an unused variable warning in native

### DIFF
--- a/micro-rdk-server/Cargo.toml
+++ b/micro-rdk-server/Cargo.toml
@@ -18,7 +18,7 @@ ota = ["micro-rdk/ota"]
 [target.'cfg(not(target_os = "espidf"))'.dependencies]
 env_logger.workspace = true
 local-ip-address.workspace = true
-micro-rdk = { workspace = true, features = ["native"], default-features = true }
+micro-rdk = { workspace = true, features = ["native", "builtin-components", "data", "local-signaling"], default-features = false }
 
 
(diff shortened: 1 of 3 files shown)


micro-rdk/src/common/conn/viam.rs
```diff
@@ -418,11 +418,10 @@ where
     pub fn run_forever(&mut self) -> ! {
         #[cfg(feature = "esp32")]
         {
-            use esp_idf_svc::sys::{esp_task_wdt_config_t, CONFIG_FREERTOS_NUMBER_OF_CORES};
```
npmenard, line 421: don't change that for now it was handled in #457 which we reverted for now but are probably going to merge back in

micro-rdk-server/Cargo.toml
```diff
@@ -18,7 +18,7 @@ ota = ["micro-rdk/ota"]
 [target.'cfg(not(target_os = "espidf"))'.dependencies]
 env_logger.workspace = true
 local-ip-address.workspace = true
-micro-rdk = { workspace = true, features = ["native"], default-features = true }
+micro-rdk = { workspace = true, features = ["native", "builtin-components", "data", "local-signaling"], default-features = false }
```
npmenard, line 21: make sense with me

npmenard (approved):
LGTM

PR #465, opened 2025-04-04
[RSDK-10384] Add initial riscv board support above ESP-IDF 5

acmorrow:


Makefile
```diff
@@ -85,6 +85,20 @@ build-esp32-bin:
 		--merge \
 		target/xtensa-esp32-espidf/release/micro-rdk-server-esp32.bin
 
+build-esp32c6-bin:
```
acmorrow, line 88: I don't love how this Makefile is evolving but I don't have a better answer just yet. Thinking about it.

Makefile
```diff
@@ -96,8 +110,32 @@ build-esp32-ota:
...
+		--bin=micro-rdk-server-esp32 \
+		--partition-table=micro-rdk-server/esp32/partitions.csv \
+		--target=riscv32imac-esp-espidf \
+		-Zbuild-std=std,panic_abort \
+		--release \
+		target/riscv32imac-esp-espidf/release/micro-rdk-server-esp32-ota.bin
+
 build-qemu-bin:
-	cargo +esp espflash save-image --release --package micro-rdk-server --features qemu --merge --chip esp32 target/xtensa-esp32-espidf/release/qemu.bin -T micro-rdk-server/esp32/partitions.csv -s 8mb  --bin micro-rdk-server-esp32 --target=xtensa-esp32-espidf -Zbuild-std=std,panic_abort
+	cargo +esp espflash save-image \
```
acmorrow, line 127: Drive-by: I linewrapped these, re-ordered them to be consistent with the other `build-*-bin` targets, and normalized flags to the long form.

Makefile
```diff
@@ -85,6 +85,20 @@ build-esp32-bin:
 		--merge \
 		target/xtensa-esp32-espidf/release/micro-rdk-server-esp32.bin
 
+build-esp32c6-bin:
+	cargo +nightly espflash save-image \
```
acmorrow, line 89: Note: this decouples us from a pinned rust version for using the non-espressif rust toolchain for riscv compiles. I'm not sure I love that, open to suggestions. We could also use the `+esp` toolchain for this, but I was sort of enthusiastic about leaving that behind.

micro-rdk/build.rs
```diff
@@ -2,6 +2,12 @@ use regex::Regex;
 
 fn main() {
     println!("cargo::rustc-check-cfg=cfg(esp32)");
+    println!("cargo::rustc-check-cfg=cfg(esp32s2)");
```
acmorrow, line 5: [q] - Does this belong inside the `if Regex...` block below?

micro-rdk/src/esp32/i2c.rs
```diff
@@ -4,9 +4,12 @@ use crate::common::config::{AttributeError, Kind};
 use crate::common::i2c::{I2CErrors, I2CHandle};
 use crate::esp32::esp_idf_svc::hal::delay::BLOCK;
 use crate::esp32::esp_idf_svc::hal::gpio::AnyIOPin;
-use crate::esp32::esp_idf_svc::hal::i2c::{I2cConfig, I2cDriver, I2C0, I2C1};
+use crate::esp32::esp_idf_svc::hal::i2c::{I2cConfig, I2cDriver, I2C0};
 use crate::esp32::esp_idf_svc::hal::units::Hertz;
 
+#[cfg(not(any(esp32c3, esp32c2, esp32c6)))]
```
acmorrow, line 10: Until we have board support in our architecture, this is how we will need to do it.

micro-rdk/src/esp32/tcp.rs
```diff
@@ -72,7 +72,8 @@ struct Esp32ServerConfig {
 
 impl Esp32ServerConfig {
     fn new(srv_cert: &[u8], srv_key: &[u8]) -> Self {
-        let mut alpn_proto: Vec<_> = vec![ALPN_PROTOCOLS.as_ptr() as *const i8, std::ptr::null()];
+        let mut alpn_proto: Vec<_> =
+            vec![ALPN_PROTOCOLS.as_ptr() as *const c_char, std::ptr::null()];
```
acmorrow, line 76: I think this is the right fix. On rust 1.83 this wants i8 on xtensa. On rust 1.85 it wants u8 for xtensa. This is due to a change in how they are handling c_char. And I think for riscv on 1.85 it wants something else again.

acmorrow:
@npmenard - The new targets build for me, but I haven't yet been able to get the flashed software to run. Hopefully early next week we can take a look. It isn't even getting off the ground - it doesn't seem to find the right magic number in the app image header and it just boot loops from there. Almost certainly something I'm doing wrong flashing it.

acmorrow:
```
ESP-ROM:esp32c6-20220919
Build:Sep 19 2022
rst:0x3 (LP_SW_HPSYS),boot:0xc (SPI_FAST_FLASH_BOOT)
Saved PC:0x4001974a
SPIWP:0xee
mode:DIO, clock div:2
load:0x40875720,len:0x16b8
load:0x4086c110,len:0xe84
load:0x4086e610,len:0x304c
entry 0x4086c11a
I (24) boot: ESP-IDF v5.4 2nd stage bootloader
I (24) boot: compile time Apr  3 2025 16:41:48
I (25) boot: chip revision: v0.0
I (25) boot: efuse block revision: v0.1
I (27) boot.esp32c6: SPI Speed      : 80MHz
I (31) boot.esp32c6: SPI Mode       : DIO
I (35) boot.esp32c6: SPI Flash Size : 8MB
I (39) boot: Enabling RNG early entropy source...
I (43) boot: Partition Table:
I (46) boot: ## Label            Usage          Type ST Offset   Length
I (52) boot:  0 phy_init         RF data          01 01 00009000 00001000
I (59) boot:  1 otadata          OTA data         01 00 0000a000 00002000
I (65) boot:  2 nvs              WiFi data        01 02 0000c000 00014000
I (72) boot:  3 ota_0            OTA app          00 10 00020000 003e0000
I (78) boot:  4 ota_1            OTA app          00 11 00400000 003e0000
I (85) boot:  5 coredump         Unknown data     01 03 007e0000 00020000
I (91) boot: End of partition table
I (95) boot: No factory image, trying OTA 0
E (98) esp_image: Failed to fetch app description header!
E (104) boot: OTA app partition slot 0 is not bootable
E (108) esp_image: image at 0x400000 has invalid magic byte (nothing flashed here?)
E (116) boot: OTA app partition slot 1 is not bootable
E (121) boot: No bootable app partitions in the partition table
```

micro-rdk/build.rs
npmenard: probably

micro-rdk/src/esp32/tcp.rs
npmenard: alpn_protos are already stored as c_char so you should be able to just call as_ptr and avoid the casting

Makefile
npmenard: do we have to use nightly channel here?

acmorrow: Yes, per https://docs.esp-rs.org/book/installation/riscv.html, for `std`, we need the nightly channel.

micro-rdk/build.rs
acmorrow: Moved it

micro-rdk/src/esp32/tcp.rs
acmorrow: Discussed in person, we will revisit once we are rust 1.85 or newer everywhere

PR #468, opened 2025-04-07
native clippy fixes

mattjperez:
to address all the clippy warnings to get CI green

### DIFF
--- a/micro-rdk/src/native/certificate.rs
+++ b/micro-rdk/src/native/certificate.rs
@@ -26,7 +26,7 @@ impl WebRtcCertificate {
         let cert = rcgen::Certificate::from_params(param).unwrap();
         let cert_der = cert.serialize_der().unwrap();
 
-        let fp_hashed = Sha256::new_with_prefix(&cert_der)
+        let fp_hashed = Sha256::new_with_prefix(cert_der.clone())
             .finalize()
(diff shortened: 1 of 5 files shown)


micro-rdk/src/native/certificate.rs
```diff
@@ -26,7 +26,7 @@ impl WebRtcCertificate {
         let cert = rcgen::Certificate::from_params(param).unwrap();
         let cert_der = cert.serialize_der().unwrap();
 
-        let fp_hashed = Sha256::new_with_prefix(&cert_der)
```
mattjperez, line 29: this one I'm not sure about. might be better to ignore the warning instead of cloning.

micro-rdk/src/native/certificate.rs
```diff
@@ -26,7 +26,7 @@ impl WebRtcCertificate {
...
+        let fp_hashed = Sha256::new_with_prefix(cert_der.clone())
```
mattjperez, line 29: this one I'm not sure about. might be better to ignore the warning instead of cloning.

acmorrow: Can you try `.into` here?

acmorrow (approved):
LGTM

PR #470, opened 2025-04-07
[RSDK-10360] Rust 1.85 again

acmorrow:
@npmenard - Only the second commit in this review is interesting, where we are patching in your `concurrent-queue`. The first is just a revert of the revert and associated conflict resolutions. I was able to flash this build and confirmed that it did not have immediate crashes like the prior rust 1.85 build did.

Note that it seems to me that applying the patch in the workspace `Cargo.toml` worked just fine. In fact, when I put it in the `Cargo.toml` for `micro-rdk-server` it told me it was being ignored. Happy to show you in person if you would like to see.


npmenard (approved):
LGTM

PR #474, opened 2025-04-08
Fix publish.yml

mattjperez:
the publish jobs are failing. without it, we can't generate the release binaries


acmorrow (approved):
LGTM, and thanks for fixing

PR #475, opened 2025-04-09
[RSDK-10405] Publish after test

npmenard:


acmorrow (approved):
LGTM

.github/workflows/publish.yml
```diff
@@ -1,10 +1,7 @@
 name: publish
 
 on:
-  workflow_dispatch:
```
mattjperez, line 4: [optional] consider leaving this in if it wasn't causing issues. This allowed us to test PRs to the publish workflow. With it, you can go to workflow in the Actions tab and run the workflow on the PR.

mattjperez (approved):
LGTM with a note

.github/workflows/publish.yml
```diff
@@ -1,10 +1,7 @@
 name: publish
```
mattjperez, line 1: also, while you're here, could you make this `Publish`. Not sure what changed, but this name isn't being rendered properly like the other workflows in the Actions tab. the only difference being the capitalization

PR #480, opened 2025-04-14
[RSDK-10470] add Button component

mattjperez:


micro-rdk/src/common/button.rs
```diff
@@ -0,0 +1,84 @@
...
+
+#[derive(Debug, Error)]
+pub enum ButtonError {
+    #[error("failed to press button")]
+    FailedPress,
+    #[error("test error")]
+    TestError,
+}
+
+pub trait Button: Status + DoCommand {
```
gvaradarajan, line 23: can you also have this require `Send`, per the explanation [here](https://viam.atlassian.net/browse/RSDK-9955)?

micro-rdk/src/common/button.rs
```diff
@@ -0,0 +1,84 @@
...
+use thiserror::Error;
+
+pub type ButtonType = Arc<Mutex<dyn Button>>;
+
+#[derive(Debug, Error)]
+pub enum ButtonError {
+    #[error("failed to press button")]
+    FailedPress,
+    #[error("test error")]
+    TestError,
```
gvaradarajan, line 20: Maybe `OtherButtonError(#[from] Box<dyn std::error::Error + Send + Sync>)`? (see `OtherBoardError` for an example) Then we can define a different error for `FakeButton` and set the precedent for future implementations/modules

gvaradarajan (commented):
Also, you create the `FakeButton` but never register the model

micro-rdk/src/common/registry.rs
```diff
@@ -202,6 +209,18 @@ impl ComponentRegistry {
...
+    pub fn register_button(
+        &mut self,
+        model: impl Into<String>,
+        constructor: &'static ButtonConstructor,
+    ) -> Result<(), RegistryError> {
+        let model = model.into();
+        if self.buttons.contains_key(&model) {
+            return Err(RegistryError::ModelAlreadyRegistered(model));
+        }
+        let _ = self.buttons.insert(model, constructor);
```
mattjperez, line 221: this is the same as everything else, but not sure why we ignore the result in the first place

micro-rdk/src/common/button.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+        Ok(Arc::new(Mutex::new(Self::new())))
+    }
+}
+
+#[cfg(feature = "builtin-components")]
+impl Button for FakeButton {
+    fn push(&mut self) -> Result<(), ButtonError> {
+        self.count += 1;
+        log::info!("count: {}", self.count);
```
mattjperez, line 85: added this for testing, but maybe we want to leave it in?
there's no other way to tell that the button was pushed from the UI in app, and the `Status` where I report the count doesn't seem to be accessible from app's UI

micro-rdk/src/common/grpc.rs
```diff
@@ -796,6 +798,36 @@ impl<'a> GrpcServerInner<'a> {
...
 
+    fn button_push(&mut self, message: &[u8]) -> Result<Bytes, ServerError> {
+        let req = component::button::v1::PushRequest::decode(message)
+            .map_err(|_| ServerError::from(GrpcError::RpcInvalidArgument))?;
+        let button = match self.robot.lock().unwrap().get_button_by_name(req.name) {
+            Some(b) => b,
+            None => return Err(ServerError::from(GrpcError::RpcUnavailable)),
+        };
+
+        // TODO map to grpc server error
```
mattjperez, line 809: will do before merging

micro-rdk/src/common/button.rs
```diff
@@ -0,0 +1,111 @@
...
+pub type ButtonType = Arc<Mutex<dyn Button>>;
+
+#[derive(Debug, Error)]
+pub enum ButtonError {
+    #[error("failed to press button")]
+    FailedPress,
+    #[error("test error")]
+    TestError,
+    #[error(transparent)]
+    Other(#[from] Box<dyn std::error::Error + Send + Sync>),
```
npmenard, line 29: catch all error are useful for wrapping errors only defined in a target (like EspError) do we expect button to be build using esp (or other) specific implementation?

micro-rdk/src/common/button.rs
```diff
@@ -0,0 +1,111 @@
...
+        }
+        Ok(Arc::new(Mutex::new(Self::new())))
+    }
+}
+
+#[cfg(feature = "builtin-components")]
+impl Button for FakeButton {
+    fn push(&mut self) -> Result<(), ButtonError> {
+        self.count += 1;
+        log::info!("count: {}", self.count);
```
npmenard, line 90: push count

micro-rdk/src/common/button.rs
```diff
@@ -0,0 +1,111 @@
...
+        Ok(())
+    }
+}
+
+#[cfg(feature = "builtin-components")]
+impl Status for FakeButton {
+    fn get_status(&self) -> Result<Option<google::protobuf::Struct>, StatusError> {
+        let mut hm = HashMap::new();
```
npmenard, line 98: status was deprecated no need to populate it. we will be removing it

micro-rdk/src/common/grpc.rs
```diff
@@ -796,6 +798,36 @@ impl<'a> GrpcServerInner<'a> {
...
+        button.lock().unwrap().push().unwrap();
```
npmenard, line 810: no unwrap

micro-rdk/src/common/registry.rs
npmenard, line 221: it returns an option if there was a value associated with that key above. we check if there is already a value above.

npmenard (commented):
done with first pass

micro-rdk/src/common/button.rs
```diff
@@ -0,0 +1,111 @@
...
+
+pub static COMPONENT_NAME: &str = "button";
+
+pub type ButtonType = Arc<Mutex<dyn Button>>;
+
+#[derive(Debug, Error)]
+pub enum ButtonError {
+    #[error("failed to press button")]
+    FailedPress,
+    #[error("test error")]
```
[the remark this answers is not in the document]
mattjperez: L79, in `from_config`, there's an attribute you can set to see what happens when the component fails to initialize. I saw it in [`FakeBoard`](https://github.com/viamrobotics/micro-rdk/blob/9ca4bec2810a79a1a1e3af46330eeb7b2a59682c/micro-rdk/src/common/board.rs#L137), which I referenced at some point.

micro-rdk/src/common/button.rs
```diff
@@ -0,0 +1,111 @@
...
+};
+
+pub static COMPONENT_NAME: &str = "button";
+
+pub type ButtonType = Arc<Mutex<dyn Button>>;
+
+#[derive(Debug, Error)]
+pub enum ButtonError {
+    #[error("failed to press button")]
+    FailedPress,
```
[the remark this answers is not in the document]
mattjperez: ah, I was planning on adding another test error using an attribute so the pressing fails, not the component initialization. can remove it.

micro-rdk/src/common/button.rs
mattjperez: maybe? what's the alternative for a module writer? should this just be a string instead?

what errors do you suggest we have?

micro-rdk/src/common/robot.rs
```diff
@@ -612,6 +623,7 @@ impl LocalRobot {
                             status,
                         });
                     }
+                    _ => continue
```
mattjperez, line 626: to cover the `ResourceType::Button` case since it doesn't `impl Status` anymore. This will also help for the `Switch` PR.

micro-rdk/src/common/robot.rs
```diff
@@ -703,6 +715,7 @@ impl LocalRobot {
                                 status,
                             });
                         }
+                        _ => continue
```
mattjperez, line 718: same as above

micro-rdk/src/common/button.rs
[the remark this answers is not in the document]
npmenard: then we should rename it to something that doesn't look like we push Test  in prod :P 
And we should avoid code that is not useful unless it is associated with a test

micro-rdk/src/common/button.rs
npmenard: i guess we don't have a good idea of what failure mode of a button looks like or even how we would use it so let's leave it as is

micro-rdk/src/common/grpc.rs
```diff
@@ -796,6 +798,39 @@ impl<'a> GrpcServerInner<'a> {
...
+            .map_err(|_| ServerError::from(GrpcError::RpcInvalidArgument))?;
+        let button = match self.robot.lock().unwrap().get_button_by_name(req.name) {
+            Some(m) => m,
+            None => return Err(ServerError::from(GrpcError::RpcUnavailable)),
+        };
+        let res = button
+            .lock()
+            .unwrap()
+            .do_command(req.command)
+            .map_err(|_| ServerError::from(GrpcError::RpcInvalidArgument))?;
```
npmenard, line 829: Wrap the error that is returned

npmenard (approved):
LGTM with comments

PR #481, opened 2025-04-15
use flash size argument for OTA targets

acmorrow:
Future versions of `cargo-espflash` (I'm running a build from HEAD that has fixes needed to flash an esp36c6) demand this flag, or error out with:

```
Partition table:   micro-rdk-server/esp32/partitions.csv
Error: espflash::partition_table::does_not_fit

  × The partition table does not fit into the flash (4MB)
  help: Make sure you set the correct flash size with the `--flash-size`
        option

make: *** [build-esp32-ota] Error 1
```

Add this flag now so things don't break when we upgrade `cargo-espflash` to a version that introduces this requirement.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,6 +109,7 @@ build-esp32-ota:
 		--target=xtensa-esp32-espidf \
 		-Zbuild-std=std,panic_abort \
 		--release \
+		--flash-size=8mb \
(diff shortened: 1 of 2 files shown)


mattjperez (approved):
LGTM

PR #482, opened 2025-04-15
[CONSULT-1023] - support polymorphic NMEA fields

gvaradarajan:
Polymorphic fields are what we call canboat's [Lookup with FieldType enumerations](https://canboat.github.io/canboat/canboat.html#lookupfieldtype-enumerations). The example implementation in this PR is for [SIMNET_KEY_VALUE](https://canboat.github.io/canboat/canboat.html#lookup-SIMNET_KEY_VALUE).

### DIFF
--- a/micro-rdk-nmea-macros/src/attributes.rs
+++ b/micro-rdk-nmea-macros/src/attributes.rs
@@ -229,7 +229,9 @@ impl MacroAttributes {
                             Meta::NameValue(named) => {
                                 if let Expr::Lit(ref expr_lit) = named.value {
                                     if let Lit::Str(lit_str) = &expr_lit.lit {
-                                        lit_str.parse().map_err(|_| error_tokens("uh oh"))?
+                                        lit_str
+                                            .parse()
+                                            .map_err(|_| error_tokens("length_field unparseable"))?
                                     } else {
                                         return Err(error_tokens(
                                             "length_field parameter must be string",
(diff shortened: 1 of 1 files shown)


micro-rdk-nmea-macros/src/attributes.rs
```diff
@@ -252,6 +256,38 @@ impl MacroAttributes {
                                 .clone(),
                         );
                     }
+                    "lookup_field" => {
+                        let lookup_field_path: syn::Path = match &attr.meta {
+                            Meta::NameValue(named) => {
+                                if let Expr::Lit(ref expr_lit) = named.value {
+                                    if let Lit::Str(lit_str) = &expr_lit.lit {
+                                        lit_str.parse().map_err(|_| error_tokens("uh oh"))?
```
npmenard, line 264: can't it be a bit more verbose that "uh oh"? :P

micro-rdk-nmea/src/parse_helpers/parsers.rs
```diff
@@ -109,6 +124,20 @@ macro_rules! generate_number_field_readers {
                     }
                 }
             }
+
+            impl FieldReader for NumberFieldWithScale<$t> {
+                type FieldType = f64;
+                fn read_from_cursor(&self, cursor: &mut DataCursor) -> Result<Self::FieldType, NmeaParseError> {
+                    self.number_field.read_from_cursor(cursor).and_then(|x| {
+                        match x {
+                            x if x == <$t>::MAX => { return Err(NumberFieldError::FieldNotPresent("type unavailable".to_string()).into()); },
```
npmenard, line 133: do we still want to return a ReadingResponse if the field is absent or is it fine to return a error and no readings even if other values are present?

npmenard (approved):
LGTM with a minor comment

micro-rdk-nmea/src/parse_helpers/parsers.rs
gvaradarajan: I believe last I checked, gonmea's behavior was to error completely

PR #484, opened 2025-04-16
[RSDK-10466] map pins from builder-initialized incremental encoder config

mattjperez:


micro-rdk/src/esp32/encoder.rs
```diff
@@ -57,6 +57,43 @@ pub struct Esp32Encoder<A, B> {
     b: B,
 }
 
+pub struct Esp32EncoderConfig {
+    pub(crate) a: Option<i32>,
```
npmenard, line 61: why do we use an Option here?

micro-rdk/src/esp32/encoder.rs
```diff
@@ -94,9 +131,16 @@ where
...
         _: Vec<Dependency>,
     ) -> Result<EncoderType, EncoderError> {
-        let pin_a_num = cfg.get_attribute::<i32>("a")?;
+        let pins = cfg
+            .get_attribute::<Esp32EncoderConfig>("pins")
+            .map_err(EncoderError::EncoderConfigAttributeError)?;
 
-        let pin_b_num = cfg.get_attribute::<i32>("b")?;
+        let pin_a_num = pins.a.ok_or(EncoderError::EncoderConfigurationError(
+            "no 'a' pin specified".to_string(),
```
npmenard, line 139: before we would let the config parsing report the different errors (usually), why do we have a new error type?

micro-rdk/src/esp32/encoder.rs
```diff
@@ -57,6 +57,43 @@ pub struct Esp32Encoder<A, B> {
...
+                AttributeError::KeyNotFound(_) => None,
+                _ => {
+                    return Err(err);
+                }
+            },
+        };
+        let b = match value.get("b") {
+            Ok(opt) => match opt {
+                Some(val) => Some(val.try_into()?),
+                None => None,
```
gvaradarajan, line 84: I think, instead of having `a` and `b` be Options, this case should return an `AttributeError::ValidationError`

micro-rdk/src/common/encoder.rs
```diff
@@ -31,6 +31,8 @@ pub enum EncoderError {
     EncoderConfigAttributeError(#[from] AttributeError),
     #[error("encoder error code: {0}")]
     EncoderCodeError(i32),
+    #[error("configuration error: {0}")]
```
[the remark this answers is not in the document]
mattjperez: ah, this was leftover from a previous iteration. thanks for catching

acmorrow (approved):
LGTM

micro-rdk/src/esp32/encoder.rs
```diff
@@ -57,6 +57,32 @@ pub struct Esp32Encoder<A, B> {
     b: B,
 }
 
+pub struct Esp32EncoderConfig {
```
gvaradarajan, line 60: should the struct just be `pub(crate)`, since all of its members are?

gvaradarajan (approved):
LGTM, with a small comment

micro-rdk/src/esp32/encoder.rs
```diff
@@ -57,6 +57,32 @@ pub struct Esp32Encoder<A, B> {
...
+                "`{:?}` is an invalid value for pin `a",
+                a
+            )))??;
+        let b = value.get("b")?;
+        let b = b
+            .map(|v| v.try_into())
+            .ok_or(AttributeError::ValidationError(format!(
+                "`{:?}` is an invalid value for pin `b`",
+                b
+            )))??;
```
npmenard, line 81: just a question when using ?? doesn't that potentially silence the intermediate error if two are chained?

micro-rdk/src/esp32/encoder.rs
npmenard, line 60: i don't think it should be pub at all unless we are using it elsewhere

micro-rdk/src/esp32/encoder.rs
acmorrow: I missed this first time through: this is happening because of the call to `ok_or` on an `Option`, which ends up creating a `Result<Result`

micro-rdk/src/esp32/encoder.rs
```diff
@@ -57,6 +57,32 @@ pub struct Esp32Encoder<A, B> {
...
+}
+
+impl TryFrom<&Kind> for Esp32EncoderConfig {
+    type Error = AttributeError;
+    fn try_from(value: &Kind) -> Result<Self, Self::Error> {
+        let a = value.get("a")?;
+        let a = a
+            .map(|v| v.try_into())
+            .ok_or(AttributeError::ValidationError(format!(
+                "`{:?}` is an invalid value for pin `a",
```
acmorrow, line 72: Missing a backtick here after `a`

acmorrow (approved):
LGTM

micro-rdk/src/esp32/encoder.rs
mattjperez: Note: while verifying output for different approaches to this, found that we don't actually propagate useful, descriptive errors to the user. made a [ticket](https://viam.atlassian.net/browse/RSDK-10534?atlOrigin=eyJpIjoiODdmNjk2YjhkMDg1NDNhMTgyYjlkYjhlMjVmNDNlM2MiLCJwIjoiaiJ9)

PR #476, opened 2025-04-10
[RSDK-9148] Revert "Fix boot loop when upgrading a device to 0.3.1 (#343)"

npmenard:
This reverts commit 0aca25f529543d5161a40cd26790687b13be1a8c.

PR #485, opened 2025-04-16
[RSDK-10492] Apply missed updates to project template

acmorrow:


templates/project/Cargo.toml.liquid
```diff
@@ -28,15 +28,11 @@ features = [
 ]
 
 [dependencies]
-async-channel = "2"
-embedded-hal = { version = "~0.2", features = ["unproven"]}
```
acmorrow, line 32: Originally, I was going to update these to match what was in the top-level Micro-RDK repository `Cargo.toml`, but then I realized that generated projects don't actually have any direct dependencies on these crates anymore, so I've just removed them.

templates/project/Cargo.toml.liquid
```diff
@@ -28,15 +28,11 @@ features = [
...
-embedded-svc = "~0.28"
-futures-lite = "2"
 log = "0.4"
 
 [build-dependencies]
 cargo_metadata = "0.19"
-embuild = "0.32"
```
acmorrow, line 39: This however was different from what was in `Micro-RDK's` top-level `Cargo.toml` so I have updated it.

templates/project/src/main.rs
```diff
@@ -90,7 +90,7 @@ fn main() {
                     .expect("failed to parse app address")
                     .into(),
                 )
-                .expect("failed to store robot credentials to NVS");
```
acmorrow, line 93: I believe this regressed as part of #476

templates/project/src/main.rs
```diff
@@ -74,7 +74,7 @@ fn main() {
     }
 
     if !storage.has_robot_credentials() {
-        log::warn!("no machine configuration was found in storage");
```
acmorrow, line 77: Both `esp32/main.rs` and `native/main.rs` have this as "machine credentials were". This is why I want to move this code out of the template.

mattjperez (approved):
LGTM,

will be nice once the template gets trimmed down to just a builder

PR #487, opened 2025-04-17
[CONSULT-1023] support custom PGN formats used for proprietary messages

gvaradarajan:
Manufacturers with specialized NMEA messages have to share a range of PGNs for defining their formats. As a result, there are different message formats with the same PGN that match on the value of a certain set of fields in the data past the appearance of the PGN. This PR treats all of the variants as a single PGN message enum with each variant being a struct that implements the Fieldset trait.

### DIFF
--- a/micro-rdk-nmea/src/messages/message.rs
+++ b/micro-rdk-nmea/src/messages/message.rs
@@ -6,7 +6,10 @@ use micro_rdk::{
     google::protobuf::{value::Kind, Value},
 };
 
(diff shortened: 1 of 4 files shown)


gvaradarajan:
@npmenard and @acmorrow Could I get a review on the formatting of the macro before I delete the example (because it's a proprietary PGN)?

npmenard (approved):
LGTM

acmorrow (approved):
LGTM

PR #488, opened 2025-04-17
[CONSULT-1024] generate NMEA lookups from JSON definitions

gvaradarajan:
Just generates the enums from NMEA lookups defined in JSON

First of a few PRs for generally importing all the PGN definitions from a JSON file. Next up will be polynomial fields, then standard PGNs, then proprietary PGNs, and then logic to download the JSON file from a remote location. 

Until then, if testing is desired, ask @gvaradarajan for a definitions.json file.


micro-rdk-nmea/src/gen/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod enums;
```
mattjperez, line 1: should this file also be auto-generated?
If it's here but the enums.rs generation get skipped because there was no json, does this cause issues?

micro-rdk-nmea/build.rs
```diff
@@ -0,0 +1,142 @@
...
+    }
+}
+
+fn main() {
+    println!("cargo:rerun-if-changed=definitions.json");
+    println!("cargo::rustc-check-cfg=cfg(autogen_definitions)");
+
+    let enums_path = "src/gen/enums_gen.rs";
+
+    let mut enums_file = OpenOptions::new()
```
npmenard, line 50: can we generate the file in the target dir?

micro-rdk-nmea/build.rs
```diff
@@ -0,0 +1,142 @@
...
+        .open(enums_path)
+        .expect("could not create new enums_gen.rs");
+
+    enums_file
+        .write_all("// AUTO-GENERATED CODE; DO NOT DELETE OR EDIT\n".as_bytes())
+        .expect("failed to write warning statement");
+
+    let file_path = "definitions.json";
+    if !Path::new(file_path).exists() {
+        println!("No definitions file, skipping auto-generation...");
```
npmenard, line 64: what happen to the compile step when no definition are supplied? is the nmea driver even usable?

npmenard (commented):
done with first pass

micro-rdk-nmea/build.rs
gvaradarajan: Can you import from files in the target directory as if they were modules?

micro-rdk-nmea/build.rs
gvaradarajan: The files are empty and the viamboat module doesn't even exist. So no it is not usable if compiled without definitions

npmenard: maybe we should not allow building if json is not present, if feature gating the nmea module isn't to complex

micro-rdk-nmea/build.rs
gvaradarajan: I figured out what to do here, I'll make the change

acmorrow (commented):
Mostly questions. I have some unease about the effects this will have on builds, and some uncertainty about how builds where generation is enabled will become consumable products, but we can discuss in person.

micro-rdk-nmea/build.rs
gvaradarajan: the reason I didn't go in that direction is that I want to be able to run tests on the  individual data-type parsers

micro-rdk-nmea/src/lib.rs
```diff
@@ -1,32 +1,38 @@
...
 pub mod viamboat;
 
+#[allow(unused_variables)]
 pub fn register_models(registry: &mut ComponentRegistry) -> Result<(), RegistryError> {
-    viamboat::register_models(registry)
+    #[cfg(autogen_definitions)]
+    let res = viamboat::register_models(registry);
+    #[cfg(not(autogen_definitions))]
+    let res = Ok(());
+    res
```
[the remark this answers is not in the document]
gvaradarajan: I tried that and it didn't work, but I can give it another shot

micro-rdk-nmea/build.rs
```diff
@@ -0,0 +1,142 @@
...
+            )
+        };
+        format!("UnformattableVariant{}", suffix)
+    } else {
+        filtered
+    }
+}
+
+fn main() {
+    println!("cargo:rerun-if-changed=definitions.json");
```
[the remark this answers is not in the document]
gvaradarajan: No. In fact, a problem that I've found is that when there isn't such a file and then such a file is included, Cargo does not detect that rebuild is necessary

micro-rdk-nmea/src/lib.rs
```diff
@@ -1,32 +1,38 @@
...
 }
 
+#[cfg(autogen_definitions)]
```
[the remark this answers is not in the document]
gvaradarajan: the utilities are still exported (the readers and such)

micro-rdk-nmea/build.rs
```diff
@@ -0,0 +1,154 @@
...
+        .chars()
+        .position(|c| (c.is_numeric() || (c == '_')))
+        .map(|idx| idx == 0)
+        .unwrap_or_default()
+    {
+        // We want to use a prefix with an alphabetical suffix (i.e. "UnformattableVariantA")
+        // when the cleaned string still starts with a number. We map however many variants we've
+        // seen so far to an ASCII code in range 65-90 (A-Z). If there are more than 25 variants
+        // we want to use a second letter (i.e. "UnformattableVariantAB")
+        let suffix = if counter <= 25 {
```
npmenard, line 32: So we loose information about what the name  of the thing is right? or can we retrieve it latter? does it even matter?

micro-rdk-nmea/build.rs
```diff
@@ -0,0 +1,154 @@
...
+// want to remove special characters and introduce a default convention for
+// when a name starts with a number.
+fn clean_string_for_rust(input: &str, counter: usize) -> String {
+    let filtered: String = input
+        .chars()
+        .map(|c| {
+            if c.is_alphanumeric() || c.is_whitespace() {
+                c
+            } else {
+                '_'
```
npmenard, line 18: what non alphanumeric character we running into? Should we just filter them out instead of replacing?

npmenard (approved):
Left a couple of comments but otherwise LGTM

micro-rdk-nmea/build.rs
gvaradarajan: A future PR replaces the number with its word representation instead. This will go away soon. Although I can try to bring the code over in this PR if you prefer

micro-rdk-nmea/build.rs
gvaradarajan: I found that the replacement with underscore made swapping to pascal/snake case easier in the case of a name like "@23..."

PR #492, opened 2025-04-21
Update README.md example JSON

JessamyT:


JessamyT:
Turns out this is only working by coincidence and won’t be updated until https://viam.atlassian.net/browse/RSDK-7822

PR #494, opened 2025-04-23
[RSDK-10287] Don't run license checker in the container

acmorrow:
Trying this in a PR, let's see what happens.


mattjperez:
did this work as you expected? or needs a merge to test fully?

acmorrow:
It does not work as expected, because the license checker infra isn't available outside our container. Personally, from what I see running the license checker inside the container takes about one minute, which isn't close to the longest running task in a CI iteration. So, personally, I'm not sure we should spend time optimizing this. I will post those findings to the ticket.

PR #496, opened 2025-04-28
[RSDK-10184] scan for networks before attempting

mattjperez:
`try_connect_by_priority` is being called in a while loop, so no need to re-scan within the function


micro-rdk/src/esp32/conn/network.rs
```diff
@@ -268,10 +270,31 @@ impl Esp32WifiNetwork {
...
+                .map(|net| net.0.ssid),
+        );
+
+        log::info!("establishing WiFi connection...");
+
         networks.sort();
         for network in networks.iter() {
+            if !available.contains(&network.ssid) {
+                log::info!(
+                    "`{}` network not found in initial scan, skipping...",
```
mattjperez, line 294: Will remove the word 'initial'

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -268,12 +270,31 @@ impl Esp32WifiNetwork {
...
+            .set_station_mode(NetworkSetting::new("".to_string(), "".to_string(), 0))
+            .await;
+
+        log::info!("scanning available networks");
+        let available: HashSet<String> = HashSet::from_iter(
+            self.scan_networks()
+                .await
+                .unwrap_or_default()
+                .into_iter()
+                .map(|net| net.0.ssid),
```
mattjperez, line 285: we could do a filter before mapping to only consider networks above a certain signal strength

acmorrow: I think below the [-75,-80] range is usually considered the threshold for a negative experience. But it is tricky: if we have multiple candidate networks available, should we pick the stronger one? But if we can see only one, should we still try to connect to it even if the signal is really bad? If we have a higher priority network, but it has a bad signal, should we still prefer it over a slightly lower priority network that is available with a strong signal? If we have multiple networks with the same priority, should we opt for highest signal first? What does linux do? Is there some sort of known algorithm for deciding this?

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -268,12 +270,31 @@ impl Esp32WifiNetwork {
         &self,
         mut networks: Vec<NetworkSetting>,
     ) -> Result<(), WifiManagerError> {
-        // TODO(RSDK-10184): scan available networks first
-        // attempt to connect only if available
+        log::info!("initializing wifi module");
+        // `set_station_mode` ends by calling `connect`, which will error here
+        let _err = self
```
acmorrow, line 275: Rather than this, maybe split `set_station_mode` into `connect_station_mode` which ends with a connect, and `set_station_mode`, which `connect_station_mode` calls? Then you can do `set_station_mode` here without needing to deal with an error out of the `connect` part.

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -268,12 +270,31 @@ impl Esp32WifiNetwork {
...
+            .set_station_mode(NetworkSetting::new("".to_string(), "".to_string(), 0))
+            .await;
+
+        log::info!("scanning available networks");
```
acmorrow, line 279: Do you want to log the available networks afterwards?

micro-rdk/src/esp32/conn/network.rs
mattjperez: I was debating between that or removing the `connect` from `set_station_mode` and having the caller call connect directly afterward since they will still have ownership of the struct. setting the station mode shouldn't do more than set the station anyway.

which do you prefer?

micro-rdk/src/esp32/conn/network.rs
acmorrow: No, I think I agree, having considered it more and reading your comment, that it is likely too chatty.

micro-rdk/src/esp32/conn/network.rs
acmorrow: Whichever is easier / fewer changes.

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -268,12 +270,31 @@ impl Esp32WifiNetwork {
         &self,
         mut networks: Vec<NetworkSetting>,
     ) -> Result<(), WifiManagerError> {
-        // TODO(RSDK-10184): scan available networks first
-        // attempt to connect only if available
+        log::info!("initializing wifi module");
```
acmorrow, line 273: I think this log is maybe not useful.

micro-rdk/src/esp32/conn/network.rs
mattjperez: taken from https://askubuntu.com/a/1348319
> 
    - Scan for potential networks
    - Rank the scanned networks by priority and connect to the highest-priority network
    - Rank the prioritized networks by connection strength and connect to the strongest one
    - Rank the networks by order in the configuration file and connect to the first one

micro-rdk/src/esp32/conn/network.rs
mattjperez: actually, turns out we do need to call `connect()` with the default network. it appears like some things need to be set aside from setting the station mode config itself.

playing around with it, the way it is currently appears to be the 'cleanest'. so gonna leave it as-is.
will probably change the comment though to be more accurate

micro-rdk/src/esp32/conn/network.rs
mattjperez: made a follow-up ticket [RSDK-10612](https://viam.atlassian.net/browse/RSDK-10612?atlOrigin=eyJpIjoiYmY0ZjVjNjY5ZjgwNGNkY2I1Yzk5ZDM2YzFkMTY2NGYiLCJwIjoiaiJ9)

micro-rdk/src/esp32/conn/network.rs
npmenard, line 275: I don't like the idea of silencing the error. If something else goes wrong then may be hard to track down the root cause.

The flow of attempting to connect to WiFi with an empty ssid is flawed imo. We have no guarantee this will continue to work moving forward

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -268,12 +270,31 @@ impl Esp32WifiNetwork {
...
+        log::info!("establishing WiFi connection...");
+
         networks.sort();
         for network in networks.iter() {
+            if !available.contains(&network.ssid) {
+                log::info!("network `{}` not found in scan, skipping...", network.ssid);
+                continue;
+            }
+            log::info!("attempting connection to network `{}`", network.ssid);
             if let Err(e) = self.set_station_mode(network.clone()).await {
```
npmenard, line 296: This start/stop the WiFi on each connection attempt not super ideal

npmenard (changes requested):

npmenard (approved):
LGTM

micro-rdk/src/esp32/conn/network.rs
```diff
@@ -268,12 +275,36 @@ impl Esp32WifiNetwork {
         &self,
         mut networks: Vec<NetworkSetting>,
     ) -> Result<(), WifiManagerError> {
-        // TODO(RSDK-10184): scan available networks first
-        // attempt to connect only if available
+        self.set_station_mode(NetworkSetting::new("".to_string(), "".to_string(), 0))
```
acmorrow, line 278: Is this still needed?

acmorrow (approved):
LGTM, but with a question about whether the initial `set_station_mode` is still necessary. If so, why?

micro-rdk/src/esp32/conn/network.rs
mattjperez: Will test if needed. I recall some c-based esp-IDF examples setting an empty configuration before starting. 

If not needed will remove.

mattjperez: looking a little deeper, I think I'd like to keep this. 

Looking at the EspWifi driver, the behavior of [`start`](https://docs.esp-rs.org/esp-idf-svc/esp_idf_svc/sys/fn.esp_wifi_start.html) is determined by the station mode. So the important thing isn't the empty network settings as much as it is ensuring the[ `ClientConfiguration`](https://docs.esp-rs.org/esp-idf-svc/esp_idf_svc/wifi/struct.ClientConfiguration.htmll) is set appropriately. 

will use Default instead of new here though to make it nicer looking

I haven't tested it, but it's possible it might come up in mixed/AP mode when started if we don't set it appropriately.

PR #497, opened 2025-04-28
[CONSULT-1024] - generate NMEA fieldtype lookups from JSON definitions

gvaradarajan:
Adds auto-generation of [lookups with field type enumerations](https://canboat.github.io/canboat/canboat.html#lookupfieldtype-enumerations) from a JSON definitions file


npmenard (approved):
LGTM

PR #498, opened 2025-04-29
[RSDK-7893] use ntp server

mattjperez:
PoC for using NTP to set the system clock (`settimeofday` called by SNTP service).


micro-rdk/src/common/app_client.rs
```diff
@@ -296,39 +296,6 @@ impl AppClient {
             None
         };
 
-        #[cfg(feature = "esp32")]
```
mattjperez, line 299: I'm not sure what the benefit is of having both ntp at startup (setting right after connecting) and this during `get_app_config`. 

Additionally, the sntp library handles periodic setting (seen below in sdkconfig)

micro-rdk/src/common/conn/viam.rs
```diff
@@ -523,6 +523,23 @@ where
...
+            use crate::esp32::esp_idf_svc::sys;
+            log::info!("initializing SNTP service");
+            sys::esp_sntp_setoperatingmode(sys::esp_sntp_operatingmode_t_ESP_SNTP_OPMODE_POLL);
+            let ntp_addr = std::ffi::CString::new("pool.ntp.org").expect("stsr");
+            sys::esp_sntp_setservername(0, ntp_addr.as_ptr());
+            sys::esp_sntp_init();
+
+            while sys::sntp_get_sync_status() != sys::sntp_sync_status_t_SNTP_SYNC_STATUS_COMPLETED
+            {
+                Timer::after(Duration::from_secs(1)).await;
```
mattjperez, line 538: it takes about 1 second for the ntp call to be made and the system clock to be set. 
I'm not sure what a better strategy for handling this is. I wanted to ensure the clock is set before the programs moves forward to more log-generating things

templates/project/sdkconfig.defaults
```diff
@@ -57,3 +57,6 @@ CONFIG_ESP_COREDUMP_CHECKSUM_SHA256=y
 CONFIG_ESP_COREDUMP_ENABLE=y
 CONFIG_ESP_COREDUMP_MAX_TASKS_NUM=64
 CONFIG_ESP_COREDUMP_DECODE="disable"
+
+# NTP Config
+CONFIG_LWIP_SNTP_UPDATE_DELAY=86400000
```
mattjperez, line 62: can save this for a template update PR if preferrded

micro-rdk-server/sdkconfig.defaults
```diff
@@ -41,3 +41,5 @@ CONFIG_ESP_COREDUMP_ENABLE=y
 CONFIG_ESP_COREDUMP_MAX_TASKS_NUM=64
 CONFIG_ESP_COREDUMP_DECODE="disable"
 
+# NTP Config
+CONFIG_LWIP_SNTP_UPDATE_DELAY=86400000
```
mattjperez, line 45: 1 day in ms

micro-rdk/src/common/conn/viam.rs
```diff
@@ -523,6 +523,23 @@ where
...
         }
 
+        // Initiate SNTP
+        #[cfg(feature = "esp32")]
+        unsafe {
+            use crate::esp32::esp_idf_svc::sys;
+            log::info!("initializing SNTP service");
+            sys::esp_sntp_setoperatingmode(sys::esp_sntp_operatingmode_t_ESP_SNTP_OPMODE_POLL);
+            let ntp_addr = std::ffi::CString::new("pool.ntp.org").unwrap();
+            sys::esp_sntp_setservername(0, ntp_addr.as_ptr());
```
npmenard, line 533: Why not using the esp idf svc driver?

mattjperez: Because I somehow missed the fact there was one and went straight to the c bindings 🙃.

Will rework this to use the rust driver

micro-rdk-server/sdkconfig.defaults
acmorrow: That feels much too long. The default is an hour. Why change it/

templates/project/sdkconfig.defaults
acmorrow: Same thoughts. I'm not sure why we should override the default of an hour.

micro-rdk/src/common/app_client.rs
```diff
@@ -296,39 +296,6 @@ impl AppClient {
...
-        {
-            // In many cases when the ESP32 reboots the RTC clock will not be cleared and will retain the previous
-            // time. If the reboot was caused by a WDT timeout there is a chance the clock would drift significantly
-            // The new logic is to always set the clock once per boot
-
-            if let Some(current_dt) = datetime {
-                CLOCK_SET.call_once(|| {
```
acmorrow, line 306: If this does end up being entirely removed, remember to remove the declaration of `CLOCK_SET` above.

micro-rdk/src/common/app_client.rs
acmorrow: I could see this being retained as a fallback in case ntp isn't working. If the NTP stuff set `CLOCK_SET` when the time was synced, if you got here and it still wasn't synced you could use the app time.

micro-rdk/src/common/conn/viam.rs
```diff
@@ -523,6 +523,23 @@ where
...
+            sys::esp_sntp_init();
+
+            while sys::sntp_get_sync_status() != sys::sntp_sync_status_t_SNTP_SYNC_STATUS_COMPLETED
```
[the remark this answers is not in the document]
mattjperez: using the rust driver (esp when combined with the original way from the config) makes this step unnecessary

micro-rdk/src/common/conn/viam.rs
mattjperez: reworked

micro-rdk/src/common/conn/viam.rs
```diff
@@ -523,6 +523,22 @@ where
             {}
         }
 
+        let _sntp: Box<_>;
```
mattjperez, line 526: I'm a bit reluctant to add more to the ViamServer struct, though it might be cleaner in the end. That or have it as a conditional global.  How it is currently would need to be `Box<dyn Any>` to pass native CI though.

micro-rdk-server/sdkconfig.defaults
npmenard: there shouldn't be much drift at run time. It's mostly during deep sleep that the clock we incur a lot of errors.

npmenard: But I wonder if when the wifi we are connected to doesn't have internet if failing to sync time will schedule the next sync to +24h

micro-rdk/src/common/app_client.rs
```diff
@@ -321,7 +321,7 @@ impl AppClient {
                     })
                     .inspect(|_| {
                         log::info!(
-                            "time of day has been set to to {}",
+                            "time of day has been set by config metadata to {}",
```
npmenard, line 324: unless there is a strong need there is no point of keep this path for time sync. The question is , when internet is restored is the time sync mostly instantaneous (less than a few seconds)

micro-rdk-server/sdkconfig.defaults
mattjperez: then having the default 1hr would be better, no? Then the gap would be smaller in this case.

Another thing I'm not sure about is, if the device is waking only for brief periods of time, will the time sync get called during that small timeframe.

acmorrow (approved):
LGTM

mattjperez:
I tested the SNTP behavior by setting the interval to one minute and the sync callback to log each sync (without using `Once`), and disabling the app_config.

There are two scenarios:
- a device comes up without internet - as `ViamServer` loops trying to get internet on startup and sntp is directly after it, if sntp will never be reached or set until internet is established, in which case the sntp server will be initialized, syncing immediately.
- a device comes up with internet, sync time with sntp server, then loses connection but does not reboot - I confirmed that the delay is still adhered to and **does not sync immediately after connection is established**. Additionally, because the `Once` was already used shortly after startup, the `get_app_config` will never happen again before a reboot, even when connection is re-established.

From this, it seems like we have two options:
1. lean on sntp completely and set the sync interval to be that of the max amount of time passed after connection is established to resync (15 seconds smallest configurable).
2. instead of using `Once`, use a `TIME_OF_LAST_SYNC` that we reset at each successful sync and use the app_config when it goes above a certain threshold, restarting it. We can maybe use [this](https://docs.esp-rs.org/esp-idf-svc/esp_idf_svc/hal/sys/fn.sntp_get_sync_interval.html) to determine what the delay interval is at runtime.

npmenard:
The drift is not significant over short period of times. What matters here is  1) time sync as soon as internet can be reached. 2) Sync on a period accepting missing sync pulse.
If 24h is too feels too high we can reduce it. But I feel it's somewhat appropriate.

If we want the system to sync opportunistically meaning we have a use case where the device is always on but only has access to internet on a short period of time then we will need other logic anyway.

mattjperez:
made a follow-up ticket for opportunistically syncing on re-connection
[RSDK-10655](https://viam.atlassian.net/browse/RSDK-10655?atlOrigin=eyJpIjoiZTYzZWNkMTc2MmE2NDIyOTg1NWNjODllZjM2ZTNkYmIiLCJwIjoiaiJ9)

PR #499, opened 2025-05-01
[RSDK-7822] Use API & Model triplet

npmenard:
This covers components only. I am debating on updating services too

### DIFF
--- a/micro-rdk/src/common/config.rs
+++ b/micro-rdk/src/common/config.rs
@@ -772,8 +772,91 @@ mod tests {
     use std::collections::HashMap;
 
     use crate::common::config::{
-        AttributeError, Component, DynamicComponentConfig, Kind, Model, ResourceName,
+        AttributeError, Component, ConfigModelError, DynamicComponentConfig, Kind, Model,
+        ResourceName,
     };
+
+    use super::{ConfigApiError, API};
+
+    #[test_log::test]
+    fn test_model_api() {
+        let model1: Result<Model, ConfigModelError> = "rdk:builtin:sensor".try_into();
+        assert!(model1.is_ok());
+
+        let model1: Result<Model, ConfigModelError> = "rdk:builtin_b:sensor".try_into();
+        assert!(model1.is_ok());
+
+        let model1: Result<Model, ConfigModelError> = "rdk:builtin:sensor-d".try_into();
+        assert!(model1.is_ok());
+
+        let model1: Result<Model, ConfigModelError> = "rdk:builtin:movement_sensor".try_into();
+        assert!(model1.is_ok());
+
+        let model1: Result<Model, ConfigModelError> = "rdk::sensor".try_into();
+        assert!(model1.is_err());
+        assert!(matches!(
+            model1.unwrap_err(),
+            ConfigModelError::ConfigModelErrorInvalidFamilly(_)
+        ));
+
+        let model1: Result<Model, ConfigModelError> = "rdk:ok:sensor+g".try_into();
+        assert!(model1.is_err());
+        assert!(matches!(
+            model1.unwrap_err(),
+            ConfigModelError::ConfigModelErrorInvalidModel(_)
+        ));
+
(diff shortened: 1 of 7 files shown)


micro-rdk/src/common/config.rs
```diff
@@ -270,17 +299,225 @@ impl TryFrom<&google::protobuf::value::Kind> for Kind {
...
 }
 
-#[derive(Debug, Default)]
+pub struct Model {
+    familly: String,
+    model: String,
+    namespace: String,
+}
+
+impl Model {
```
npmenard, line 308: added that change too so we are closer to parity with RDK

micro-rdk/src/common/config.rs
```diff
@@ -270,17 +299,225 @@ impl TryFrom<&google::protobuf::value::Kind> for Kind {
...
+}
+// this is equivalent to the regex \w-
+fn is_valid_api_model_char(c: char) -> bool {
+    c.is_alphanumeric() || c == '_' || c == '-'
+}
+
+// for both model and api we follow the triplet constrains of RDK
+impl TryFrom<&str> for Model {
+    type Error = ConfigModelError;
+    fn try_from(value: &str) -> Result<Self, Self::Error> {
```
npmenard, line 342: this is a bit exhaustive and we may consider using a Regex instead (like RDK) the code of doing so is unclear to me however

micro-rdk/src/common/config.rs
```diff
@@ -270,17 +299,225 @@ impl TryFrom<&google::protobuf::value::Kind> for Kind {
     }
 }
 
-#[derive(Debug, Default)]
+pub struct Model {
+    familly: String,
```
gvaradarajan, line 303: [sic]?

micro-rdk/src/common/config.rs
gvaradarajan: haven't tested it myself, but could we not just do something like this?
```rust
let reg = Regex::new(r"<namespace>:<family>:<model>").unwrap()?
let model_components = reg.captures(value)?;
let namespace = reg.get("namespace").ok_or(...).and_then(...)?;
...
```

micro-rdk/src/common/config.rs
```diff
@@ -354,50 +590,66 @@ impl<'a> ConfigType<'a> {
...
+        match self {
+            Self::Dynamic(cfg) => cfg.get_resource_name().get_subtype(),
         }
     }
 }
 
 pub trait Component {
+    fn get_resource_name(&self) -> &ResourceName;
+    fn get_model(&self) -> &Model;
+    #[deprecated(
```
gvaradarajan, line 609: @acmorrow must be so proud

micro-rdk/src/common/robot.rs
```diff
@@ -230,13 +205,19 @@ impl LocalRobot {
...
+                    cfg.get_model()
+                );
                 // capture the error and make it available to LocalRobot so it can be pushed in the logs?
                 if let Err(e) = self.build_resource(cfg, board.clone(), board_key.clone(), registry)
                 {
                     log::error!(
                         "Failed to build resource `{}` of type `{}`: {:?}",
-                        cfg.name,
-                        cfg.r#type,
+                        cfg.get_resource_name().get_name(),
```
gvaradarajan, line 219: is it worth implementing these kind of functions directly on `DynamicComponentConfig`?

gvaradarajan (approved):
Just a couple notes

micro-rdk/src/common/config.rs
acmorrow: Nothing makes me happier than a well executed deprecation!

micro-rdk/src/common/config.rs
npmenard: Sorry I meant the cost of using a regex :P. Seems like using the regex crate blows up the binary by 700k and blow up the stack. I kinda expected it but this is way more than i anticipated :P

PR #504, opened 2025-05-14
[RSDK-10661] register interrupt callbacks thru board

mattjperez:


micro-rdk/src/common/board.rs
```diff
@@ -85,6 +85,22 @@ pub trait Board: DoCommand {
         ))
     }
 
+    #[cfg(feature = "esp32")]
+    fn add_digital_interrupt_callback(
+        &mut self,
+        _pin: i32,
+        _intr_type: crate::esp32::esp_idf_svc::hal::gpio::InterruptType,
+        _cb: crate::esp32::esp_idf_svc::hal::sys::gpio_isr_t,
+        _arg: Option<*mut core::ffi::c_void>,
```
mattjperez, line 94: I considered making this a struct, but it complicated things without adding much value.
While we'd be able to provide a sane default for the callback, we wouldn't be able to do the same for the callback arg

micro-rdk/src/common/board.rs
```diff
@@ -85,6 +85,22 @@ pub trait Board: DoCommand {
...
+    ) -> Result<(), BoardError> {
+        unimplemented!();
+    }
+
+    #[cfg(feature = "esp32")]
+    fn remove_digital_interrupt_callback(&mut self, _pin: i32) -> Result<(), BoardError> {
```
mattjperez, line 100: will probably remove this. more info in the esp32 impl

micro-rdk/src/esp32/board.rs
```diff
@@ -213,39 +216,51 @@ impl EspBoard {
                 vec![]
             };
 
+            if let Ok(interrupt_confs) =
+                cfg.get_attribute::<Vec<DigitalInterruptConfig>>("digital_interrupts")
+            {
```
mattjperez, line 221: might embed this in the chain above.

micro-rdk/src/esp32/board.rs
```diff
@@ -365,4 +380,44 @@ impl Board for EspBoard {
...
+                BoardError::GpioPinError(
+                    pin as u32,
+                    "pin not found, failed to register interrupt callback",
+                )
+            })?;
+
+        if cb.is_some() {
+            p.setup_interrupt(intr_type, cb, arg.unwrap_or_else(core::ptr::null_mut))?;
+        } else {
+            let ptr = &mut p.event_count as *mut Arc<std::sync::atomic::AtomicU32> as *mut _;
```
mattjperez, line 405: since the  `event_count` is a field in the individual `Esp32GPIOPin` and pins are instantiated once per session, this should be fine.

micro-rdk/src/esp32/board.rs
```diff
@@ -365,4 +380,44 @@ impl Board for EspBoard {
...
+        Ok(())
+    }
+
+    fn remove_digital_interrupt_callback(&mut self, pin: i32) -> Result<(), BoardError> {
+        unsafe {
+            esp!(gpio_isr_handler_remove(pin))
+                .map_err(|e| BoardError::GpioPinOtherError(pin as u32, Box::new(e)))?;
+        }
+        Ok(())
+    }
```
mattjperez, line 422: will probably remove this. It was meant to be a wrapper for what the ultrasonic sensor does in its `impl Drop`, however the struct wouldn't have a handle to the board to use this method anyway.

micro-rdk/src/esp32/pin.rs
```diff
@@ -185,34 +185,26 @@ impl Esp32GPIOPin {
...
-                self.interrupt_type = Some(intr_type);
-            }
-        };
+    pub(crate) fn setup_interrupt(
+        &mut self,
+        intr_type: InterruptType,
+        cb: gpio_isr_t,
+        arg: *mut core::ffi::c_void,
+    ) -> Result<(), BoardError> {
+        self.interrupt_type = Some(intr_type);
```
mattjperez, line 194: regardless of what the previous interrupt type was, we still set it. before, we just returned `Ok` if the interrupt type was the same, but we wouldn't be able to handle if a module wanted to register a different callback that uses the same interrupt type as was originally set during board setup.

micro-rdk/src/esp32/board.rs
```diff
@@ -365,4 +372,36 @@ impl Board for EspBoard {
...
     }
+
+    fn add_digital_interrupt_callback(
+        &mut self,
+        pin: i32,
+        intr_type: InterruptType,
+        cb: gpio_isr_t,
+        arg: Option<*mut core::ffi::c_void>,
+    ) -> Result<(), BoardError> {
+        let p = self
```
npmenard, line 383: doesn't it check if the pin was configured as a digital interrupt pin? does it matter?

npmenard (changes requested):
Done with first pass.
It might make more sense to go down the road of using a  digital interrupt object.

micro-rdk/src/common/board.rs
```diff
@@ -85,6 +85,17 @@ pub trait Board: DoCommand {
         ))
     }
 
+    #[cfg(feature = "esp32")]
+    fn add_digital_interrupt_callback(
```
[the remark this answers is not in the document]
mattjperez: are you saying this should be part of `impl Esp32Board` instead?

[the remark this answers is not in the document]
mattjperez: removed esp32 flag

micro-rdk/src/common/board.rs
```diff
@@ -85,6 +85,16 @@ pub trait Board: DoCommand {
...
     }
 
+    fn add_digital_interrupt_callback(
+        &mut self,
+        _pin: i32,
+        _intr_type: InterruptType,
+        _cb: Option<unsafe extern "C" fn(_: *mut core::ffi::c_void)>,
+        _arg: Option<*mut core::ffi::c_void>,
+    ) -> Result<(), BoardError> {
+        unimplemented!();
```
npmenard, line 95: shouldn't return unimplemented here. Either make it a mandatory method or return a safe default

micro-rdk/src/common/board.rs
```diff
@@ -277,3 +287,17 @@ where
         self.lock().unwrap().set_pwm_frequency(pin, frequency_hz)
     }
 }
+
+pub struct DigitalInterruptConfig {
```
npmenard, line 291: I don't see this struct being used anywhere, remove it if unneeded

micro-rdk/src/common/board.rs
```diff
@@ -277,3 +287,17 @@ where
...
+    pub intr_type: InterruptType,
+    pub cb: Option<unsafe extern "C" fn(_: *mut core::ffi::c_void)>,
+    pub arg: Option<*mut core::ffi::c_void>,
+}
+
+pub enum InterruptType {
```
npmenard, line 297: place it above the FakeBoard implementation

micro-rdk/src/esp32/board.rs
npmenard: so it means that we can register any pin as interrupt as long as they were in the pin array attribute at least?

micro-rdk/src/esp32/pin.rs
```diff
@@ -185,34 +185,26 @@ impl Esp32GPIOPin {
...
-                }
-            }
-            None => {
-                self.interrupt_type = Some(intr_type);
-            }
-        };
+    pub(crate) fn setup_interrupt(
+        &mut self,
+        intr_type: InterruptType,
+        cb: gpio_isr_t,
```
npmenard, line 191: should check cb is some

micro-rdk/src/esp32/board.rs
npmenard: what is the risk  otherwise?

mattjperez: that the address of `event_count` might change if the `Esp32GPIOPin` moves somehow, or if the pin is re-instantiated, there might be a dangling pointer as the registered arg at the esp-idf level.

micro-rdk/src/esp32/board.rs
mattjperez: yes. it should fail when making the esp-idf interrupt handler [if the pin is not GPIO](https://github.com/espressif/esp-idf/blob/f9765d03162efd48feece14a4949de66356e00ae/components/esp_driver_gpio/src/gpio.c#L554). 

I just came across a warning that pins 36 and 39 shouldn't be used as an interrupt, so added a check for those.

micro-rdk/src/esp32/pin.rs
```diff
@@ -185,34 +185,43 @@ impl Esp32GPIOPin {
...
+        arg: *mut core::ffi::c_void,
+    ) -> Result<(), BoardError> {
+        if cb.is_none() {
+            return Err(BoardError::GpioPinOtherError(
+                self.pin as u32,
+                "interrupt callback cannot be `None`".to_string().into(),
+            ));
+        }
+
+        if self.pin == 36 || self.pin == 39 {
```
gvaradarajan, line 201: are there masks for determining this exposed by the `wifi` or `adc` headers like we do to determine valid output pins?

micro-rdk/src/common/board.rs
```diff
@@ -210,6 +226,16 @@ impl Board for FakeBoard {
         Err(BoardError::I2CBusNotFound(name))
     }
 
+    fn add_digital_interrupt_callback(
```
gvaradarajan, line 229: I think it would be good to have a doc string comment here? I think there's enough to explain as to why the signature is the way it is and the way developers should call this method

micro-rdk/src/common/board.rs
```diff
@@ -210,6 +226,16 @@ impl Board for FakeBoard {
...
+        &mut self,
+        _pin: i32,
+        _intr_type: InterruptType,
+        _cb: Option<unsafe extern "C" fn(_: *mut core::ffi::c_void)>,
```
gvaradarajan, line 233: `callback` instead of `cb`?

gvaradarajan (approved):
LGTM. Only blocking thing for me is the doc string comment

micro-rdk/src/esp32/pin.rs
mattjperez: not that I can find. Espressif doesn't do a hard check for this and only advises not to use them [here](https://docs.espressif.com/projects/esp-idf/en/stable/esp32/api-reference/peripherals/gpio.html#_CPPv416gpio_intr_enable10gpio_num_t) and in the doc linked

micro-rdk/src/esp32/pin.rs
```diff
@@ -185,34 +185,43 @@ impl Esp32GPIOPin {
...
+            // When ADC or WiFi and Bluetooth with sleep mode enabled are powered on,
```
npmenard, line 202: while I follow the logic here I am against adding exclusion that affect only specifics model of esp32s. In theory the same board implementation is used for esp32s2/s3/c6 and that bug doesn't exists there. We usually try to avoid this type of exclusion leaving the user to figure out what went wrong. 
lastly since gpio36 & 39 are input only they will be excluded by the is_gpio_valid check in pin.rs.

micro-rdk/src/common/board.rs
```diff
@@ -210,6 +228,16 @@ impl Board for FakeBoard {
...
     }
 
+    fn add_digital_interrupt_callback(
+        &mut self,
+        _pin: i32,
+        _intr_type: InterruptType,
+        _callback: Option<unsafe extern "C" fn(_: *mut core::ffi::c_void)>,
+        _arg: Option<*mut core::ffi::c_void>,
+    ) -> Result<(), BoardError> {
+        unimplemented!();
```
npmenard, line 238: we shouldn't call a function that will cause a panic even for a fakeboard implementation. The correct thing to do here is the return an error saying that the method is not supported

npmenard (approved):
LGTM

micro-rdk/src/esp32/board.rs
```diff
@@ -198,6 +211,12 @@ impl EspBoard {
                 };
 
             let pins = if let Ok(mut pins) = cfg.get_attribute::<Vec<i32>>("pins") {
+                if let Ok(interrupt_confs) =
+                    cfg.get_attribute::<Vec<DigitalInterruptConfig>>("digital_interrupts")
```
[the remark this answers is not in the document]
mattjperez: I'm not sure what you mean. The extension with the digital interrupt pins persists through the creation of the board, as it's part of what's collected on L230, so they will be present in the board's `pins` field.

This is a union of the config's pins and digital interrupts, so all `Esp32GPIOPin`s are initialized at once.
Originally, the initialization of the digital interrupt pins and adding them to `pins` were done in at the same time `setup_interrupt` was called.

micro-rdk/src/esp32/pin.rs
mattjperez: removed

PR #505, opened 2025-05-15
[RSDK-10692] use revision to check config change

mattjperez:
Tested using a config that has a `modules` attribute, prevents reboot.

Example logging of the revision
```sh
I (45445) micro_rdk::common::config_monitor: mperez-
prev_revision: 1747332178801
next_revision: 1747332219939
```

### DIFF
--- a/micro-rdk/src/common/config_monitor.rs
+++ b/micro-rdk/src/common/config_monitor.rs
@@ -17,7 +17,7 @@ use crate::common::{exec::Executor, ota};
 
 pub struct ConfigMonitor<'a, Storage> {
     /// revision of running `RobotConfig`
-    revision: String,
+    config_revision: String,
     storage: Storage,
     #[cfg(feature = "ota")]
     executor: Executor,
@@ -37,7 +37,7 @@ where
         restart_hook: impl Fn() + 'a,
     ) -> Self {
         Self {
-            revision: curr_config.revision.to_string(),
+            config_revision: curr_config.revision.to_string(),
             storage,
             #[cfg(feature = "ota")]
             executor,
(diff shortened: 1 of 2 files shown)


micro-rdk/src/common/config_monitor.rs
```diff
@@ -108,7 +108,7 @@ where
                     }
                 }
 
-                if *config != *self.curr_config {
+                if config.revision != self.curr_config.revision {
```
gvaradarajan, line 111: should we store the revision in NVS separately so that we don't have to load the whole config for this comparison?

mattjperez: yeah that would be better, and is how the OTA check works.

acmorrow: I'm not sure I like that idea, since it means we have the same information stored in two places, which means there is the opportunity for the two to desynchronize. However, it may be possible to reduce the memory requirements for this class by storing only the config revision rather than a copy of the whole config.

micro-rdk/src/common/config_monitor.rs
```diff
@@ -16,7 +16,8 @@ use std::{fmt::Debug, pin::Pin, time::Duration};
 use crate::common::{exec::Executor, ota};
 
 pub struct ConfigMonitor<'a, Storage> {
-    curr_config: Box<RobotConfig>, //config for robot gotten from last robot startup, aka inputted from entry
+    /// revision of running `RobotConfig`
+    revision: String,
```
npmenard, line 20: nit: config_revision

npmenard (approved):
LGTM might consider storing the revision as a timestamp to protect against potentially getting an older config (unlikely but never know)

acmorrow (approved):
LGTM

PR #506, opened 2025-05-16
implement ulp

npmenard:
This is the ULP example for the BME280, it went a bit further than I expected since I had to re-implement the BME280 so we can access the calibration  tables to compensate measurement gathered by the ULP.

The ULP is supposed to start when the BME280 sensor is dropped, the reasoning is that the BME280 needs to operate as a regular sensor since it will communicate overt I2C. As such the ULP can only run in deepsleep /sleep because it will take ownership of the SDA and SCL lines.

The ULP is supposed to support `I2C_RD` and `I2C_WR` instructions but because READ operation samples SDA on the falling edge of the SCL line then it may receive incorrect data. (See esp32 technical documentation section 30.6.21 for additional information) Since the issue affects us we are using bit-banging to communicate with BME280

The flow of the ULP is rather simple it accepts a sample number and data offset as arguments. It will read one sample updating variables afterwards, if sample becomes 0 then the ULP wakes the processor and stops itself otherwise it will sleep for  `ulp_set_wakeup_period(0, self.cfg.period.as_micros() as u32);`
To accommodate configuring pins we re-write some instructions  as `REG_RD` and `REG_WR` do not accept register as arguments

A config example looks like this : 
```
{
      "name": "board-1",
      "api": "rdk:component:board",
      "model": "rdk:builtin:esp32",
      "attributes": {
        "i2cs": [
          {
            "data_pin": 33,
            "bus": "i2c0",
            "name": "i2c_0",
            "clock_pin": 25
          }
        ]
      }
    },
    {
      "name": "bme280",
      "api": "rdk:component:sensor",
      "model": "rdk:builtin:bme280_ulp",
      "attributes": {
        "i2c_name": "i2c_0",
        "ulp_config": {
          "frequency_hz": 0.1,
          "ulp_scl_pin": 25,
          "ulp_sda_pin": 33
        }
      }
    }
```

There is currently a bug, when the chips wakes up from deep sleep the SDA and SCL are still tied to the RTC domain as such the onboard I2C hardware cannot access the peripheral. I am trying to identify a quick workaround but a solution might just be to forcibly reset the gpio state of SDA and SCL when configuring I2C.


examples/modular-drivers/src/bme280_ulp.rs
```diff
@@ -0,0 +1,970 @@
...
+    fn from(value: &'a UnsafeRtcMemory<[u32; SAMPLE_ARRAY_SIZE]>) -> Self {
+        let ptr = unsafe { &*value.0.get() };
+        Self {
+            memory: ptr,
+            offset: 0,
+        }
+    }
+}
+
+impl Iterator for ULPResultsMemory<'_> {
```
gvaradarajan, line 72: Is this the iterator that the data sync task is supposed to read from?

examples/modular-drivers/src/bme280_ulp.rs
```diff
@@ -0,0 +1,970 @@
...
+
+impl Sensor for BME280 {}
+
+impl Readings for BME280 {
+    fn get_generic_readings(
+        &mut self,
+    ) -> Result<micro_rdk::common::sensor::GenericReadingsResult, SensorError> {
+        let measurement = self.read_raw_measurement()?;
+        let measurement = CompensatedMeasurement::from_raw(measurement, &mut self.calib);
+        let mut res = GenericReadingsResult::new();
```
gvaradarajan, line 943: maybe implement `From<RawMeasurement>` for `GenericReadingsResult`?

gvaradarajan (commented):
Still in the process of diving deeper into the PR, but had a clarifying question.

examples/modular-drivers/src/bme280_ulp.rs
npmenard: no we would use it to construct Readings with their associated timestamp

examples/modular-drivers/src/bme280_ulp.rs
npmenard: we need the calibration data so can't impl from

examples/modular-drivers/build.rs
```diff
@@ -0,0 +1,58 @@
...
+    if Regex::new(r"\w+-esp3?2?s?\d?-espidf")
+        .unwrap()
+        .is_match(&std::env::var("TARGET").unwrap())
+    {
+        embuild::espidf::sysenv::output();
+        embuild::espidf::sysenv::env_path();
+        embuild::espidf::sysenv::idf_path();
+
+        let esp_idf_env =
+            PathBuf::from(std::env::var("DEP_MICRO_RDK_EMBUILD_ESP_IDF_PATH").unwrap());
```
npmenard, line 40: note that env can be accessed here because they are propagated by micro-rdk/build.rs which always run before this build script.

examples/modular-drivers/src/bme280_ulp.rs
```diff
@@ -0,0 +1,970 @@
...
+                            "frequency_hz should be a number".to_owned(),
+                        ))
+                    }
+                })?;
+            let cfg_sample = config
+                .get("sample")
+                .ok_or(AttributeError::KeyNotFound("sample".to_owned()))
+                .and_then(|kind| {
+                    if let Kind::NumberValue(smp) = kind {
+                        if *smp == 0.0 {
```
mattjperez, line 161: should this be <= 0.0?

mattjperez, line 161: should this be <= 0.0?

examples/modular-drivers/src/bme280_ulp.rs
```diff
@@ -0,0 +1,970 @@
...
+                    micro_rdk::esp32::esp_idf_svc::hal::ulp::ULP::MEM_START as *const u32,
+                ) as *const u32)
+                .map_err(BME280Error::BME280UlpCannotStart)?;
+        }
+
+        Ok(())
+    }
+}
+
+#[allow(dead_code)]
```
mattjperez, line 323: is this needed?

examples/modular-drivers/src/bme280_ulp.rs
```diff
@@ -0,0 +1,970 @@
...
+    HUM_REG = 0xFD,
+}
+
+impl From<Register> for u8 {
+    fn from(value: Register) -> Self {
+        value as u8
+    }
+}
+
+#[allow(dead_code)]
```
mattjperez, line 347: same as above

examples/modular-drivers/src/bme280_ulp.rs
```diff
@@ -0,0 +1,970 @@
...
+            .and_then(|_| {
+                esp!(unsafe {
+                    rtc_gpio_set_direction(self.cfg.pin_sda, RTC_GPIO_MODE_INPUT_OUTPUT)
+                })
+            })
+            .map_err(|err| BME280Error::BME280UlpRtcPinConfigError(self.cfg.pin_sda, err))?;
+
+        // config NOP reserved insts
+        unsafe {
+            *set_SCL = Self::ulp_fsm_reg_wr_inst(
```
mattjperez, line 263: so `i2c.S` (and others) are globals (whose values represent registers used by the ULP program) that you're populating here?

examples/modular-drivers/src/bme280_ulp.rs
```diff
@@ -0,0 +1,970 @@
...
+    fn drop(&mut self) {
+        if let Some(ulp) = &mut self.ulp {
+            if let Err(err) = ulp.start_ulp(self.config.clone()) {
+                log::error!("couldn't start BME280 ULP program reason : {}", err);
+            } else {
+                log::info!("started BME280 ULP program")
+            }
+        }
+    }
+}
```
mattjperez, line 644: what prompts the dropping of the sensor?

examples/modular-drivers/src/bme280_ulp.rs
npmenard: we are replacing NOP instruction with their respective REG_RD/REG_WR instructions. This portion of code is link the the data section (RW) and not the text section (normally RO)

examples/modular-drivers/src/bme280_ulp.rs
npmenard: nope was useful during dev :P

examples/modular-drivers/src/bme280_ulp.rs
npmenard: dropping the robot

examples/modular-drivers/src/bme280_ulp.rs
npmenard: yes

gvaradarajan (approved):
LGTM (I assume you're aware of the `#[allow(dead_code)]`s and plan to remove them before merge?)

One last question: was there a noticeable increase in build time resulting from having to additionally build the ULP sources?

PR #508, opened 2025-05-21
add simple test for digital interrupt pins, better error

mattjperez:
- add digital interrupt configs to robot test
- test that, in the absence of a pins array, digital interrupt pins are still configured as regular gpio pins
- fix an error return type

### DIFF
--- a/micro-rdk/src/common/robot.rs
+++ b/micro-rdk/src/common/robot.rs
@@ -741,6 +741,23 @@ mod tests {
                         ]),
                     ),
                     (
+                        "digital_interrupts".to_owned(),
+                        Kind::VecValue(vec![
+                            Kind::StructValue(HashMap::from([(
+                                "pin".to_owned(),
+                                Kind::StringValue("13".to_owned()),
+                            )])),
+                            Kind::StructValue(HashMap::from([(
+                                "pin".to_owned(),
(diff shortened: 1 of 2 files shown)


micro-rdk/src/common/robot.rs
```diff
@@ -740,6 +740,23 @@ mod tests {
...
+                    (
+                        "digital_interrupts".to_owned(),
+                        Kind::VecValue(vec![
+                            Kind::StructValue(HashMap::from([(
+                                "pin".to_owned(),
+                                Kind::StringValue("13".to_owned()),
+                            )])),
+                            Kind::StructValue(HashMap::from([(
+                                "pin".to_owned(),
+                                Kind::StringValue("14".to_owned()),
```
npmenard, line 752: should it be NumberValue? Probably doesn't matter anyway

npmenard (approved):
LGTM

micro-rdk/src/common/robot.rs
mattjperez: I just checked, either works, note that `pins` should also use `NumberValue(f64)`.

Will change these for digital interrupts to use `NumberValue`, then we'd indirectly catch any issues with the `Kind` conversion

PR #510, opened 2025-05-22
Reset GPIO pin in when configuring I2C

npmenard:
This a fix to reset the state of GPIO pin back to reset after waking up from sleep. Since the chip doesn't reset it's GPIO pin when waking up from sleep the pin may be configured in a way that make them unusable for I2C.
This shouldn't affect normal reset (power on, brown out etc..)


gvaradarajan (commented):
should we be doing this before every restart?

npmenard:
better to leave constrained to peripherals for now

gvaradarajan (approved):

PR #512, opened 2025-05-27
[RSDK-8749] add FakeServo, servo data capture

mattjperez:
verified on app by enabling data collection, and making a `FakeServo` with the following attribute
```
{
  "fake_position": 10
}
```


micro-rdk/src/common/servo.rs
```diff
@@ -47,3 +53,66 @@ where
...
+    fn move_to(&mut self, angle_deg: u32) -> Result<(), ServoError> {
+        self.pos = angle_deg;
+        Ok(())
+    }
+}
+
+#[cfg(feature = "builtin-components")]
+impl Actuator for FakeServo {
+    fn is_moving(&mut self) -> Result<bool, super::actuator::ActuatorError> {
+        Ok(self.pow > 0)
```
acmorrow, line 112: The fake servo moves to a new position instantaneously, so, is it ever moving? But pow is always 1, so it will also always be moving, even if no `move_to` was ever called.

acmorrow (approved):
Code looks fine, though the `FakeServo` has some somewhat inconsistent behavior. I don't know if we expect fakes to be even at all a little realistic, so maybe it doesn't matter.

mattjperez:
@acmorrow updated the `FakeServo` to mimic rdk's fake behavior

PR #513, opened 2025-05-27
[RSDK-10785] - implement firmware mode for power saving data collection

gvaradarajan:


micro-rdk/src/common/data_collector.rs
```diff
@@ -271,12 +271,12 @@ impl DataCollector {
...
     /// calls the method associated with the collector and returns the resulting data
     pub(crate) fn call_method(
-        &mut self,
+        &self,
         robot_start_time: Instant,
     ) -> Result<SensorData, DataCollectionError> {
         let reading_requested_ts = robot_start_time.elapsed();
-        let data = match &mut self.resource {
-            ResourceType::Board(ref mut res) => match &self.method {
+        let data = match self.resource.clone() {
```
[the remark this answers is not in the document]
gvaradarajan: I needed interior mutability because `PeriodicAppClientTask::invoke` uses an immutable reference

micro-rdk/src/common/registry.rs
```diff
@@ -649,6 +649,7 @@ mod tests {
             &config,
             &mut Box::new(registry),
             None,
+            common::conn::viam::FirmwareMode::Normal,
```
npmenard, line 652: why do we pull this here?

micro-rdk/src/common/conn/viam.rs
```diff
@@ -717,6 +734,36 @@ where
             );
         }
 
+        let (firmware_mode, active_timeout) = if let Some(app_client) = app_client.as_ref() {
+            let settings = app_client
+                .get_agent_config()
+                .map_ok(|res| res.advanced_settings.clone())
```
npmenard, line 740: avoid clone if possible

micro-rdk/src/common/conn/viam.rs
```diff
@@ -472,6 +473,22 @@ where
     }
 }
 
+const DEFAULT_ACTIVE_TIMEOUT_SECS: u64 = 120;
+
+pub enum FirmwareMode {
```
npmenard, line 478: i think it fits better in system

npmenard (changes requested):
Done with second pass

I don't like the idea of having a second task that mostly duplicates code. It will make maintaining it and iterating on it much harder over time. Can we refactor it such as that if we have to keep a second task then most of the code is common between them?

If its not feasible in time then we should at least add comments that clearly state why the code was duplicated and what contributors should be doing when changing this code, also would file a ticket with an detail explanation of how to iterate over that so we can reach a common code trunk between modes

micro-rdk/src/common/conn/viam.rs
```diff
@@ -717,6 +734,36 @@ where
...
+                    Some(Kind::StringValue(mode_str)) => mode_str.as_str(),
+                    _ => "normal",
+                });
+            (firmware_mode_str.into(), Duration::from_secs(timeout_secs))
+        } else {
+            (
+                FirmwareMode::Normal,
+                Duration::from_secs(DEFAULT_ACTIVE_TIMEOUT_SECS),
+            )
+        };
```
npmenard, line 765: do we print which firmware mode was selected in the logs?

micro-rdk/src/common/data_manager.rs
```diff
@@ -64,6 +65,10 @@ pub enum DataManagerError {
     MultipleConfigError,
     #[error(transparent)]
     InitializationRobotError(#[from] RobotError),
+    #[error("sync_interval required for this process")]
```
npmenard, line 68: better error message

micro-rdk/src/common/data_manager.rs
```diff
@@ -610,6 +635,117 @@ where
...
+        robot_start_time: Instant,
+    ) -> Result<Self, DataManagerError> {
+        if let Some(cfg) = get_data_service_config(cfg)? {
+            let (collectors, sync_interval, part_id) =
+                get_collectors_and_settings_from_robot_and_config(robot, &cfg)?;
+            match sync_interval {
+                None => Err(DataManagerError::MissingSyncInterval),
+                Some(sync_interval) => {
+                    if collectors.iter().any(|c| {
+                        (c.time_interval().as_secs_f32() - sync_interval.as_secs_f32()).abs()
```
npmenard, line 712: not sure what we are checking for here, can you provide more background?

micro-rdk/src/common/robot.rs
```diff
@@ -276,22 +277,46 @@ impl LocalRobot {
...
-                    if let Some(task) = data_manager.get_sync_task(robot.start_time) {
-                        let _ = robot.data_manager_sync_task.insert(Box::new(task));
+            match firmware_mode {
+                // TODO(RSDK-8125): Support selection of a DataStore trait other than
+                // DefaultDataStore in a way that is configurable
+                FirmwareMode::Normal => {
+                    match DataManager::<DefaultDataStore>::from_robot_and_config(&robot, config) {
+                        Ok(None) => {}
+                        Ok(Some(mut data_manager)) => {
+                            log::info!("running normal data management");
```
npmenard, line 287: i see the re is a log here. I would rather just have a log saying "firmware is in normal mode" or "firmware will sleep between cloud data sync" or something like that

micro-rdk/src/common/data_manager.rs
```diff
@@ -610,6 +635,117 @@ where
...
+        }
+    }
+}
+
+impl PeriodicAppClientTask for DataCollectAndSyncTask {
+    fn name(&self) -> &str {
+        "DataCollectAndSync"
+    }
+    fn get_default_period(&self) -> Duration {
+        Duration::from_secs(0)
```
npmenard, line 739: so if the data sync task fails during collection/sync it will run indefinitely?

gvaradarajan: yes, I have a fix for this in the next PR, but I'll bring it into this one

micro-rdk/src/common/data_manager.rs
gvaradarajan: will put a comment detailing this, but just to answer the question; we discussed wanting to temporarily validate that every collection frequency was equivalent to the sync frequency when using this mode. However due to variances in float division we cannot merely check "==", so we need to instead validate that the discrepancy in seconds is lower than a tolerated threshold

micro-rdk/src/common/data_manager.rs
```diff
@@ -329,6 +346,54 @@ pub enum DataSyncError {
...
+    let time_to_subtract = chrono::Duration::new(
+        time_to_subtract.as_secs() as i64,
+        time_to_subtract.subsec_nanos(),
+    )
+    .ok_or(DataSyncError::TimeOutOfBoundsError)?;
+    Ok(time_to_subtract)
+}
+
+fn time_correct_reading(
+    robot_start_time: Instant,
```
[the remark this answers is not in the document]
gvaradarajan: It's in the follow-up PR, but essentially I've created an alternate trait method in `Readings` for implementers that manage their own data caching with a return signature of `Result<Vec<SensorData>, SensorError>`. The instances of `SensorData` will already have the correct timestamps, so I believe the correction math will retain the original value?

micro-rdk/src/common/robot.rs
```diff
@@ -276,22 +277,46 @@ impl LocalRobot {
         // TODO: When cfg's on expressions are valid, remove the outer scope.
         #[cfg(feature = "data")]
```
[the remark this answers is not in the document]
gvaradarajan: it is, I'll make a ticket

micro-rdk/src/common/robot.rs
gvaradarajan: I'll move this logging to when the `FirmwareMode` is first parsed

micro-rdk/src/common/data_collector.rs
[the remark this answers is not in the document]
gvaradarajan: the alternative is to make `PeriodicAppClientTask::invoke` take a mutable reference, do we prefer that?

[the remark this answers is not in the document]
gvaradarajan: resolved offline, keeping the clone for now

micro-rdk/src/common/conn/viam.rs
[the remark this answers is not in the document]
gvaradarajan: it can't be `pub(crate)` because `from_cloud_config` is `pub`, should that be `pub(crate)` as well?

npmenard (approved):
LGTM

PR #514, opened 2025-05-28
[RSDK-8747] data capture encoder

mattjperez:
Tested using a fake encoder and viewing the data on app.


mattjperez:
the changes made to get fake data to the app for testing broke a couple tests. need to update the tests appropriately before merging.

micro-rdk/src/common/encoder.rs
```diff
@@ -218,8 +243,8 @@ impl Default for FakeEncoder {
 impl FakeEncoder {
     pub fn new() -> Self {
         Self {
-            angle_degrees: 0.0,
-            ticks_per_rotation: 1,
+            angle_degrees: 360.0,
+            ticks: AtomicU32::new(0),
```
acmorrow, line 247: Why does this need to be an atomic now?

mattjperez: it's how we can increment the ticks from inside `get_position` without a mutable reference.
RDK uses a background thread that uses speed and time durations to calculate these, but we don't have that.

acmorrow (approved):
LGTM

gvaradarajan (approved):

PR #526, opened 2026-06-17
Replace defunct buildjet runners with GitHub-hosted runners

stuqdog:
The buildjet runners no longer work, breaking these workflows. This maps them to GitHub-hosted runners, matching the fix already made in [viamrobotics/rust-utils#178](https://github.com/viamrobotics/rust-utils/pull/178):

- `buildjet-*-ubuntu-2204-arm` (native arm64 builds) -> `github-linux-arm64-8core`
- `buildjet-*-ubuntu-2204` (x86) -> `ubuntu-large`

These are all native per-arch builds (arm64 container images / `--platform linux/arm64`), so the arm jobs map to the arm GitHub runner. Where present, the now-unused buildjet labels are also dropped from `.github/actionlint.yml`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -66,7 +66,7 @@ jobs:
 
   # github-actions does not have a runner that supports arm64 linux and cross-compilation does not seem to work for the ring crate
   build-installer-arm64-linux:
-    runs-on: [buildjet-8vcpu-ubuntu-2204-arm]
+    runs-on: [github-linux-arm64-8core]
     # by default github use sh as shell
     defaults:
       run:


acmorrow (approved):
The failing format job isn't your doing. I think this is fine to merge. Let me know if you are unable to do so and I'll see what I can do to unblock it.

stuqdog:
> The failing format job isn't your doing. I think this is fine to merge. Let me know if you are unable to do so and I'll see what I can do to unblock it.

Yeah I can't merge, if you can do so that'd be great, thanks!

